PR #114, opened 2014-11-30
Add thrift

ahawkins:
I'm not sure if I'm following the correct format. It seems there  is a `content.md` and `README.md` in the folder. They seem to share similar content. `README.md` usually contained the most information so I created
one for Thrift.

### DIFF
--- a/thrift/user-feedback.md
+++ b/thrift/user-feedback.md
@@ -1,0 +1,20 @@
+## Issues
+
+If you have any problems with or questions about this image, please contact me
+through a [GitHub issue](https://github.com/ahawkins/docker-thrift/issues).
+
+You can also reach many of the official image maintainers via the
+`#docker-library` IRC channel on [Freenode](https://freenode.net).
+
+## Contributing
+
+You are invited to contribute new features, fixes, or updates, large
+or small; I am always thrilled to receive pull requests, and do my
+best to process them as fast as I can.
+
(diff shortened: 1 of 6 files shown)


ahawkins:
@tianon there you go! Thanks for the helpful guide on how to make this happen. Sorry it took so long :(

thrift/user-feedback.md
```diff
@@ -0,0 +1,20 @@
...
+You are invited to contribute new features, fixes, or updates, large
+or small; I am always thrilled to receive pull requests, and do my
+best to process them as fast as I can.
+
+Before you start to code, I recommend discussing your plans through a
+[GitHub issue](https://github.com/ahawkins/docker-thrift/issues),
+especially for more ambitious contributions. This gives other
+contributors a chance to point you in the right direction, give you
+feedback on your design, and help you find out if someone else is
+working on the same thing.
```
tianon: This file seems like it's pretty close to the standard template -- is that right?

tianon:
Generally LGTM, and don't worry -- it was the holidays after all.

Only the one comment from me; cc @yosifkit.

thrift/user-feedback.md
yosifkit: Looks like just a `s/ we / I /g`?

yosifkit:
LGTM

PR #120, opened 2014-12-10
[ADD] Odoo documentation

aab-odoo:


odoo/README-short.txt
```diff
@@ -0,0 +1 @@
+Odoo (formerly known as OpenERP) is a suite of open-source business apps.
```
tianon: Ohhh, Odoo literally is the continuation of the OpenERP work?  I thought it was more of a wrapper/add-on, but if it's a fork/rename that makes much more sense. :+1:

> Odoo was formerly known as OpenERP until May 2014. It was rebranded because version 8 of the software included apps including website builder, e-commerce, point of sale and business intelligence. The software conforms to standard expectations of ERP systems, while providing additional modules beyond the coverage of traditional ERP systems.

odoo/content.md
```diff
@@ -0,0 +1,83 @@
...
+	
+	docker run -p 127.0.0.1:8070:8069 --name odoo2 --link db:db -t odoo
+	docker run -p 127.0.0.1:8071:8069 --name odoo3 --link db:db -t odoo
+
+Please note that for plain use of mails and reports functionalities, when the 
+host and container ports differ (e.g. 8070 and 8069), one has to set, 
+in Odoo, Settings->Parameters->System Parameters (requires technical features), 
+web.base.url to the container port (e.g. 127.0.0.1:8069).
+
+## Access an existing database with a new Odoo instance
```
tianon: I think this should be the "how to upgrade" section.  Using `docker exec` to upgrade a running container in-place is not something we recommend anywhere, especially since you then miss out on updates to the base image.  We instead canonically recommend that users download the updated image and replace their existing containers, as described here.

psftw:
LGTM.  I just ran an instance following the instructions and it worked well.  The only nitpick I could come up with is that you can pass just the configuration file in as a volume instead of the entire folder, but it isn't something that needs to be changed (unless you want to appease my OCD).

odoo/content.md
aab-odoo: Up!

yosifkit:
LGTM

odoo/README.md
```diff
@@ -0,0 +1,111 @@
...
+The default configuration file for the server (located at `/etc/odoo/openerp-server.conf`)
+can be overriden at startup using volumes. Suppose you have a custom configuration
+at `/path/to/config/openerp-server.conf`, then
+
+	docker run -v /path/to/config:/etc/odoo -p 127.0.0.1:8069:8069 --name odoo --link db:db -t odoo
+
+## Run multiple Odoo instances
+	
+	docker run -p 127.0.0.1:8070:8069 --name odoo2 --link db:db -t odoo
+	docker run -p 127.0.0.1:8071:8069 --name odoo3 --link db:db -t odoo
```
yosifkit: Will this work with multiple odoo connected to the same database?

aab-odoo: It will work even if it doesn't really make sense... The classic use case would be to create a database for each odoo instance.

yosifkit:
Ping @tianon, (in other words, LGTM).

tianon:
Excellent, thanks :heart:

LGTM

PR #130, opened 2014-12-31
Add description for `docker-php-ext-install`

AaronJan:
Let people know there's a script so cool.

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -72,7 +72,7 @@ easily install PHP extension.
 
 For example, if you want to have a PHP-FPM image with `gd` and `mcrypt` 
 extensions, you can inheriting the base image that you like, and write your own 
-Dockerfile like this:
+`Dockerfile` like this:
 
     FROM php:5.5.19-fpm
     # Install modules
--- a/php/content.md
+++ b/php/content.md
@@ -65,6 +65,27 @@ same commands to build and run:
 Where `src/` is the directory containing all your php code and `config/`
 contains your `php.ini` file.
 
+### How to install more PHP extensions
+
+We provide a convenient script named `docker-php-ext-install`, you can use it to
+easily install PHP extension.
+
+For example, if you want to have a PHP container with `gd` and `mcrypt` 
+extensions, you can write your own Dockerfile like this:
+
+    FROM php:5.5.19-fpm
+    # Install modules
+    RUN apt-get update && apt-get install -y \
+        apt-utils re2c g++ \
+        zlib1g zlib1g-dbg zlib1g-dev zlibc \
+        libpng12-0 libpng12-dev libpng3 \
+        libjpeg9 libjpeg9-dbg libjpeg9-dev \
+        libmcrypt-dev libmcrypt4 mcrypt \
+        && docker-php-ext-install gd mcrypt
+    CMD ["php-fpm"]
+
+Remember, you must install dependencies for your extensions manually.
+
 ### Without a `Dockerfile`
 
 If you don't want to include a `Dockerfile` in your project, it is sufficient to
--- a/php/content.md
+++ b/php/content.md
@@ -67,25 +67,24 @@ contains your `php.ini` file.
 
 ### How to install more PHP extensions
 
-We provide a convenient script named `docker-php-ext-install`, you can use it to
+We provide two convenient scripts named `docker-php-ext-configure` and `docker-php-ext-install`, you can use them to
 easily install PHP extension.
 
-For example, if you want to have a PHP-FPM image with `gd` and `mcrypt` 
+For example, if you want to have a PHP-FPM image with `iconv`, `mcrypt` and `gd` 
 extensions, you can inheriting the base image that you like, and write your own 
 `Dockerfile` like this:
 
-    FROM php:5.5.19-fpm
+    FROM php:5.5-fpm
     # Install modules
     RUN apt-get update && apt-get install -y \
-        apt-utils re2c \
-        zlib1g zlib1g-dbg zlib1g-dev zlibc \
-        libpng12-0 libpng12-dev libpng3 \
-        libjpeg9 libjpeg9-dbg libjpeg9-dev \
-        libmcrypt-dev libmcrypt4 mcrypt \
-        && docker-php-ext-install gd mcrypt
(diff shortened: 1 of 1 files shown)


AaronJan:
Updated.    : )
Is this version good?

php/content.md
```diff
@@ -65,6 +65,27 @@ same commands to build and run:
...
+    FROM php:5.5.19-fpm
+    # Install modules
+    RUN apt-get update && apt-get install -y \
+        apt-utils re2c g++ \
+        zlib1g zlib1g-dbg zlib1g-dev zlibc \
+        libpng12-0 libpng12-dev libpng3 \
+        libjpeg9 libjpeg9-dbg libjpeg9-dev \
+        libmcrypt-dev libmcrypt4 mcrypt \
+        && docker-php-ext-install gd mcrypt
+    CMD ["php-fpm"]
```
tianon: The following simple example works well for me (to get both `gd` and `mcrypt`):
```Dockerfile
FROM php:5.6-fpm
RUN apt-get update && apt-get install -y \
        libpng12-dev \
        libmcrypt-dev \
    && docker-php-ext-install gd mcrypt
```

php/content.md
```diff
@@ -65,6 +65,28 @@ same commands to build and run:
...
 contains your `php.ini` file.
 
+### How to install more PHP extensions
+
+We provide a convenient script named `docker-php-ext-install`, you can use it to
+easily install PHP extension.
+
+For example, if you want to have a PHP-FPM image with `gd` and `mcrypt` 
+extensions, you can inheriting the base image that you like, and write your own 
+Dockerfile like this:
```
tianon: Also, `Dockerfile` should be in backticks too. :+1:

AaronJan:
Updated.  : )

php/content.md
yosifkit: The example is using php 5.5, which is probably more common than @tianon's example with 5.6 (since it is still quite new).

We should minimally change to `php:5.5-fpm` so that it does not have to be updated quite as often, rather than the `5.5.19-fpm` which is already out of date.

Also, `libjpeg9` (plus dev and dbg versions) does not exist in `debian:jessie`; it has been replaced by `libjpeg62-turbo`.  Does it really need all those dependencies to build and later run?  Can we simplify it like @tianon's example?

schmunk42:
Why doesn't the script `docker-php-ext-install` install the required packages? We would only have collect the deps once and put them into the script.

tianon:
"only"

We'd have to collect the dependencies of every module in the PHP core, which is not a small list: (`Possible values for ext-name:`)
```console
$ docker run -it --rm php:5.6 docker-php-ext-install
usage: /usr/local/bin/docker-php-ext-install ext-name [ext-name ...]
   ie: /usr/local/bin/docker-php-ext-install gd mysqli
       /usr/local/bin/docker-php-ext-install pdo pdo_mysql

if custom ./configure arguments are necessary, see docker-php-ext-configure

Possible values for ext-name:
bcmath bz2 calendar ctype curl dba dom enchant exif fileinfo filter ftp gd gettext gmp hash iconv imap interbase intl json ldap mbstring mcrypt mssql mysql mysqli oci8 odbc opcache pcntl pdo pdo_dblib pdo_firebird pdo_mysql pdo_oci pdo_odbc pdo_pgsql pdo_sqlite pgsql phar posix pspell readline recode reflection session shmop simplexml snmp soap sockets spl standard sybase_ct sysvmsg sysvsem sysvshm tidy tokenizer wddx xml xmlreader xmlrpc xmlwriter xsl zip
```

schmunk42:
But every developer has to look up the deps manually or did I miss something here. How do you know which packages to install first?

tianon:
I usually just try it.  When it complains about a header file (like, `png.h` for example), I search for that on https://packages.debian.org (ie, https://packages.debian.org/file:png.h, from which I find that `libpng12-dev` is the likely candidate, which means that `libpng12` is _probably_ necessary at runtime too, but it's worth double-checking that by removing `libpng12-dev` afterwards and then running the image -- if I get errors about `libpng` missing, I clearly need the runtime library too).

There's not an existing reference of every module's required packages that I'm aware of, and as you can see, it would be fairly tedious to create and would have quite the potential for getting outdated since we have such a large surface area, so I wouldn't be comfortable with such a list unless it also came with a nice way to verify that the list is still current and accurate (including whether the modules referenced still do require the packages we think they do).

schmunk42:
I totally agree with you that the list needs to be maintained and for sure this is some work to do.

But all the things you mentioned are a reason to maintain such a list, everyone has to find out what works over and over again, especially when there are updates, IMHO.

At the moment, to be honest, this keeps me a away from using this image (which I'd really like to do) - but it's much more to work to try and error with the dependencies and build over and over again, than just running `apt-get install php5-foobar` on a debian image, which is much more error prone for me.

tianon:
If someone is willing to maintain such a list, then by all means, please step up.  I'm not opposed to such a list if it has a dedicated maintainer who's willing to constantly update it will all possible permutations.

AaronJan:
Intresting, this list could be really useful.

Is there any one?

AaronJan:
Had a problem.

Here is the `gd_info()` output:

```
[GD Version] => bundled (2.1.0 compatible)
[FreeType Support] => 
[T1Lib Support] => 
[GIF Read Support] => 1
[GIF Create Support] => 1
[JPEG Support] => 
[PNG Support] => 1
[WBMP Support] => 1
[XPM Support] => 
[XBM Support] => 1
[JIS-mapped Japanese Font Support] => 
```

`FreeType` and `JPEG` are not working, so I start a new `php:5.5.19-fpm` container then I try this: 
`apt-get update && apt-get install -y zlib1g zlib1g-dbg zlib1g-dev zlibc libtool libfreetype6 libfreetype6-dev libjpeg62-turbo libjpeg62-turbo-dbg libjpeg62-turbo-dev libpng12-0 libpng12-dev libpng3 `

Still can't make them enabled. What should I do?

php/content.md
AaronJan: Updated.

I keep getting errors like `403  Forbidden [IP: 193.140.100.100 80]`, can't test it now.

yosifkit:
LGTM

tianon:
Awesome, thanks!  LGTM

thaJeztah:
Thanks!

PR #137, opened 2015-01-14
Add docs for RethinkDB

stuartpb:
Following up from docker-library/official-images#375


rethinkdb/content.md
```diff
@@ -0,0 +1,46 @@
...
+groupings, aggregations, and functions.
+
+%%LOGO%%
+
+# How to use this image
+
+## start an instance allowing connections on all interfaces
+
+```bash
+docker run --name some-rethink -d rethinkdb --bind all
```
yosifkit: This currently won't run without another `rethinkdb` in there since this first `rethinkdb` is the name of the image, or since CMD already has `--bind all`, we can drop that.

So, one of these:
```
docker run --name some-rethink -d rethinkdb
docker run --name some-rethink -d rethinkdb rethinkdb --bind all
```

rethinkdb/content.md
```diff
@@ -0,0 +1,46 @@
...
+
+## ... via the web admin interface ...
+
+```bash
+# browsing on the same machine:
+xdg-open \
+  "http://$(docker inspect --format '{{ .NetworkSettings.IPAddress }}'):8080"
+
+# exposing and connecting remotely:
+docker run -d --link some-rethink:rdb -p 8080:8080 svendowideit/ambassador
```
yosifkit: I think you meant `rethinkdb` and not `svendowideit/ambassador`. :wink:

update.sh
```diff
@@ -58,67 +59,67 @@ for repo in "${repos[@]}"; do
 	if [ -x "$repo/update.sh" ]; then
 		( set -x; "$repo/update.sh" )
 	fi
-	
+
```
yosifkit: Can you drop all these unrelated whitespace changes?

rethinkdb/content.md
yosifkit: If we remove the "--bind all" here, I would add an explanation that it is part of the default CMD.

rethinkdb/content.md
stuartpb: No, that's supposed to be ambassador. It runs `socat` to forward ports from the running RethinkDB instance to the exposed host port (the ambassador pattern).

rethinkdb/content.md
stuartpb: Moreover, I think I should probably add a volume to the arg, since otherwise you're just running an ephemeral DB, which is somewhat useless.

rethinkdb/content.md
```diff
@@ -0,0 +1,46 @@
...
+```bash
+docker run --name some-app --link some-rethink:rdb -d application-that-uses-rdb
+```
+
+## ... via the web admin interface ...
+
+```bash
+# browsing on the same machine:
+xdg-open \
+  "http://$(docker inspect --format '{{ .NetworkSettings.IPAddress }}'):8080"
```
stuartpb: I just noticed I forgot `some-rethink` in this command.

update.sh
stuartpb: I can, although I'd still like to have a commit that strips trailing whitespace merged in at some point.

stuartpb:
Amended to incorporate feedback.

rethinkdb/content.md
```diff
@@ -0,0 +1,50 @@
...
+# How to use this image
+
+## start an instance with data mounted in the working directory
+
+The default CMD of the image is `rethinkdb --bind all`, so the RethinkDB daemon
+will bind to all network interfaces available to the container (by default,
+RethinkDB only accepts connections from `localhost`).
+
+```bash
+docker run --name some-rethink -v "$PWD:/data" -d rethinkdb
```
yosifkit: This should be `-v "$(pwd):/data"`, or it could be `-v "/path/to/my/data:/data"` or even left out and the data will be preserved in a docker volume since "Volumes persist until no containers use them" ([docker docs](https://docs.docker.com/userguide/dockervolumes/)).

rethinkdb/content.md
```diff
@@ -0,0 +1,50 @@
...
+
+## ... via the web admin interface ...
+
+```bash
+# browsing on the same machine:
+xdg-open "http://$(docker inspect --format \
+  '{{ .NetworkSettings.IPAddress }}' some-rethink):8080"
+
+# exposing and connecting remotely:
+docker run -d --link some-rethink:rdb -p 8080:8080 svendowideit/ambassador
```
yosifkit: I am not sure we want to advocate running a non-official image here.  I don't see anything it is doing that cannot be done with docker directly.  Wouldn't it be simpler to just have:
```
docker run --name some-rethink -d -p 8080:8080 rethinkdb
```

stuartpb: That doesn't connect to the *running* container, though, it starts a *new* one (which isn't much help if you're opening the admin page to diagnose an issue in the container you already started).

This ambassador pattern comes directly from [the official Docker documentation](https://docs.docker.com/articles/ambassador_pattern_linking/).

rethinkdb/content.md
stuartpb: The `$PWD` variable will work in any shell that supports `$()` for command substitution (both of these have been widespread standards since somewhere around 1980).

update.sh
stuartpb: (proposed separately and comprehensively as #138)

rethinkdb/README.md
```diff
@@ -0,0 +1,92 @@
...
+docker run -d --link some-rethink:rdb -p 8080:8080 svendowideit/ambassador
+echo "click me: http://$(hostname):8080"
+```
+
+## ... or via the command-line interface
+
+```bash
+docker run -it --link some-rethink:rdb --rm rethinkdb sh -c 'exec \
+  rethinkdb admin --join "$RDB_PORT_29015_TCP_ADDR:$RDB_PORT_29015_TCP_PORT"'
+```
```
stuartpb: This will need to be either replaced or altogether removed, as 1.16.0 has removed the `rethinkdb admin` CLI. (Presumably, the best way to administer RethinkDB from the command line now is using the REPL of a language with an official RethinkDB driver.)

rethinkdb/content.md
tianon: For now, would you mind switching to `-v "$(pwd)":/data` so that we're consistent with all the other current docs, and then we can open a new issue/PR to discuss this change across the board?  I'm +1 on the change from a quick bit of research, but I'd rather we stay consistent and change this all at once if we're going to.

rethinkdb/content.md
tianon: This would be the first official image description that advocates the use of a non-official image, which is what @yosifkit was trying to get at.

Besides that, I personally don't feel like mentioning the ambassador adds much to this example -- as you've pointed out, ambassadors are described in the Docker docs, so it feels like orthogonal information that'd certainly be useful for every image (official or not), which is why I think leaving that as an exercise for the reader is acceptable.

rethinkdb/README.md
tianon: I think for now, let's just remove it and leave this as an exercise for the
reader. :+1:

rethinkdb/content.md
stuartpb: Sure.

rethinkdb/content.md
stuartpb: There are lots of things in this doc that are "orthogonal information that'd certainly be useful for every image (official or not)", like explaining how to start an application container with `--link` to connect to it. These things *are* listed in the docs for each image.

I agree that explaining that you can expose *any* ports via `ambassador`, ie. for the purpose of *ops*, would be too broad to mention in every image's documentation (especially considering that any ops configuration that could use it will require further thought to decide how it fits in). However, administering a currently-running container is a *common and straightforward* use scenario, and it's one that's liable to be otherwise misunderstood by the user (as seen by @yosifkit's comment).

If there's an issue with `ambassador` not being an official image, I think *making* it one is the obvious solution on that front.

yosifkit: I still see no need for the `ambassador` container.  You can just connect to the IP address of the container directly and if you are using running the containers in a VM (akin to boot2docker) or remotely then an ssh tunnel will do the trick.

stuartpb: So that'd be, what, `ssh -nNTL 8080:$(ssh user@remote docker inspect --format \
+  '{{ .NetworkSettings.IPAddress }}' some-rethink):8080 user@remote; xdg-open http://localhost:8080`?

yosifkit: IMO just making it simple like we do for [rails](https://github.com/docker-library/docs/tree/master/rails#create-a-dockerfile-in-your-rails-app-project) and just giving the two ways to run it is easier, but if you want to document the ssh version, let's put it in.

WDYT @stuartpb @tianon?

stuartpb:
New version pushed with `$(pwd)`, SSH forwarding, and no `rethinkdb admin`, per previous discussion.

tianon:
LGTM

yosifkit:
LGTM

PR #138, opened 2015-01-14
Strip trailing whitespace

stuartpb:
This commit brought to you by:

    grep -Irl . ./* | xargs sed -i -e's/[[:space:]]*$//'


LICENSE
```diff
@@ -1,5 +1,5 @@
  Copyright (c) 2014 Docker, Inc.
- 
+
```
tianon: Hah, I hadn't noticed that our entire `LICENSE` file is indented with a single space.  We should probably just strip all the leading whitespace in this whole file.

README.md
```diff
@@ -69,13 +69,13 @@ is generated by the script but can be overridden.  The following is a basic
 layout:
 
     # What is XYZ?
-    
+
```
tianon: This is intentionally left indented to show that this whole indented block is one section of contiguous "code" (to help ensure that Markdown translates it properly into a single contiguous `<code>` element).

buildpack-deps/README.md
```diff
@@ -53,7 +53,7 @@ You are invited to contribute new features, fixes, or updates, large or small;
 we are always thrilled to receive pull requests, and do our best to process them
 as fast as we can.
 
-Before you start to code, we recommend discussing your plans 
+Before you start to code, we recommend discussing your plans
```
tianon: This is autogenerated (as are most of the `*/README.md` files).  To fix this properly (and not have it come back), we need to change the template where this is generated (ie, remove the "mailing list" logic and have the one or two repos using it just use an entirely custom "user feedback" section instead).

generate-dockerfile-links-partial.sh
```diff
@@ -31,7 +31,7 @@ for repoDir in "${repoDirs[@]}"; do
 		# skip non-github.com for now
 		continue
 	fi
-	
+
```
tianon: This one is just because we prefer to keep our whitespace properly indented for the enclosing block so it's easy to jump in and keep going.  I could be convinced to cave on this one (since it is often a source of contributor pain when they have to edit a script and their editor doesn't preserve these, for example), but the rest of this PR's future is definitely hazy. :wink:

hello-world/README.md
```diff
@@ -13,7 +13,7 @@ repo](https://github.com/docker-library/official-images).
     $ docker run hello-world
     Hello from Docker.
     This message shows that your installation appears to be working correctly.
-    
+
```
tianon: This one is autogenerated as well.  See `hello-world/update.sh`.

jenkins/README.md
```diff
@@ -93,7 +93,7 @@ jenkins_home) at any time.
 If your volume is inside a container - you can use `docker cp
 $ID:/var/jenkins_home` command to extract the data.
 
-## Attaching build executors 
+## Attaching build executors
```
tianon: :+1:

README.md
stuartpb: Markdown doesn't need blank lines to be four-space indented, though.

jenkins/README.md
tianon: (but because it's matching in `content.md`, where this is generated from)

odoo/content.md
```diff
@@ -1,13 +1,13 @@
...
-2.000.000 users worldwide ranging from very small companies (1 user) to very 
-large ones (300 000 users). 
+Odoo, formerly known as OpenERP, is a suite of open-source business apps
+written in Python and released under the AGPL license. This suite of
+applications covers all business needs, from Website/Ecommerce down to
+manufacturing, inventory and accounting, all seamlessly integrated. It is the
+first time ever a software editor managed to reach such a functional coverage.
+Odoo is the most installed business software in the world. Odoo is used by
+2.000.000 users worldwide ranging from very small companies (1 user) to very
+large ones (300 000 users).
```
tianon: :+1: for all the changes in this file too :smile:

oraclelinux/content.md
```diff
@@ -3,16 +3,16 @@
...
-downtime kernel patching, DTrace for real-time diagnostics, the powerful Btrfs 
+downtime kernel patching, DTrace for real-time diagnostics, the powerful Btrfs
 file system, and more.
 
 ## How to use these images
 
-The Oracle Linux images are intended for use in the **FROM** field of an 
-application's `Dockerfile`. For example, to use Oracle Linux 6 as the 
+The Oracle Linux images are intended for use in the **FROM** field of an
+application's `Dockerfile`. For example, to use Oracle Linux 6 as the
```
tianon: Ahh, these are good too. :+1:

buildpack-deps/README.md
stuartpb: You're talking about this one specific space being generated as part of the template script (due to the whole mailing-list insertion logic)?

LICENSE
stuartpb: I'll make a second commit that does that.

buildpack-deps/README.md
yosifkit: Correct, the whole "User Feedback" section is generated.  We need to adjust and drop the mailing-list insertion logic, since this is the result when there is no list (and just customize the 2-3 that use it).

stuartpb: Just added a commit for exactly that.

hello-world/README.md
stuartpb: Fixed in most recent commit.

yosifkit:
While I am not against removing trailing whitespace on many of the lines, I am very -1 on removing it on empty lines when it matches the surrounding blocks.

stuartpb:
It's not worth leaving them in the file for that reason (while breaking the default behavior of `git apply` and most editors). Your editor should be stripping the space and matching the indentation for you. For example, see http://stackoverflow.com/questions/6000842/autoindent-in-vim-always-empty-the-line-if-there-are-only-whitespace-in-it-can for how to do this properly in Vim.

generate-dockerfile-links-partial.sh
stuartpb: See https://github.com/docker-library/docs/pull/138#issuecomment-72031565

yosifkit:
I explicitly tell vim not to delete the whitespace on empty lines; a user and their editor should not go about changing other people's files when it is not explicitly commanded.  If your editor just willy-nilly deletes whitespace from lines, then use `git add -p` and skip the superfluous changes.

tianon:
Also, "properly" is entirely subjective, which is why I'm not a fan of
where this is going on the whole.

I'm not seeing a solid justification for us to change our workflow to
accommodate other folks' choice in editor settings, especially since most
folks are content to match the convention of the surrounding code.

yosifkit:
This is no longer applicable since we use markdownfmt.

stuartpb:
What about the leading spaces in LICENSE?

yosifkit:
Hmm, it seems @tianon already pushed that in :smile:.

PR #140, opened 2015-01-16
Add some documentation about where the line between the new variants of buildpack-deps is

tianon:
See https://github.com/docker-library/buildpack-deps/pull/15

### DIFF
--- a/buildpack-deps/content.md
+++ b/buildpack-deps/content.md
@@ -12,3 +12,28 @@ knowing beforehand that `ssl.h` is required to build a dependent module.
 # How to use this image
 
 This stack is designed to be the foundation of a language-stack image.
+
+## What's included?
+
+The main tags of this image are the full batteries-included approach.  With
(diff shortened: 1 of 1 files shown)


buildpack-deps/content.md
```diff
@@ -12,3 +12,26 @@ knowing beforehand that `ssl.h` is required to build a dependent module.
 # How to use this image
 
 This stack is designed to be the foundation of a language-stack image.
+
+## What's included?
+
+The main tags of this image are the full batteries-included approach.  With
+them, any arbitrary `gem install` should work.  For some language stacks, that
```
tianon: This should be "the majority of", not "any".  If we go as broad as "any", we're wide open for including the entire Debian archive. :+1:

tianon: Updated with better wording.

yosifkit:
LGTM

PR #142, opened 2015-01-23
docker-library/php versions bumped

rogeriopradoj:
docker-library/php@c871678 and docker-library/php#65

PR #143, opened 2015-01-23
Bump php fpm to 5.6 in order to enforce docs with latest php version

rogeriopradoj:
docker-library/php@c871678 and docker-library/php#65


rogeriopradoj:
Sorry. The correct pull request is on #142 .

Thanks in advance!

PR #152, opened 2015-02-06
Add documentation for WORDPRESS_DB_HOST

md5:
This goes with https://github.com/docker-library/wordpress/pull/55

### DIFF
--- a/wordpress/content.md
+++ b/wordpress/content.md
@@ -39,6 +39,9 @@ Then, access it via `http://localhost:8080` or `http://host-ip:8080` in a
 browser.
 
 If you'd like to use an external database instead of a linked `mysql` container,
-specify the hostname and port with `WORDPRESS_DB_HOST`:
+specify the hostname and port with `WORDPRESS_DB_HOST` along with the password
+in `WORDPRESS_DB_PASSWORD` and the username in `WORDPRESS_DB_USER` (if it is
+something other than `root`):
 
-    docker run --name some-wordpress -e WORDPRESS_DB_HOST=10.1.2.3:3306 -d wordpress
+    docker run --name some-wordpress -e WORDPRESS_DB_HOST=10.1.2.3:3306 \
+        -e WORDPRESS_DB_USER=... -e WORDPRESS_DB_PASSWORD=... -d wordpress
--- a/wordpress/content.md
+++ b/wordpress/content.md
@@ -19,15 +19,16 @@ Indonesia.
 The following environment variables are also honored for configuring your
 WordPress instance:
 
+ - `-e WORDPRESS_DB_HOST=...` (defaults to the IP and port of the linked `mysql` container)
  - `-e WORDPRESS_DB_USER=...` (defaults to "root")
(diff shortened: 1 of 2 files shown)


wordpress/content.md
```diff
@@ -36,3 +37,8 @@ container's IP, standard port mappings can be used:
 
 Then, access it via `http://localhost:8080` or `http://host-ip:8080` in a
 browser.
+
+If you'd like to use an external database instead of a linked `mysql` container
```
md5: I just noticed there should probably be a comma at the end of this line (after "container"). Should I amend this commit?

tianon: :+1:

tianon: :heart:

wordpress/content.md
```diff
@@ -36,3 +37,8 @@ container's IP, standard port mappings can be used:
...
+specify the hostname and port with `WORDPRESS_DB_HOST`:
+
+    docker run --name some-wordpress -e WORDPRESS_DB_HOST=10.1.2.3:3306 -d wordpress
```
yosifkit: Do we want to note that they will need `WORDPRESS_DB_PASSWORD` if they use an external database (since there is no linked ENV vars)?

yosifkit: (Possibly even `WORDPRESS_DB_USER`, but they should be smart enough to know to specify when the sql user is not `root`)

md5: Sure. I'll do that with the comma change.

md5:
Updated to add a note about username and password.

yosifkit:
LGTM

tianon:
LGTM, although we'll want to get a PR to the official-images repo open before we push this up :+1:

md5:
@tianon :notebook_with_decorative_cover: noted

tianon:
@md5 want the honor of making the PR?  It's as easy as cloning both the
official-images repo and the wordpress repo somewhere, then running
"./generate-stackbrew-library.sh > ../official-images/library/wordpress"
and making a PR out of it. :wink:

md5:
Sure. I was actually playing around with that last night for the Jetty image.

md5:
Should it be showing https://github.com/docker-library/wordpress/commit/dc184cf9712bd66c1836e8ecf94987a742c29eed?

It looks like https://github.com/docker-library/wordpress/commit/aac1d2ed4245b54550e9a64a4e5951c9b891b9eb is the tip of `master`.

md5:
It looks like [`git log -1 -- .`](https://github.com/docker-library/wordpress/blob/master/generate-stackbrew-library.sh#L10) doesn't include the merge commit because it didn't touch any files in `.`.

md5:
Perhaps `git show --format='format:%H'` would be appropriate instead of `git log -1 --format='format:%H' -- .`?

tianon:
Naw, the commit it's showing is correct.  We don't care about the merge
commit for the purposes of the official-images repo, since it didn't
actually change the contents.

PR #158, opened 2015-02-07
Issue 157 Fix for Mac

### LINKED ISSUE #157: IRSSI docs need update for Mac users
While running this image I ran into an issue on my Mac with the `localtime` mount.  Searched web to find this is no longer needed on Mac.

Talking to @jfrazelle confirmed that the command line is Linux specific. The docs should be updated for Mac use --- I'm planning on doing that shortly.

moxiegirl:
I added  a fix for the Mac command line.  @jfrazelle left the uid stuff in there cause it doesn't seem to be making my Mac sick.

Tweaked some copy because I wanted to practice a bit my markdown and because technical writer. 

Fixes #157


yosifkit:
Since `README.md` is generated, can you move these changes to `content.md`?  Also, can you flow the lines at 80 columns?  One other note, any changes to the `Contributing` section would need irssi to have its own [`user-feedback.md`](https://github.com/docker-library/docs#image-nameuser-feedbackmd), unless you think we need global changes there :smile:.

moxiegirl:
Well, hit me with a stick! Didn't realize that.  Fixed.  Hope I have the wrap right; I had it soft wrapped at 80 and just applied a hard wrap at 80 --- which might not have the desired effect.  

I was a bit confused on my commits if I needed the full sig ETC as committing to the docker repo.  That seems to be missing from the guidelines.  I'm not sure I have the bandwidth to make a full on contribute to Contributing at the mo. So...

tianon:
Looking at https://github.com/docker-library/docs/pull/158/files, this diff is really messy and changing lots of unrelated formatting unnecessarily.

Are you using some kind of Markdown editor that's messing so heinously with the formatting?

Also, 7 commits for this seems a little bit overboard -- can you squash when you're finished? :smile:

moxiegirl:
I was using Text Wrangler now, I'm using BBEdit.  I can switch to emacs if need be... You using git diff to see this I'm assuming...

Heinously! Great word.

7 commits  -- the horror! Others can see how often I make a typo. Yes. I will. Flatter than a pancake. Back shortly.

moxiegirl:
@yosifkit How is this?

irssi/content.md
```diff
@@ -1,36 +1,44 @@
 # What is irssi?
-
-Irssi is a terminal based IRC client for UNIX systems. It also supports SILC
-and ICB protocols via plugins. Some people refer to it as 'the client of the
+#
```
tianon: Little things like this `#` are what I was referring to -- there's a couple more minor changes like this in this same file.

moxiegirl:
Ah, ok, new editor was auto adding in these and I was blind to it.  Fixed.

yosifkit:
LGTM

tianon:
Seems fair, thanks!

LGTM

PR #161, opened 2015-02-10
update to centos documentation 

jperrin:
separates rolling builds and minor tags for better clarity. 
Added documentation about the `nodocs` yum option for users reporting missing files after installing packages. 

Fixes https://github.com/docker-library/official-images/issues/478


yosifkit:
LGTM

tianon:
LGTM

PR #162, opened 2015-02-10
Add "--no-prompt" flag to "push.pl"

tianon:
This skips the "vimdiff" invocation so that we can update directly if we know what the changeset is going to be ahead of time.

### DIFF
--- a/push.pl
+++ b/push.pl
@@ -4,6 +4,7 @@ use warnings;
 use 5.010;
 use open ':encoding(utf8)';
 
+use File::Basename qw(fileparse);
 use File::Temp;
 use Getopt::Long;
 use Mojo::UserAgent;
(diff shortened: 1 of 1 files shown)


yosifkit:
This seems very dangerous and will wipe-all-the-things without any feedback.

tianon:
Indeed.  The problem I'm really solving is that `vimdiff` takes way too long to come up, and then I basically have to type `]cdo:wq` or `:%diffget\n:wq` repeatedly for every single image, and my fingers are sad.

Maybe a better solution would be to instead have a way to show a simple `diff -U` output with a simple `y/n` prompt to apply the changes?

yosifkit:
That seems better. Maybe make it 3 options so that you can jump in vimdiff in case you want part of it.

tianon:
Closing in favor of something new I'm working on.

PR #163, opened 2015-02-12
markdownfmt

tianon:
I've split this into separate commits to make it easier to review.

- 34b38dc is the "add markdownfmt to Travis" bits
- a71fa24 is where I actually format all the files
- 1c92655 is just `update.sh`
- b8a291b is where I update the documentation to reference `markdownfmt`


tianon:
Rebased to include `oraclelinux` in the `update.sh` commit

yosifkit:
LGTM

PR #165, opened 2015-02-18
Use official image for push script and specify mojo version

yosifkit:
Since we have a monkey patch on `Mojo::UserAgent::CookieJar` in push.pl (which broke recently), we need to specify a Mojolicious version.  And we should use an official image directly to show how easy it is.

Also, swap to entrypoint to make it easier to run.


tianon:
LGTM

PR #166, opened 2015-02-22
Docs for new Alpine Linux image

andyshinn:
These are the associated docs for the new Alpine Linux image (https://github.com/docker-library/official-images/pull/516). It looks like the `update.sh` script expects stuffs to already be committed to build so I didn't add the `README.md`.


yosifkit:
LGTM

alpine/content.md
```diff
@@ -0,0 +1,30 @@
+# What is Alpine Linux?
+
+[Alpine Linux](http://alpinelinux.org/) is a Linux distribution built around [musl libc](http://www.musl-libc.org/) and [BusyBox](http://www.busybox.net/). The image is only 5 MB in size and has access to a [package repository](http://forum.alpinelinux.org/packages) that is much more complete than other BusyBox based images. This makes Alpine Linux a great image base for utilities and even production applications. [Read more about Alpine Linux here](https://www.alpinelinux.org/about/) and you can see how their mantra fits in right at home with Docker images.
+
```
tianon: Can you add `%%LOGO%%` on a line by itself here so we can show off that perty logo? :+1:

alpine/content.md
```diff
@@ -0,0 +1,30 @@
...
+## Usage
+
+Stop doing this:
+
+	FROM ubuntu:14.04
+	RUN apt-get update -q \
+	  && DEBIAN_FRONTEND=noninteractive apt-get install -qy mysql-client \
+	  && apt-get clean \
+	  && rm -rf /var/lib/apt
+	ENTRYPOINT ["mysql"]
```
tianon: Can we change this to be more in line with best practices?  I know it's a negative example, but I'd like to avoid even bad negative examples in our docs. :+1:

```Dockerfile
FROM ubuntu:14.04
RUN apt-get update \
    && apt-get install -y mysql-client \
    && rm -rf /var/lib/apt/lists/*
ENTRYPOINT ["mysql"]
```

alpine/content.md
```diff
@@ -0,0 +1,30 @@
...
+
+This took 19 seconds to build and yields a 164 MB image. Eww.
```
tianon: I'm not sure where this size comes from -- is this supposed to be the size of the additional layers?  The size of the total image?

Also, I'm not sure where "19 seconds" comes from either.  Is that with a primed apt-cacher-ng instance?

I'm not opposed to the highly-subjective "Eww", but the facts leading to it aren't exactly accurate, and I think more accurate (or at least actually verifiable) facts will add even more punch to the "Eww". :+1:

alpine/license.md
```diff
@@ -0,0 +1 @@
+View [license information](https://github.com/gliderlabs/docker-alpine/blob/master/LICENSE) for the software contained in this image.
```
tianon: Is this the license for the software contained in the image (ie, [Alpine Linux]()), or the license for the `Dockerfile` etc used to build the image? :wink:

alpine/content.md
andyshinn: Sure thing. Missed that part..

alpine/license.md
andyshinn: You are correct, that license was for the build script we made (someone actually requested the licensing terms for it so we added one). Looking in to it more, I can't actually find a license for Alpine Linux. Seems strange that there wouldn't be one. I'll dig a little more.

alpine/content.md
andyshinn: Yea, totally. This was probably just my muscle memory.

alpine/content.md
andyshinn: Do you think just clarifying that it is the size of the total image and that the builds were using boot2docker on my MacBook Pro would be enough? If you think this section is too arbitrary, I'd probably lean more to just remove the comparison and put a more traditional usage. At this point, users coming to the image were probably already looking for size benefits.

tianon: Well, I ran the same build and got an image that was much larger in total "virtual size", and it took at least a minute to build.

alpine/license.md
andyshinn: This is actually going to be multiple licenses for the major components of Alpine (musl and BusyBox). Is the license file required or should it just be removed?

tianon: Meh, might as well just remove it for now and revisit later. :+1:

alpine/content.md
andyshinn: Yea, same here. It is closer to 232 MB. Maybe I just did my test wrong when I originally wrote that...

andyshinn:
Some updates per comments.

tianon:
LGTM :heart: :+1:

PR #167, opened 2015-02-23
Fix downloading older Mojolicious version

tianon:
Turns out BackPAN doesn't have SSL, so we can't have fully secure `cpanm` _and_ have old versions of our dependencies. :cry:


yosifkit:
LGTM

PR #172, opened 2015-02-24
Initial onbuild/slim description generation

yosifkit:
Generated one example with python and tested an `./update.sh mongo` to make sure there was no change.

Still needs :heart:.


.template-helpers/variant.md
```diff
@@ -0,0 +1,7 @@
+# Image Variants
+
+The `%%REPO%%` images come in three flavors, each designed for a specific use case.
+
+`<version>`:
+
+This is the defacto image. If you are unsure about what your needs are, you probably want to use this one. It is designed to be used both as a throw away container (mount your source code and start the container to start your app), as well as the base to build other images off of. This tag is based off of [`buildpack-deps`](https://registry.hub.docker.com/_/buildpack-deps/). `buildpack-deps` is designed for the average user of docker who has many images on thier system. It, by design, has a large number of extremely common debian packages. This reduces the number of packages that images that derive from it need to install, thus reducing the total size of all of your docker images on your system.
```
tianon: `s/thier/their/`

tianon: `s/debian/Debian/`

tianon: "all of your docker images on your system" is a bit too much "your"

How about just "... thus reducing the overall size of all images on your system" ?

.template-helpers/variant.sh
```diff
@@ -0,0 +1,43 @@
...
+dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+
+IFS=$'\n'
+lines=( $(curl -sSL 'https://raw.githubusercontent.com/docker-library/official-images/master/library/'"$repo" | grep -vE '^$|^#') )
+unset IFS
+
+tags=()
+# this loop finds all the "variants", making assumptions about what they can contain
+for line in "${lines[@]}"; do
+	tag="$(echo "$line" | awk -F ': +' '{ print $1 }' | sed -r 's/[0-9.-]+//g')"
```
tianon: Since `variant-onbuild.md` assumes `%%REPO%%:onbuild` is a valid tag, and since all current repos have that property, maybe we should just simplify this to grabbing all the tags of the repo as-is without the `sed`?  It should be pretty harmless to be checking for, for example, `variant-3.4-onbuild.md` to exist, since we'll eventually get down to just `variant-onbuild.md` which does exist.  Thoughts?

.template-helpers/variant.sh
```diff
@@ -0,0 +1,43 @@
...
+if [ -z "$repo" ]; then
+	echo >&2 "usage: $0 repo"
+	echo >&2 "   ie: $0 hylang"
+	exit 1
+fi
+
+dir="$(dirname "$(readlink -f "$BASH_SOURCE")")"
+
+IFS=$'\n'
+lines=( $(curl -sSL 'https://raw.githubusercontent.com/docker-library/official-images/master/library/'"$repo" | grep -vE '^$|^#') )
```
tianon: Technically, we could just directly `... | grep ... | cut -d':' -f1 | sort -u` here, right?

.template-helpers/variant.sh
```diff
@@ -0,0 +1,43 @@
...
+for tag in "${tags[@]}"; do
+	if [ -f "$dir/variant-${tag}.md" ]; then
+		text+=$'\n' # give a little space
+		# because parameter expansion eats the trailing newline
+		text+="$(<"$dir/variant-${tag}.md")"$'\n'
+	fi
+done
+if [ "$text" ]; then
+	# because parameter expansion eats the trailing newline
+	text=$'\n\n'"$(<"$dir/variant.md")"$'\n'"$text"
```
tianon: At this point, why not just skip the `echo` outside the `if` and do something like:
```bash
if [ "$text" ]; then
    echo
    cat "$dir/variant.md"
    echo -n "$text"
fi
```

.template-helpers/variant.md
```diff
@@ -0,0 +1,7 @@
+# Image Variants
+
+The `%%REPO%%` images come in three flavors, each designed for a specific use case.
+
+`<version>`:
```
tianon: Looking at the rendered version below, I think these will match the rest of the `README` better if we use headers, and if we qualify them with `%%REPO%%` it'll be more clear exactly what we're talking about:
```markdown
# Image Variants

...

## `%%REPO%%:<version>`

...

## `%%REPO%%:onbuild`

...

## `%%REPO%%:slim`

...
```

.template-helpers/variant.md
yosifkit: That sounds good to me.

.template-helpers/variant.sh
yosifkit: I think that is a case of premature optimization by me to cut things from the later loop.  I think I'll just try the piping and skip this loop entirely (technically it will live on in the pipes).

.template-helpers/variant.sh
yosifkit: So how about this and drop the first loop?  Keep the `sed` or not?
```bash
tags=( $(curl -sSL 'https://raw.githubusercontent.com/docker-library/official-images/master/library/'"$repo" | grep -vE '^$|^#' | cut -d':' -f1 | sed -r 's/[0-9.-]+//g' | sort -u) )
```

tianon: Seems good to me, and I don't see a compelling reason to keep the `sed` personally, since it just means a couple extra tags to loop over and noop.

tianon: `-f` is super fast :+1:

yosifkit:
@tianon, updated!

tianon:
LGTM!  Let's do an `update.sh` run across ALL THE THINGS and see if we get any false positives that make our section about `buildpack-deps` a lie! :smiling_imp:

.template-helpers/variant-onbuild.md
```diff
@@ -0,0 +1,3 @@
+## `%%REPO%%:onbuild`
+
+This image makes building derivitative images easier. For most use cases, creating a `Dockerfile` in the base of your project directory with the line `FROM %%REPO%%:onbuild` will be enough to create a stand-alone image for your project.
```
md5: "derivative"

yosifkit: Thanks, fixed!

wblankenship:
Is this good to go? If not, is there anyway I can help this ship?

tianon:
@wblankenship the problem we currently have is that it's not always accurate as-is, so we need either a good way to detect whether the image we're applying it to _is_ actually based on `buildpack-deps`, or to figure out a way to reword (possibly by moving the bulk of the language directly into the `slim` section) so that the main section only describes that "variants exist", with less detail about them, since not all images that have "slim" or "onbuild" are actually `FROM buildpack-deps`

wblankenship:
@tianon are there constraints on `update.sh`? For example:
* it has to be able to be run when the machine has no internet access
* it can only use content from the _docker-library/docs_ repo?

wblankenship:
How does https://github.com/infosiftr/docker-library-docs/pull/2 look?

.template-helpers/variant.md
```diff
@@ -0,0 +1,7 @@
+# Image Variants
+
+The `%%REPO%%` images come in many flavors, each designed for a specific use case.
```
Starefossen: I would argue that this is *one* image with many flavors, hence: "The `%%REPO%%` image come in many flavors".

tianon: IMO that's kind of deep in the semantics of the language, so I'd prefer we err on the side of pointing out and reinforcing to users that this really isn't just "one `ruby` image", but is in fact a collection of multiple related images.

As a concrete example, `ruby:slim` and `ruby:onbuild` are _very_ different images, and they don't even share a common ancestor besides `debian` itself.

tianon:
@wblankenship good enough for now, thanks :+1:

.template-helpers/variant.md
tianon: Hmm, we might actually already have examples of this being "something something `%%REPO%%` image" (singular); requires more thought.

PR #173, opened 2015-02-26
Add some official docker-compose.yml scaffolding

tianon:
This includes an example `docker-compose.yml` for WordPress.


yosifkit:
@md5, we would love to use the ``` code block instead of the indent code block but the rendering on the docker hub leaves [much](https://registry.hub.docker.com/u/tianon/rawdns/) to be desired. (vs [github](https://github.com/tianon/rawdns)).

yosifkit:
LGTM

PR #175, opened 2015-02-28
Document use of `setuid` module and user `jetty`

md5:


tianon:
LGTM

md5:
I just noticed that `JETTY_STATE` was still referring to `/usr/local/jetty`, so I've updated that.

I should probably change that explicitly to `/var/run/jetty` and make that a volume in another PR, but updating the docs to be correct is good enough for now.

md5:
@yosifkit Could you PTAL? Looks like the updated Jetty image from docker-library/official-images#525 got pushed out yesterday, but not the updated docs.

Thanks! :bow:

yosifkit:
LGTM

PR #176, opened 2015-03-03
[IMP] Custom configuration and mounted addons

aab-odoo:
Hi,

We did an update of the doc:
* How to run Odoo with a custom configuration by specifying the arguments inline or using the configuration file
* How to mount custom addons within the container
* Fix the path to filestore

Regards,


yosifkit:
The content here is fine.:+1:  But this is based on an old commit, can you rebase it to master so that we can merge it cleanly?  (There have been some significant changes like adding `markdownfmt` to keep things consistent, which also drops the 80 column requirement)

PR #188, opened 2015-03-19
Updated update.sh for new Oracle Github URL

Djelibeybi:
Also, regenerated the Oracle Linux README.md.

This will need to be regenerated again when the Oracle Linux 7.1 image is pulled into the library.


yosifkit:
LGTM

PR #192, opened 2015-03-24
Logstash docs

### LINKED ISSUE #186: logstash: No docs
I just saw this comment on Docker Hub:

> For an official repo, this seems to be lacking in documentation.

cc/ @willroden

Moghedrin:
Fixes #186


logstash/content.md
```diff
@@ -0,0 +1,21 @@
...
+
+%%LOGO%%
+
+# How to use this image
+
+## Start Logstash with commandline configuration
+
+If you need to run logstash with configuration provided on the commandline, you can use the logstash image as follows:
+
+`docker run -it --rm logstash logstash -e 'input { stdin { } } output { stdout { } }'`
```
tianon: Can we use an indented code block here instead of an inline one?  The rendering on the Hub will be less strange. :+1:

tianon:
Content LGTM, but I think you should rebase, especially with that merge commit in there. :smile:

Something like this should do the trick:
```console
$ git fetch https://github.com/docker-library/docs.git master
$ git rebase -i FETCH_HEAD
 ... editor opens, change all but the first line's first word to "s"
```
(followed by a force push to the branch)

yosifkit:
LGTM

tianon:
LGTM :+1:

PR #193, opened 2015-03-24
Documentation how to add extensions

suchja:
Based on the discussion in docker-library/wordpress#68 the process to add php-extensions was added. Focus is to make it as easy as possible for the user of the image to create and maintain its own extended image based on docker-library/wordpress.

Not sure whether it makes sense to restrict this description to the wordpress image. Maybe it makes sense to at least partially add the description to all images in docker-library.

However, this should be done once we agree on the documentation for wordpress.


yosifkit:
LGTM

tianon:
LGTM

yosifkit:
Thanks for the contribution.

PR #200, opened 2015-04-06
MySQL custom config files must end in .cnf

lorin:
Document that custom config files in /etc/mysql/conf.d must end in .cnf. Otherwise, MySQL will ignore the files.

In particular, the /etc/mysql/my.cnf file inside of the container has the following lines:

```
# * IMPORTANT: Additional settings that can override those from this file!
#   The files must end with '.cnf', otherwise they'll be ignored.
#
!includedir /etc/mysql/conf.d/

```

### DIFF
--- a/mysql/content.md
+++ b/mysql/content.md
@@ -44,13 +44,13 @@ The MySQL Server log is available through Docker's container log:
 
 ## Using a custom MySQL configuration file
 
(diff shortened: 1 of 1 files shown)


yosifkit:
LGTM

tianon:
LGTM

cc @yngvesv

yngvesv:
Yes, good catch.

PR #209, opened 2015-04-16
examples of volume mapping

booyaa:


yosifkit:
LGTM

yosifkit:
ping @tianon

ghost/content.md
```diff
@@ -17,3 +17,11 @@ If you'd like to be able to access the instance from the host without the contai
 	docker run --name some-ghost -p 8080:2368 -d ghost
 
 Then, access it via `http://localhost:8080` or `http://host-ip:8080` in a browser.
+
+You can also point the image to your existing content on your host
+
+	docker run --name some-ghost -v /path/to/ghost/blog:/var/lib/ghost ghost
+
+...or a data container
```
md5: How about making this a complete sentence and making it clear that the data container will need to have a `/var/lib/ghost` mount?

booyaa:
@md5 good call

ghost/content.md
```diff
@@ -17,3 +17,13 @@ If you'd like to be able to access the instance from the host without the contai
 	docker run --name some-ghost -p 8080:2368 -d ghost
 
 Then, access it via `http://localhost:8080` or `http://host-ip:8080` in a browser.
+
+You can also point the image to your existing content on your host:
+
+	docker run --name some-ghost -v /path/to/ghost/blog:/var/lib/ghost ghost
+
+Alternatively you can use a data container that points to ```/var/lib/ghost```:
```
yosifkit: just a single set of ` is needed (which is probably why markdownfmt is "failing" on the build).

yosifkit:
Just my one comment and then LGTM

ghost/content.md
```diff
@@ -17,3 +17,13 @@ If you'd like to be able to access the instance from the host without the contai
...
+
+	docker create --name some-ghost-data -v /var/lib/ghost ghost /bin/true
```
md5: If the image being used to create the data container is `ghost`, it already has a `/var/lib/ghost` volume in the Dockerfile. Not sure if this will confuse users or not.

I've generally seen people create "data containers" with images like `tianon/true`, in which case you *would* need to declare the volume yourself as you've done in your `docker create` command.

md5: I'd probably leave the data container creation out of the docs, just mention that you need a `/var/lib/ghost` volume, and link the words "data container" to here: http://docs.docker.com/userguide/dockervolumes/

booyaa:
argh tried removing the extra backticks, still failing to build. any ideas? also simplified the data container bit.

md5:
The docs changes LGTM :+1:

booyaa:
Any idea why Travis CI is still failing?

yosifkit:
I think it might be the empty blank line at the end.

booyaa:
KERPOW! cheers @yosifkit

tianon:
LGTM

PR #218, opened 2015-04-26
Added postgre example in doc

pheanex:
Could be helpful to others (at least was to me)
http://stackoverflow.com/questions/26598738/how-to-create-user-database-in-script-for-docker-postgres

Maybe you also want to extend the example to show how to import a whole sql-file into a specific database

### DIFF
--- a/postgres/README.md
+++ b/postgres/README.md
@@ -49,7 +49,14 @@ This optional environment variable is used in conjunction with `POSTGRES_PASSWOR
 
 # How to extend this image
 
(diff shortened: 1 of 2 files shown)


yosifkit:
Looks like markdownfmt is grumpy about the `\t\tSQL stuff;` and wants `\t\s\s\s\sSQL command`.  Otherwise, LGTM.

pheanex:
i guess the failed checks are not my fault?

PR #219, opened 2015-04-28
Add initial drupal documentation

tianon:
See https://github.com/docker-library/official-images/pull/689 for the image addition.

PR #221, opened 2015-04-28
Point out that wordpress's own Dockerfile has an "extra libraries" example

tianon:
As added in https://github.com/docker-library/docs/pull/219 for `drupal`.


yosifkit:
LGTM

PR #229, opened 2015-05-13
Add Couchbase Server Official Docker image

ceejatec:


couchbase/content.md
```diff
@@ -0,0 +1,258 @@
...
+
+There are several deployment scenarios which this Docker image can easily support. These will be detailed below, along with recommended network arrangements for each.
+
+## Volumes
+
+A Couchbase Server Docker container will write all persistent and node-specific data under the directory `/opt/couchbase/var`. We recommend mapping this directory to a directory on the host filesystem (using the `-v` option to `docker run`) for the following reasons:
+
+* **Persistence** Storing `/opt/couchbase/var` outside the container allows you to delete the container and re-create it later. You can even update to a container running a later point release of Couchbase Server without losing your data.
+* **Performance** In a standard Docker environment using a union filesystem, leaving `/opt/couchbase/var` "inside" the container will result in some amount of performance degradation.
```
yosifkit: Since this directory is declared as a `VOLUME` in the `Dockerfile`, it will actually be automounted to a bind mount on the host by docker, just not in a place easily known to the user. But a user could just use `volumes-from` to migrate:
```console
$ docker run -d --name my-couchbase couchbase
$ # use database
$ docker stop my-couchbase
$ docker run -d --name new-couchbase --volumes-from my-couchbase couchbase
$ docker rm my-couchbase
```

couchbase/content.md
```diff
@@ -0,0 +1,258 @@
...
+│  └─────────────────┘  │                                                      
+└───────────────────────┘                                                      
+```
+
+This is a quick way to try out Couchbase Server on your own machine with no installation overhead - just *download and run*. In this case, any networking configuration will work; the only real requirement is that port 8091 be exposed so that you can access the Couchbase Admin Console.
+
+**Start the container**
+
+```
+docker run -d -v ~/couchbase:/opt/couchbase/var -p 8091:8091 couchbase/server
```
yosifkit: `s!/server!!` --will probably need to check the rest of the doc to update to the "official image" :smile:.

yosifkit: If you like, you could just add a `--name my-couchbase` and then the "Verify container start" section will be slightly simpler: `docker logs my-couchbase`.

couchbase/content.md
```diff
@@ -0,0 +1,258 @@
...
+
+* [Community Edition](http://www.couchbase.com/community) -- free for unrestricted use.
+
+* [Enterprise Edition](http://www.couchbase.com/agreement/subscription) -- free for development, paid subscription required for production deployment.
+
+By default, the `latest` docker tag points to the latest Enterprise Edition, as does the `enterprise` docker tag.  If you want the Community Edition instead, you should add the `community` tag, which will point to the latest available Community Edition.
+
+```
+docker run couchbase/server:community
+```
```
yosifkit: Delete this "Licensing" section from `content.md` and move L254-258 to the `license.md` as that is used to generate the "License" section in `README.md` (which is why there is two setcions in `README.md`).

yosifkit:
Could you also rebase on top of master? That will allow the Travis build to show the `markdownfmt` diff to make it easy for you to get any other changes it requires.  Or just run `./markdownfmt.sh -d couchbase` to see what needs to change.

ceejatec:
I've incorporated your comments, thanks!

I've run markdownfmt.sh -d couchbase before, but I don't understand the result. It seems like it's telling me I need to change the markup into a format that GitHub won't display correctly. Is that right?

ceejatec:
Ok, so I ran markdownfmt.sh -w and commited the result, and I guess that's what Travis wanted.

Any other comments?

I see the official image itself is already up (https://registry.hub.docker.com/_/couchbase/); once this proposal here is merged, will the docs automatically appear there?

Thanks!

tianon:
LGTM

PR #230, opened 2015-05-13
Complex configuration clarification

stephenwithav:
The provided `docker cp` command will successfully copy the configuration file from the container to the host, but the `COPY` line in the `Dockerfile` example will fail to do the same.

What is the goal of the section beginning with the "If you wish to adapt the default configuration" sentence?

You could always modify it on the host, then create a new container with the newly modified configuration file.  Or you could modify the file from within the container and then `docker commit` the changes.

If I know the goal, I will happily contribute to the documentation.

### DIFF
--- a/nginx/README.md
+++ b/nginx/README.md
@@ -45,11 +45,6 @@ If you wish to adapt the default configuration, use something like the following
 
 	docker cp some-nginx:/etc/nginx/nginx.conf /some/nginx.conf
 
-As above, this can also be accomplished more cleanly using a simple `Dockerfile`:
-
-	FROM nginx
-	COPY nginx.conf /etc/nginx/nginx.conf
-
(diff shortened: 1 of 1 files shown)


yosifkit:
This would need to go in `nginx/content.md` since `nginx/README.md` is generated.  As far as content, I'll let @thresheek from Nginx make the decisions on how they would like to change the documentation.

tianon:
What's documented is basically how I personally use the `nginx` container in a lot of cases -- I don't want to always bind-mount in my config, so I use a `Dockerfile` to build the configuration into a new image, which is what this is documenting.

Using `docker cp` or even `docker run -v` are equally useful, but none is "more useful" than the others, IMO; each of them has pros and cons.

yosifkit:
Closing old PR; feel free to comment if you feel this was closed in error.

PR #235, opened 2015-05-14
Update "scratch" and "hello-world" to be more complimentary

tianon:
This also explains what goes into the sausage a little bit. :horse: :pig2: :dog2:


tianon:
Adding a second commit with `update.sh` run so this is easier to review.

yosifkit:
LGTM

PR #237, opened 2015-05-18
Document MYSQL_CHARSET environmental variable.

buckett:
This is the documentation update for docker-library/mysql#14


yosifkit:
`s/it's/its/`, move this change to the `content.md` since `README.md` is generated. 

Thanks for the PR!

yosifkit:
Can we also make sure users know that this only applies to the database created by the `$MYSQL_DATABASE` env?

buckett:
Updated to incorporate changes from @yosifkit 
Sorry for not reading the toplevel readme about the project layout.

PR #240, opened 2015-05-21
Update travis.yml to give diff of needed changes

yosifkit:


.travis.yml
```diff
@@ -10,8 +10,8 @@ install:
 script:
   - files="$(find -name '*.md' -print0 | xargs -0 markdownfmt -l)";
     if [ "$files" ]; then
-      echo >&2 "Need markdownfmt:";
-      echo >&2 "$files";
+      echo >&2 "Error: markdownfmt changes required";
+      find -name '*.md' -print0 | xargs -0 markdownfmt -d >&2;
```
tianon: Why not reuse `$files` here instead of redoing the `find`?

yosifkit:
Fixed!

tianon:
LGTM

PR #241, opened 2015-05-22
Single user mode doesn't work for many commands

SethRobertson:
You cannot have multiline commands, table creates do not seem to always work, etc.

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -41,7 +41,7 @@ This optional environment variable is used in conjunction with `POSTGRES_PASSWOR
 
 # How to extend this image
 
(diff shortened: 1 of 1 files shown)


md5:
You're right that `--single` is limited in what it can do. The main limitation I'm aware of with `--single` is that implicit array types are not created. That being said, the multiline issue can be avoided by passing the `-j` flag to `postgres --single`.

I thought there was an issue in the `docker-library/postgres` repo regarding the implicit array type think, but the only discussion I can find is for my `appropriate/postgis` image: https://github.com/appropriate/docker-postgis/issues/2#issuecomment-71470124

md5:
I believe this PR is superceeded by the changes in #291.

PR #247, opened 2015-05-28
Cassandra docs for new env vars

yosifkit:
Contains #245 with markdownfmts and some links.


yosifkit:
Needs https://github.com/docker-library/official-images/pull/769 before it is applied.

tianon:
LGTM

PR #251, opened 2015-06-02
Add more explicit description for why "onbuild" variants exist and for when they're appropriate to use (and when it's time to "graduate")

tianon:


yosifkit:
LGTM

PR #252, opened 2015-06-04
Add mongo docs for volume, storageEngine, mmap on vboxsf

### LINKED ISSUE #236: Documentation for mongo image misses data volume
In the documentation for the MongoDB image it should be mentioned, that the data is stored under /data/db

yosifkit:
fixes #236
fixes https://github.com/docker-library/mongo/issues/30


tianon:
LGTM

PR #253, opened 2015-06-04
Aerospike 3.5.12 release

volmarl:


tianon:
This will be updated on an automatic cron job after https://github.com/docker-library/official-images/pull/787 merges. :+1:

PR #254, opened 2015-06-05
Create an Official Repo for IBM DB2 Express-C

wzymaster:
We want to create a new Official IBM DB2 Express-C docker image in Docker Hub Registry. 

In fact, we already have one in our own organization at [ibmcom/db2express-c](https://registry.hub.docker.com/u/ibmcom/db2express-c/). Please feel free to check it out.

a pull request to docker-library/official-images will be submitted soon.


psftw:
The docs PR here, the linked-to docs on Docker Hub, and the actual image definition PR are inconsistent about tag naming.  There is also inconsistency about what docker versions are supported.  Please describe any kernel version and kernel configuration requirements if users need to be aware of them.  Please read the instructions at https://github.com/docker-library/docs and address the missing requirements.  I would prefer that the parts recommending machine and the link to the "what is docker" page be removed as they are not relevant.

The first suggested ``docker run`` command fails with:

    Changing password for user db2inst1.
    New password: BAD PASSWORD: The password contains the user name in some form
    Retype new password: passwd: all authentication tokens updated successfully.

At this point I found the "-bash" option and was able to start and create a sample database.  The workflow of logging in as root and then changing to the service user was awkward and is unnecessary.  I would suggest looking at other database images for ways to think about this.  The docs mention mounting a local folder into the image as ``/share``, but doesn't tell me why that would be useful for this particular image.  The actual database data appears to be written to ``/home/db2inst1``, so is that where I should mount a data volume?  The part about where the binaries are installed doesn't seem to be correct?  The docs assume the user knows how to interact with the db2express database, but it would be nice to at least provide a simple example of querying the database to verify functionality.

wzymaster:
@psftw thanks for your review.  Regarding to your concerns, please see my answers below :  

Please check REAME.md and other related document files at https://github.com/IMC3ofC/db2express-c.docker,  instead of our existing repo at __ibmcom/db2express-c__

1.  For official repo of DB2 Image,  we only support below tag along with kernel version : 
10.5.0.5-3.10.0
it's addressed in README.md at https://github.com/IMC3ofC/db2express-c.docker/blob/master/README.md
  We have mentioned the kernel requirements in the doc. above.
 
Sorry, I don't understand what inconsistency and issue with supported docker. 

2.  Regarding to logging in as root and switch to `db2inst1` user ,  it is required to start default db2 instance , which is characteristic of DB2 product itself.  However, user can always create their own instance under given user in docker later.

3.  For now, I can remove "mount volume" parts in the documents to avoid any unnecessary confusion. Originally, we are trying to help educate user how to mount an external storage to DB2 docker container for general purpose. 

4. Regarding to simple example of querying the database to verify functionality, I will update the doc at https://github.com/IMC3ofC/db2express-c.docker.

Please let me know if I can make a pull request again once I fix all the issues above ?

psftw:
The contents of this docs PR are all that matter for the purposes of proposing a new official image.  I am not going to look in other places going forward.  The reference to `10.5.0.5-3.10.0` for a tag name is inconsistent with the official-images PR which has the `-3.10.0` removed.  

> DB2 Docker images can be deployed on a Docker Engine with any flavour of Linux , as long as it has the compatible kernel-3.10.0, like CentOS 7 respectively.

Does this mean DB2 depends on 3.10+ Linux kernel in general?  Are there special OOB configuration settings that users need to know about, or simply new features in 3.10+?

> Regarding to logging in as root and switch to db2inst1 user , it is required to start default db2 instance , which is characteristic of DB2 product itself. However, user can always create their own instance under given user in docker later.

I would argue that database images should self-initialize upon the first run and not require this extra dance.  Is the current flow really a hard requirement, or could this image be changed to take care of these details? None of the other database images require these extra hoops to jump through just to get an instance up and running, so I would recommend checking out how they are created (i.e. https://github.com/docker-library/postgres).


> For now, I can remove "mount volume" parts in the documents to avoid any unnecessary confusion. Originally, we are trying to help educate user how to mount an external storage to DB2 docker container for general purpose.

It's definitely useful to include volume information, but not for the purpose of educating the user about volumes.  Instead, the image docs should educate the user about using volumes related to this particular image.  

> Regarding to simple example of querying the database to verify functionality, I will update the doc at https://github.com/IMC3ofC/db2express-c.docker.

Per my previous comment, I would like to only review things in PR form to official-images and docs repos.  You can push the latest changes to the branches that your PRs come from and it will update here (note: it won't ping me unless you add a comment as well).

wzymaster:
@psftw 

1. Sorry, I missed that in Official Image PR and just fixed it

2. Yes, DB2 Express-C Docker image is dedicatedly tested, validated with Kernel 3.10.0 Linux kernel since some features may be associated with  it.

3. Yes, definitely I can handle this by entrypoint shell script if that's OK in daemon mode. However, from the nature of db2 instance, it always creates a new user along with each db2 instance.  So it's required to switch to the instance owner to start DB2 service.  Here, we just suggest user to switch to a default db2 instance owner and start it. I don't think other DB products has the similar nature.

4. I would like to remove "mount" information for now and add it later once we have a better plan to help user to use DB2 docker with external storage.

wzymaster:
@psftw 
I have updated everything at https://github.com/IMC3ofC/db2express-c.docker and please review it !

psftw:
@wyzmaster I'd be happy to look over an updated PR here, which would require commits to your master branch of https://github.com/IMC3ofC/docs.

wzymaster:
@psftw  I just commit and please check it out.

psftw:
Closing since the effort has stalled.

PR #261, opened 2015-06-25
Aerospike 3.5.14

volmarl:


yosifkit:
We actually have an hourly cron script now to regenerate the `README` files, so this'll happen automatically once it merges in `official-images`.

tianon:
https://github.com/docker-library/docs/pull/220#issuecomment-97245587 :wink:

PR #263, opened 2015-06-29
Add Bonita BPM Official Docker image

JeremJR:
https://github.com/docker-library/official-images/pull/852


bonita/content.md
```diff
@@ -0,0 +1,51 @@
...
+
+![logo](https://github.com/bonitasoft/docker/blob/master/bonita/7.0.0/logo.png?raw=true)
+
+# How to use this image
+
+## Quick start
+
+	docker run --name bonita -d -p 8080:8080 bonita
+
+This will start a container running the [Tomcat Bundle](http://documentation.bonitasoft.com/tomcat-bundle-1) with Bonita BPM Engine + Portal. As you didn't sepecify any environment variables it's almost like if you have launched the Bundle on your host using startup.{sh|bat} (with security hardening on REST and HTTP APIs, cf Security part). It means that Bonita BPM uses a H2 database here.
```
psftw: "sepecify"

bonita/content.md
```diff
@@ -0,0 +1,51 @@
...
+
+You can access to the portal on http://localhost:8080/bonita and login using the default credentials : install / install
+
+## Link Bonita BPM to a database
+
+### MySQL
+
+	docker run --name mydbmysql -e MYSQL_ROOT_PASSWORD=mysecretpassword -d bonitasoft/mysql
```
psftw: We can't recommend using non-official images.  I see that ``bonitasoft/mysql`` is just a minor configuration change difference from the official ``mysql`` image.  I think we are stuck providing users with instructions on how to deal with this situation.  The best option is likely to defer to the instructions in the ``mysql`` image for mounting in a custom configuration file and specifying what should be changed.

bonita/content.md
```diff
@@ -0,0 +1,51 @@
...
+	docker run --name bonita_mysql --link mydbmysql:mysql -d -p 8080:8080 bonita
+
+### PostgreSQL
+
+	docker run --name mydbpostgres -e POSTGRES_PASSWORD=mysecretpassword -d bonitasoft/postgres
```
psftw: The same basic idea applies here as well, except there is the added complication that the config file doesn't exist until the initial startup which generates it.  The workaround for this kind of customization is covered in the ``postgres`` image docs and involves mounting in a shell script to modify the postgresql.conf after it is generated, but before the database is started.  Let me know if you run in to trouble.

bonita/content.md
```diff
@@ -0,0 +1,51 @@
...
+	docker run --name bonita_postgres --link mydbpostgres:postgres -d -p 8080:8080 bonita
+
+## Modify default credentials
+
+	docker run --name=bonita -e "TENANT_LOGIN=tech_user" -e "TENANT_PASSWORD=secret" -e "PLATFORM_LOGIN=pfadmin" -e "PLATFORM_PASSWORD=pfsecret" -d -p 8080:8080 bonita
+
+If you do so, you can access to the portal on http://localhost:8080/bonita and login using : tech_user / secret
+
+# Security
```
psftw: hmm... the documentation at http://documentation.bonitasoft.com/bos-version/bonita-bpm-70 looks fantastic, and perhaps it would be useful to reference that and talk about how this image can be customized in similar ways outside of the environment variables mentioned here?  As a new Docker user who may have previous experience with Bonita (or vice versa), it would be useful to get a sense of what the options were for making my own custom derived image if that would be useful?  Maybe this is overkill and not something you'd want to support.  I'll defer to your expertise.

psftw:
The links point to older versions of the documentation.  The topic of managing persistent state, especially for an application like this which may be critical to a business, should be addressed.  It can be confusing for new Docker users, and is dependent on how the image works and how volumes are utilized, so providing some guidance will go a long way.

bonita/content.md
JeremJR: thank you, I've fixed that

bonita/content.md
JeremJR: Ok, thanks. I've updated accordingly the documentation.

bonita/content.md
JeremJR: Ok, thanks. I've updated accordingly the documentation.

bonita/content.md
JeremJR: I've added an "How to extend this image" in the similar way that on the PostgreSQL container.

JeremJR:
Thanks, I've updated the links and added a "Where to Store Data" section and some guidance on how to migrate from an earlier version of Bonita BPM.

bonita/content.md
```diff
@@ -0,0 +1,273 @@
...
+
+		docker inspect --format '{{ .NetworkSettings.IPAddress }}' mydbpostgres
+		172.17.0.26
+
+5.	Dump the database
+
+		export PGPASSWORD=mysecretpassword
+		pg_dump -O -x -h 172.17.0.26 -U postgres bonitadb > /tmp/bonitadb.sql
+
+	Note that businessdb won't be updated with the migration tool but you may wan't to also backup/move it.
```
psftw: wan't -> want

JeremJR: thank you, it's updated

bonita/content.md
```diff
@@ -0,0 +1,276 @@
...
+[Set max_prepared_transactions to 100](http://documentation.bonitasoft.com/database-configuration-business-data-1):
+
+	mkdir -p ~/Documents/Docker/Volumes/custom_postgres
+	echo '#!/bin/bash' > ~/Documents/Docker/Volumes/custom_postgres/bonita.sh
+	echo 'sed -i "s/^.*max_prepared_transactions\s*=\s*\(.*\)$/max_prepared_transactions = 100/" "$PGDATA"/postgresql.conf' >> ~/Documents/Docker/Volumes/custom_postgres/bonita.sh
+	chmod +x ~/Documents/Docker/Volumes/custom_postgres/bonita.sh
+
+Mount that directory location as /docker-entrypoint-initdb.d inside the PostgreSQL container:
+
+	docker run --name mydbpostgres -v ~/Documents/Docker/Volumes/custom_postgres/:/docker-entrypoint-initdb.d -e POSTGRES_PASSWORD=mysecretpassword -d postgres:9.3
```
yosifkit: Should probably simplify this local directory in this and previous lines since `~/` may not expand properly.  I think just `mkdir custom_postgres` and then this run would be `-v "$PWD"/custom_postgres/:/docker-entrypoint-initdb.d`.  What do you think?

Could even skip the `mkdir` and just `-v "$PWD"/bonita.sh:/docker-entrypoint-initdb.d/bonita.sh`.

yosifkit:
For the code blocks, you can move to a fenced code block (\`\`\`dockerfile or ```console) if you want specific language highlighting.  The new hub actually renders them nicely.

JeremJR:
@yosifkit Ok thanks, I've simplified the paths

yosifkit:
LGTM, pending https://github.com/docker-library/official-images/pull/852

tianon:
LGTM

PR #265, opened 2015-07-01
Adding Gazebo Docs

ruffsl:
See main PR [here](https://github.com/docker-library/official-images/pull/855).


gazebo/content.md
```diff
@@ -0,0 +1,30 @@
+# What is [Gazebo](http://www.gazebosim.org/)?
+
+Robot simulation is an essential tool in every roboticist's toolbox. A well-designed simulator makes it possible to rapidly test algorithms, design robots, and perform regression testing using realistic scenarios. Gazebo offers the ability to accurately and efficiently simulate populations of robots in complex indoor and outdoor environments. At your fingertips is a robust physics engine, high-quality graphics, and convenient programmatic interfaces. Best of all, Gazebo is free with a vibrant community.
+
+> [wikipedia.org/wiki/Gazebo_simulator ](https://en.wikipedia.org/wiki/Gazebo_simulator)
```
psftw: dead link

psftw:
After looking around the docs a bit, and the discussion on the image PR, what's included here today is just the gzserver component, which is a headless simulator, and then I need a gzclient (which depends on hardware accelerated graphics through OGRE) to be able to "see" what's going on.  I'm guessing there are still useful things that can be done without firing up the client?  It would be great to provide more context about use cases for those of us that are new to Gazebo.

It looks like my OS has a package for gazebo 5.1.0, and it connected to my gzserver container when it was running (and failed "Waiting on master" otherwise).  Because of the way the gzclient works, I don't think we would be able to support it as an official image due to the X socket mounting and graphics drivers mess.  The unlisted gzclient image runs in to issues for me, but not that interested in playing with it much further at the moment.

It looks like gzserver puts some persistent data in  /root/.gazebo, so having guidance on how to manage that would be helpful.  Do you think it makes sense to make that folder a volume as well?

psftw:
robotarmy

:+1: cool stuff

gazebo/content.md
ruffsl: still working on this bit.

psftw:
Lots of spelling/grammar issues in the latest changes.  In general, I still don't feel like I know what a "good" use case looks like.  A trivial example would be useful to illustrate the flow of running a job and then collecting the results.  Seeing ``--rm`` on the command line was surprising -- is it typical to only be interested in logging output?  I don't mind suggesting some edits once it is a little closer on these fronts.

BTW, the reason TravisCI isn't happy with this is due to the extra spaces at the end of content.md -- the markdownfmt.sh script can be used to check this with ``./markdownfmt.sh -d gazebo``.

ruffsl:
@psftw 
Yep, thats what I get when I break my atom spellcheck plugin again.
I've added a deployment example looking at logging, so check it out.
Also, `./markdownfmt.sh` no longer gives me trouble be TravisCI still is unhappy.

psftw:
@ruffsl nice work on the expanded docs.  LGTM apart from the broken Wikipedia link.  I've sent over some suggested wording tweaks to also consider.  I'm not sure how to appease the markdown gods either, but I added some extra newlines to the last section which might help.

psftw:
https://github.com/ros-infrastructure/docs/pull/2

gazebo/content.md
```diff
@@ -0,0 +1,112 @@
...
+> [wikipedia.org/wiki/Gazebo_simulator](https://en.wikipedia.org/wiki/Gazebo_simulator)
+
+[%%LOGO%%](http://www.gazebosim.org/)
+
+# How to use this image
+
+## Create a `Dockerfile` in your Gazebo project
+
+	FROM gazebo:gzserver5
+	# place here your application's setup specifics
```
yosifkit: Is there a specific folder that a user should `ADD` their files to so that gzserver will find it?

gazebo/content.md
```diff
@@ -0,0 +1,112 @@
...
+
+Conducting such complex simulations with high validity remains computationally demanding, and often times outside the capacity of a modest local workstation. With the added complexity of the algorithms being benchmarked, we can soon exceed the capacity of even the most formidable servers. This is why a more distributed approach remains attractive for those who begin to encounter limitations of a centralized computing host. However, the added complication of building and maintaining a distributed testbed over a set of clusters has for a while required more time and effort than many smaller labs and businesses would have deemed appropriate to implement.
```
yosifkit: `s/often times/oftentimes/` or `often` (either word is valid)

gazebo/content.md
```diff
@@ -0,0 +1,112 @@
...
+
+With the advancements and standardization of software containers, roboticists are primed to acquire a host of improved developer tooling for building and shipping software. To help alleviate the growing pains and technical challenges of adopting new practices, we have focused on providing an official resource for using Gazebo with these new technologies.
+
+## Deployment suggestions
+
+The `gzserver` tags are designed to have a small footprint and simple configuration, thus only include the required Gazebo dependencies are included. The standard messaging port `11345` is exposed to allow for client connections and messages API.
```
yosifkit: > thus only include the required Gazebo dependencies are included.

include less include :wink:

yosifkit:
Just a few small suggestions/nitpicks above.

The travis failure is saying that you have `README.md` changes. This is to help direct users when they don't modify the `content.md` and can be ignored here. In later PR's you could ignore the `README` file entirely and it would be auto generated by the `docker-library-bot`.

gazebo/content.md
ruffsl: Where users should mount their files for gazebo is described in detail under the volumes section. I wouldn't recommend users bake-in worlds models or robot. sdf files into image itself using the `ADD` argument in the Dockerfile. Those files are usually dynamic in nature as the user tweaks or iterates designs during runtime.

gazebo/content.md
ruffsl: :ok_hand:

gazebo/content.md
ruffsl: :ok_hand:

yosifkit:
LGTM pending https://github.com/docker-library/official-images/pull/855

tianon:
LGTM

PR #268, opened 2015-07-01
Update readme.rd "own conf file" section

gitshawn:
Update readme.rd "own conf file" without dockerfile section.

### DIFF
--- a/redis/README.md
+++ b/redis/README.md
@@ -48,7 +48,7 @@ You can create your own Dockerfile that adds a redis.conf from the context into 
 
 Alternatively, you can specify something along the same lines with `docker run` options.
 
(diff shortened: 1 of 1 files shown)


Moghedrin:
So, where you change the `-v` to a `-d` actually still needs to be `-v` for that volume.

You probably meant:
`docker run -d -v /myredis/conf/redis.conf:/usr/local/etc/redis/redis.conf --name myredis redis redis-server /usr/local/etc/redis/redis.conf`

Other than needing to fix that, this looks good.

tianon:
and moving the change over to the "content.md" file instead, since
"README.md" is generated. :+1:

yosifkit:
fixed by #271

PR #272, opened 2015-07-06
Update rabbitmq image documentation

tianon:
Fixes https://github.com/docker-library/rabbitmq/issues/6 (see https://github.com/docker-library/rabbitmq/issues/6#issuecomment-97721548; thanks @dumbbell! :heart:)

This also documents the change in https://github.com/docker-library/rabbitmq/pull/30 which makes cookies much easier to use.

### DIFF
--- a/rabbitmq/content.md
+++ b/rabbitmq/content.md
@@ -10,32 +10,32 @@ RabbitMQ is open source message broker software (sometimes called message-orient
 
 ## Running the daemon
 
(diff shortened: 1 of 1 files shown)


tianon:
Now that https://github.com/docker-library/official-images/pull/871 is actually merged, this is fair game. :+1:

yosifkit:
LGTM

PR #273, opened 2015-07-09
kibana documentation : environnement variable

tgermain:
The official doc does not mention that you can pass the address of you elastic search instance with the `ELASTICSEARCH_URL` environment variable.

See https://github.com/docker-library/kibana/blob/master/4.1/docker-entrypoint.sh

### DIFF
--- a/kibana/README.md
+++ b/kibana/README.md
@@ -29,6 +29,11 @@ This image includes `EXPOSE 5601` ([default `port`](https://www.elastic.co/guide
 
 	docker run --name some-kibana --link some-elasticsearch:elasticsearch -p 5601:5601 -d kibana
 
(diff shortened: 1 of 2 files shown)


yosifkit:
Looks good, can you put this change in content.md (readme.md is generated)?

tgermain:
Done, a bit dirtily : one commit reverse changes to readme.md and the other for changes on content.md. I do'n't have a proper git client on windows :(

tianon:
@tgermain that's alright -- I'll pull your commits locally and squash them for you at this point to merge (now that we're into the correct file) :+1:

Thanks!

tianon:
9893125c62d2b83fd102b6accbbad3e3c8362f7a :+1:

PR #86, opened 2014-10-26
Add "VSize" to "Supported tags and respective `Dockerfile` links"

tianon:
Opening this for discussion.  It takes quite a long time to fetch the "virtual size", since it has to walk the entire ancestry and the registry is _really_ slow at handing back the JSON objects for each layer.

Fixes https://github.com/docker-library/java/issues/15

Here's an example from `debian`:

- [`jessie` (*jessie/Dockerfile*)](https://github.com/tianon/docker-brew-debian/blob/b34a02cb006a698c1b4be59fb1d4ba8eabc502b6/jessie/Dockerfile) [VSize: 121.8 MB]
- [`oldstable` (*oldstable/Dockerfile*)](https://github.com/tianon/docker-brew-debian/blob/b34a02cb006a698c1b4be59fb1d4ba8eabc502b6/oldstable/Dockerfile) [VSize: 78.4 MB]
- [`sid` (*sid/Dockerfile*)](https://github.com/tianon/docker-brew-debian/blob/b34a02cb006a698c1b4be59fb1d4ba8eabc502b6/sid/Dockerfile) [VSize: 121.9 MB]
- [`6.0.10`, `6.0`, `6`, `squeeze` (*squeeze/Dockerfile*)](https://github.com/tianon/docker-brew-debian/blob/b34a02cb006a698c1b4be59fb1d4ba8eabc502b6/squeeze/Dockerfile) [VSize: 78.4 MB]
- [`stable` (*stable/Dockerfile*)](https://github.com/tianon/docker-brew-debian/blob/b34a02cb006a698c1b4be59fb1d4ba8eabc502b6/stable/Dockerfile) [VSize: 85.1 MB]
- [`testing` (*testing/Dockerfile*)](https://github.com/tianon/docker-brew-debian/blob/b34a02cb006a698c1b4be59fb1d4ba8eabc502b6/testing/Dockerfile) [VSize: 121.8 MB]
- [`unstable` (*unstable/Dockerfile*)](https://github.com/tianon/docker-brew-debian/blob/b34a02cb006a698c1b4be59fb1d4ba8eabc502b6/unstable/Dockerfile) [VSize: 121.9 MB]
- [`7.7`, `7`, `wheezy`, `latest` (*wheezy/Dockerfile*)](https://github.com/tianon/docker-brew-debian/blob/b34a02cb006a698c1b4be59fb1d4ba8eabc502b6/wheezy/Dockerfile) [VSize: 85.1 MB]
[...]

PR #274, opened 2015-07-11
Match feature in docker-mysql (allow creation of more than one database)

victorhahncastell:
See pull request #86 in docker-library/mysql
https://github.com/docker-library/mysql/pull/86


yosifkit:
Closing since the companion PR (docker-library/mysql#86) was also closed.

PR #276, opened 2015-07-13
Adds Apache TomEE folder for official image

lordofthejars:
Adds Apache TomEE folder for official image


lordofthejars:
This PR adds docs for official Apache TomEE image.

psftw:
You'll definitely want to indent any code such as ``docker run`` examples so that they are formatted correctly by Markdown.  It seems the version numbers in the docs are outdated, and in general I would avoid putting version numbers directly in the docs if you can avoid it to reduce the update overhead.  Please call out the relationship between Tomitribe and the TomEE project to be clear for new users.  It seems odd to use the Tomitribe logo vs. TomEE logo since this image is really about packaging the upstream project.  I am totally on board with calling out that Tomitribe built and supports this image, but ultimately it's the TomEE project that owns the upstream releases.  I know the lines can be blurry, but it's important in my mind to be really clear about it.  It seems this image is most useful as a base for war-based applications, so it may be worth calling that out directly as the recommended path.  It would also be useful (maybe it's obvious?) to call out the use cases where enabling the manager app is actually useful (i.e. remote app deployment?).  Sorry, I'm not a tomcat pro or I might have more concrete feedback on the docs.

lordofthejars:
Yes I know that the logo it is a bit strange thing, but TomEE has no logo yet. So we decided to use the Tomitribe one, but of course as soon as we have a logo we will change it. I have fixed the versions problem to be more generic document and also clarifying how to deploy a war file using this image. I have not added much information about manager since it is something that probably people should avoid for security reasons, but I mention in the document to maintain an equivalence to Tomcat document.

tomee/content.md
```diff
@@ -0,0 +1,58 @@
...
+-	Apache TomEE Plus delivers all that is in the Web Profile and JAX-RS (RESTfull Services), plus EJB Full, Java EE Connector Architecture, JMS (Java Message Service) and JAX-WS (Web Services).
+-	Apache TomEE Plume delivers all that is in the Plus Profile, but includes Mojarra and EclipseLink support.
+
+All Dockerfile files of supported versions can be found at https://github.com/tomitribe/docker-tomee
+
+# How to use this image.
+
+Run the default TomEE server (`CMD ["catalina.sh", "run"]`):
+
+docker run -it --rm tomee/<flavor>:<tomeeversion>\-<java-version>\-
```
psftw: All of the example commands have image names that don't match what was proposed in the PR to ``docker-library/official-images``.  I'm seeing tag names of the form: ``tomee:8-jre-1.7.2-plus``.

psftw:
Posted a comment on an outdated diff (oops).  The issue I noticed was that the tag names used in example commands don't match the proposed tag names.

lordofthejars:
is it ok? Where is the definition of tags?

tomee/content.md
```diff
@@ -0,0 +1,61 @@
...
+
+You can then go to `http://localhost:8888` or `http://host-ip:8888` in a browser.
+
+The configuration files are available in `/usr/local/tomee/conf/`. By default, no user is included in the "manager-gui" role required to operate the "/manager/html" web application. If you wish to use this app, you must define such a user in `tomcat-users.xml`.
+
+You can also use this image as base image for deploying your war application. To do it you need to create a Dockerfile based on Tomee Docker image and add the war file in `webapps` directory:
+
+	ADD <locationofapplication>/<warfile> /usr/local/tomee/webapps/<warfile>
+
+# Issues & Contributors
```
yosifkit: This section and below is auto-generated for `README.md`. The text would come from https://github.com/docker-library/docs/blob/master/.template-helpers/user-feedback.md or you can supply your own `user-feedback.md` if you want different text.

License section is generated from the `license.md` (https://github.com/docker-library/docs/blob/master/update.sh#L97-L100)

yosifkit:
Could you replace the indented code block with [fenced code blocks](https://github.com/docker-library/docs/pull/313/files#diff-9e92444ebf4f1038de4cc2c838492a43R13) with a language specifier? The new hub renders them nicer than indented blocks (https://github.com/docker/hub-feedback/issues/124#issuecomment-127434597).

Just this new comment and echo my previous [comment](https://github.com/docker-library/docs/pull/276#discussion_r37564896) about the "Issues and Contributors" and "License" sections.

Anything else @tianon?

tomee/content.md
```diff
@@ -0,0 +1,61 @@
+#What is TomEE?
```
tianon: Single space between `#` and the header text, please.

lordofthejars:
Everything is fixed in docs part. Also official image is merged (https://github.com/docker-library/official-images/pull/889#event-425624672) I think everything is ready :).

yosifkit:
LGTM

tianon:
LGTM

PR #277, opened 2015-07-14
Explain `PGDATA` env variable, which is useful when data volume is fs mount.

bakavic:
Provide info on getting past Postgres `initdb`'s grumpiness with fs mounts. Related to this [PR](https://github.com/docker-library/postgres/pull/72).


yosifkit:
LGTM

tianon:
LGTM

PR #279, opened 2015-07-15
Add code highlighting to volume mount example

sublimino:
Space indentation required to add code in a list


yosifkit:
Can you put this in `content.md` instead, readme is generated. Also the travis `markdownfmt` says that it should be two tabs instead of a tab and 4 spaces.

PR #281, opened 2015-07-16
Updates to websphere-liberty docs

davidcurrie:
Add details of shared class cache usage and note on use of boot2docker ip (as per comment submitted on Docker Hub).


yosifkit:
LGTM

davidcurrie:
Document move from IBM JRE to OpenJRE as per https://github.com/docker-library/official-images/pull/939 (at which point the previous update on shared class usage is no longer applicable).

yosifkit:
New parts LGTM too :smile:

tianon:
:+1:

PR #283, opened 2015-07-16
Update variant.sh with a bit more flexibility (so we can have "variant override" files per-image)

tianon:
See also https://github.com/docker-library/redis/pull/28, https://github.com/docker-library/official-images/pull/897, and http://redis.io/topics/memory-optimization#using-32-bit-instances.


yosifkit:
LGTM

yosifkit:
`mkdownfmt` says use ` *not*` instead of `_not_`.

tianon:
Fixed!

PR #284, opened 2015-07-17
Add documentation for how to use Celery and Redis together

tianon:
See https://github.com/docker-library/official-images/pull/907.


yosifkit:
LGTM

PR #291, opened 2015-07-23
Add documentation for new postgres initdb logic

tianon:
Documentation change for https://github.com/docker-library/postgres/pull/75.

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -45,7 +45,7 @@ This optional environment variable can be used to define another location - like
 
 # How to extend this image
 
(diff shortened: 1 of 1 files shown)


yosifkit:
LGTM

md5:
It might be good to point out that `psql` commands should generally be run as user `postgres` and that no password is necessary owing to the use of `trust` authentication for a local Unix socket connection.

tianon:
@md5 agreed; want to open an issue or PR? :innocent:

md5:
I could probably do that. However, I think that the issue of whether `*.sql` files should be run as `postgres` or `$POSTGRES_USER` should probably be resolved first, since both of those could probably be documented at the same time.

md5:
cf. https://github.com/docker-library/postgres/issues/81

tianon:
:+1:

md5:
See docker-library/postgres#82 for the entrypoint update and https://github.com/docker-library/docs/pull/319 for the docs update.

PR #292, opened 2015-07-24
Add redmine docs

yosifkit:


psftw:
Looks good.  I think it is worth mentioning the admin/admin default account directly as it is buried deep in the upstream wiki page that is referenced.

tianon:
@yosifkit want to add that note before, or after merge? :wink:

yosifkit:
No need to hurry on merge, since the library-bot cannot currently auto-update docs.  I'll get this PR updated.

yosifkit:
Updated.

tianon:
LGTM :+1:

tianon:
Currently stuck with https://github.com/docker/hub-beta-feedback/issues/238. :disappointed:

PR #295, opened 2015-07-29
Fix for deprecated mojo function

### LINKED ISSUE #294: method Pluck is no longer available in Mojo:5.80
Pluck is used in push.pl but it had been removed since Mojo:5.73: https://github.com/kraih/mojo/blob/master/Changes#L237

Can you guys update this?

yosifkit:
fixes #294


tianon:
LGTM

PR #300, opened 2015-08-05
HAProxy: Adding a stanza to test the config file

kyzh:
This testing instruction will hopefully help people in their workflow.
The goal is to test the config before building a new image.
I also removed the semi column and added a h2 to 'Build and run'

Signed-off-by: Florentin Raud <flo@ngineered.co.uk>

### DIFF
--- a/haproxy/content.md
+++ b/haproxy/content.md
@@ -21,12 +21,17 @@ Note: Many configuration examples propose to put `daemon` into the `global` sect
 	FROM haproxy:1.5
 	COPY haproxy.cfg /usr/local/etc/haproxy/haproxy.cfg
 
-## Test the configuration file	
-	docker run -it --rm  --name my-running-haproxy -v /path/to/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro haproxy:1.5 haproxy -c -f /usr/local/etc/haproxy/haproxy.cfg
-	
-## Build and run
+
+## Build the container
 
 	docker build -t my-haproxy .
+
+## Test the configuration file
+
+	docker run -it --rm  --name haproxy-syntax-check haproxy:1.5 haproxy -c -f /usr/local/etc/haproxy/haproxy.cfg
+	
+## Run the container
+
 	docker run -d --name my-running-haproxy my-haproxy
 
 ## Directly via bind mount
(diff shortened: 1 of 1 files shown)


haproxy/content.md
```diff
@@ -21,7 +21,10 @@ Note: Many configuration examples propose to put `daemon` into the `global` sect
 	FROM haproxy:1.5
 	COPY haproxy.cfg /usr/local/etc/haproxy/haproxy.cfg
 
-Build and run:
+## Test the configuration file	
+	docker run -it --rm  --name my-running-haproxy -v /path/to/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro haproxy:1.5 haproxy -c -f /usr/local/etc/haproxy/haproxy.cfg
```
tianon: Since this section is about building an image, I think it would be safer and more generally likely to work on an arbitrary user's box if we separated this out to use the built image instead of a bind-mount, ie:
```markdown
    FROM haproxy:1.5
    COPY haproxy.cfg /usr/local/etc/haproxy/haproxy.cfg

## Build

    docker build -t my-haproxy .

## Test the configuration file

    docker run --rm my-haproxy haproxy -c -f /usr/local/etc/haproxy/haproxy.cfg

## Run

    docker run -d --name my-running-haproxy my-haproxy
```

Would you like to update your PR, or would you rather I carry the PR from here?  Thanks for pointing out the enhancement regardless! :+1:

kyzh: Thanks for taking time to read the PR and the kind word
If that is ok with you I'll work the PR.

tianon: Yeah, sounds good!  Feel free to force push to the same branch you've
already got here (which will update the PR if you aren't familiar with the
GitHub flow), and give us a ping here when you're ready (since GitHub
doesn't notify when the PR is updated). :heart:

tianon: Any luck with this one, @kyzh?

kyzh:
I done the change as per @tianon comment.
If that's still not quite right, feel to go from there.
Thanks for your time, here is a :trophy:  for you :)

tianon:
LGTM :+1:

tianon:
Merged in 1741d2845d9de47fd6b3e503c24ec0f565595b48.

PR #301, opened 2015-08-05
Remove polls

brianteeman:


mbabker:
@brianteeman You'll need to update the `content.md` file; bots build the README.md file separately.  Also, this should be updated on Joomla's Wikipedia page as that's where I copied the introductory paragraph from.

brianteeman:
I will check the content.md file. I am unable to edit Wikipedia as they
will not let me as I am classed as a non neutral editor

PR #303, opened 2015-08-06
document linking a new Cassandra node to the first one

clakech:
Hi,

I am using this image to setup a spark+cassandra cluster and I was wondering if we could use the docker run --link option to link new node to the first one instead of docker inspect command value pass to an environnement variable.

With this change, adding a node to the cluster is simpler, isn't it ?
```
docker run --name some-cassandra2 -d --link some-cassandra:cassandra cassandra:tag
```

WDYT? sorry if I am wrong, I am neither a docker expert, nor a C* one ;-)

cf: https://github.com/docker-library/cassandra/pull/21

### DIFF
--- a/cassandra/README.md
+++ b/cassandra/README.md
@@ -38,6 +38,10 @@ Using the environment variables documented below, there are two cluster scenario
 
 ... where `some-cassandra` is the name of your original Cassandra Server container, taking advantage of `docker inspect` to get the IP address of the other container.
 
+Or you may use the docker run --link option to tell the new node where the first is:
+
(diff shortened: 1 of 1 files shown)


clakech:
The PR https://github.com/docker-library/cassandra/pull/21 has been merged to allow use of --link using cassandra containers

To reflect this change into the doc, merge this PR ;-)

Thanks

yosifkit:
This needs to be put in `content.md` since readme is generated. :smile:

clakech:
oups... :sweat_smile:

clakech:
done here: https://github.com/docker-library/docs/pull/334

PR #313, opened 2015-08-13
Use explicit-type code blocks in a lot of obvious places…

tianon:
… (now that the Hub actually renders these _better_ than the no-type versions)

See also https://github.com/docker/hub-beta-feedback/issues/124 and https://github.com/docker/hub-beta-feedback/issues/179.


md5:
The `jetty` changes LGTM

yosifkit:
LGTM

PR #319, opened 2015-08-22
Expand documentation of postgres initdb logic

md5:
* Explain the execution order of initdb extension files
* Document that *.sql scripts are run as $POSTGRES_USER
* Recommend running psql commands inside *.sh scripts as $POSTGRES_USER

This assumes that https://github.com/docker-library/postgres/pull/82 will be merged.


tianon:
LGTM :+1: :heart:

Moghedrin:
LGTM

md5:
Good thing docker-library/postgres#84 was merged before these docs were updated since the suggestion to use `$POSTGRES_USER` in the `*.sh` scripts wouldn't have worked before! :+1:

PR #324, opened 2015-08-25
Fixed typo in db-filter config

JacerOmri:
as seen in https://www.odoo.com/documentation/8.0/setup/deploy.html


yosifkit:
Can you move this to `content.md`, readme is generated? Otherwise seems fine.

/cc @aab-odoo is this change fine?

aab-odoo:
This is correct. Thanks

PR #326, opened 2015-08-26
Postgres link in Readme when starting celery worker and beat

Scriptkiddi:
Added a database link option to the celery run commad because it easily missed when first setting up the sentry container, that you need to link to your postgres container


yosifkit:
What is sentry doing to start the celery worker/beat that requires the database?  If there is something special then this should be the only way to run celery workers.

Also, changes should be in `content.md` since readme is generated.

PR #330, opened 2015-09-02
Clarify slightly that the image is actually updated by making pull requests against the official-images repository

tianon:


yosifkit:
LGTM

PR #334, opened 2015-09-04
document linking a new Cassandra node to the first one

clakech:
Hi,

I am using this image to setup a spark+cassandra cluster and I was wondering if we could use the docker run --link option to link new node to the first one instead of docker inspect command value pass to an environnement variable.

With this change, adding a node to the cluster is simpler, isn't it ?

docker run --name some-cassandra2 -d --link some-cassandra:cassandra cassandra:tag
WDYT? sorry if I am wrong, I am neither a docker expert, nor a C* one ;-)

cf: https://github.com/docker-library/cassandra/pull/21 and https://github.com/docker-library/docs/pull/303


clakech:
hi, you someone have a look into this one please ?

yosifkit:
It seems I totally missed this one :open_mouth:.  LGTM

clakech:
thx

PR #338, opened 2015-09-11
Don't enforce host for odoo

gagarine:
So it works directly on OSX with the Linux VM.


yosifkit:
Change seems reasonable to me; can you move it to `content.md`?

PR #339, opened 2015-09-11
Add HHVM image

baptistedonaux:


baptistedonaux:
What's the check which fails ? I have follow https://github.com/docker-library/docs/#image-namereadmemd and see others images structure but I don't see the anomaly.

hhvm/README-short.txt
```diff
@@ -0,0 +1 @@
+HHVM is an open PHP engine build by Facebook.
```
andrerom: build => built 

but maybe just use their own wording:
"HHVM is an open-source virtual machine designed for executing programs written in Hack and PHP" + ", made by Facebook."

baptistedonaux: Yes thank you for the re-read.

hhvm/update.sh
```diff
@@ -0,0 +1,27 @@
...
+
+echo
+echo '%%LOGO%%'
+
+echo
+cat <<'EOF'
+# How is this image created?
+
+This image is a prime example of using the [`debian`](https://registry.hub.docker.com/_/debian/) image effectively.
+EOF
```
yosifkit: This file isn't needed.  It is used to generate `readme.md` when there is no `content.md`, which doesn't apply here.

hhvm/content.md
```diff
@@ -0,0 +1,26 @@
...
+HHVM is an open-source virtual machine designed for executing programs written in Hack and PHP. HHVM uses a just-in-time (JIT) compilation approach to achieve superior performance while maintaining the development flexibility that PHP provides.
+
+> [wikipedia.org/wiki/HipHop_Virtual_Machine](https://en.wikipedia.org/wiki/HipHop_Virtual_Machine)
+
+![logo](https://raw.githubusercontent.com/baptistedonaux/docker-images/hhvm/logo.png)
+
+# How to use this image
+
+```console
+$ docker run --name hhvm -v /some/content:/usr/share/nginx/html:ro -P -d hhvm
```
yosifkit: Since the default of the image is currently to run it with `fpm`, I think there should be some documentation on how to run something like the nginx container in front of it.  Some users will try to just point their browser at the fpm container and this would help alleviate those issues.

It should also mean that the `-P` could be dropped since the users will be hitting nginx rather than the hhvm container.

hhvm/update.sh
baptistedonaux: Okay, I remove it.

hhvm/content.md
baptistedonaux: I will add more documentation and examples and I come back to you.

baptistedonaux: I added more documentation. I hope that it's more explicit for user.

hhvm/content.md
```diff
@@ -0,0 +1,84 @@
...
+![logo](https://raw.githubusercontent.com/baptistedonaux/docker-hhvm/master/logo.png)
+
+# How to use this image
+
+Two methods (CLI vs FastCGI):
+
+## CLI mode
+
+```console
+$ docker run -v /foo.php:/foo.php foo.php
```
yosifkit: This line is missing the `hhvm` image to run :wink: `docker run -v /foo.php:/foo.php hhvm foo.php`

hhvm/content.md
```diff
@@ -0,0 +1,84 @@
...
+$ docker run -v /foo.php:/foo.php hhvm foo.php
+$ docker run -v /foo.php:/foo.php php foo.php
```
yosifkit: This one too :wink:.

hhvm/content.md
```diff
@@ -0,0 +1,84 @@
...
+
+# How to use this image
+
+Two methods (CLI vs FastCGI):
+
+## CLI mode
+
+```console
+$ docker run -v /foo.php:/foo.php foo.php
+$ docker run -v /foo.php:/foo.php hhvm foo.php
```
yosifkit: Maybe this one is to show that you can still specify hhvm as the command? `docker run -v /foo.php:/foo.php hhvm hhvm foo.php`

baptistedonaux:
I changed Nginx configuration because I introduced a mistake.

baptistedonaux:
I don't understand what the fail in Travis. An idea ?

md5:
It looks like it doesn't like the fenced code block without a syntax type. You can try `Nginx` for the syntax.

The other failure is that it wants a newline before the fenced `yaml` code block for the Docker Compose example.

baptistedonaux:
Thanks @md5 ! Travis is strict…

Documentation and image ready.

hhvm/README-short.txt
md5: Looks like this still needs to be fixed.

update.sh
```diff
@@ -36,6 +36,7 @@ declare -A otherRepos=(
 	[gazebo]='https://github.com/osrf/docker_images'
 	[glassfish]='https://github.com/aws/aws-eb-glassfish'
 	[haskell]='https://github.com/freebroccolo/docker-haskell'
+	[hhvm]='https://github.com/baptistedonaux/docker-images/tree/master/hhvm'
```
md5: This needs to be updated to https://github.com/baptistedonaux/docker-hhvm

baptistedonaux: Fixed !

hhvm/README-short.txt
```diff
@@ -0,0 +1 @@
+HHVM is an open PHP engine built by Facebook.
```
baptistedonaux: @md5 build => built

hhvm/content.md
```diff
@@ -0,0 +1,85 @@
+# What is HHVM?
+
+HHVM is an open-source virtual machine designed for executing programs written in Hack and PHP. HHVM uses a just-in-time (JIT) compilation approach to achieve superior performance while maintaining the development flexibility that PHP provides.
+
+> [wikipedia.org/wiki/HipHop_Virtual_Machine](https://en.wikipedia.org/wiki/HipHop_Virtual_Machine)
+
+![logo](https://raw.githubusercontent.com/baptistedonaux/docker-hhvm/master/logo.png)
```
md5: I think this should use `%%LOGO%%`

hhvm/content.md
```diff
@@ -0,0 +1,73 @@
+# What is HHVM?
+
+HHVM is an open-source virtual machine designed for executing programs written in Hack and PHP. HHVM uses a just-in-time (JIT) compilation approach to achieve superior performance while maintaining the development flexibility that PHP provides.
+
+> [wikipedia.org/wiki/HipHop_Virtual_Machine](https://en.wikipedia.org/wiki/HipHop_Virtual_Machine)
+
+%%LOGO%%
```
baptistedonaux: @md5 %%LOGO%% changes for hhvm/content.md and hhvm/README.md

md5: The change here looks good. I think you may want to regenerate the `README.md` though. I'm not sure whether `README.md` should actually be left out of initial doc submission and allowed to be generated after the PR merges, though.

baptistedonaux: Regenerated !

PR #342, opened 2015-09-14
Some commands (e.g pwd) need bash

sandoracs:
Hello!

If a Makefile contains shell command, we should use a shell to launch make process.


yosifkit:
Oh, I see. Makefiles are supposed to run under a shell: https://www.gnu.org/software/make/manual/make.html#Utilities-in-Makefiles-1.

LGTM

tianon:
Man, that's a long tail; sorry for the delay!

LGTM

PR #348, opened 2015-09-24
docs for trying to be official docker images for Erlang OTP

c0b:
link with docker-library/official-images#1075


erlang/content.md
```diff
@@ -0,0 +1,28 @@
...
+> [wikipedia.org/wiki/Erlang_(programming_language)](https://en.wikipedia.org/wiki/Erlang_%28programming_language%29)
+
+%%LOGO%%
+
+# How to use this image
+
+## Create a `Dockerfile` in your Erlang app project
+
+```dockerfile
+FROM erlang:18-onbuild
```
yosifkit: You don't have an onbuild tag; probably just add the commands that would be in the onbuild. Something like this?

(warning, I am not an erlang user)
```dockerfile
FROM erlang:18
WORKDIR /usr/src/myapp/
COPY ./ /usr/src/myapp/
CMD ["escript", "myapp.erl"]
```

c0b: I'll remove these onbuild lines for now, will add the package manager `rebar` for the ONBUILD lines, it's like the pip for python
https://github.com/docker-library/python/blob/master/3.5/onbuild/Dockerfile

tianon:
LGTM

yosifkit:
LGTM

PR #351, opened 2015-09-30
Fix typo in README.md

BastianHofmann:


yosifkit:
Thanks for the fix! Could you put it in `content.md`?  `Readme.md` is generated so it would just get overwritten.

BastianHofmann:
Done.

yosifkit:
LGTM

PR #353, opened 2015-10-01
Jetty: Update docs for HTTP/2 support

md5:
Jetty has had HTTP/2 support since Jetty 9.3.2 or earlier, but the changes in Jetty 9.3.3 (cf. docker-library/official-images#1088) make it easy to configure and use in the `jetty` image as an HTTP/2 server.

### DIFF
--- a/jetty/content.md
+++ b/jetty/content.md
@@ -17,7 +17,7 @@ $ docker run -d %%REPO%%
 You can test it by visiting `http://container-ip:8080` or `https://container-ip:8443/` in a browser. To expose your Jetty server to outside requests, use a port mapping as follows:
 
 ```console
-$ docker run -d -p 8888:8080 %%REPO%%
+$ docker run -d -p 80:8080 -p 443:8443 %%REPO%%
 ```
 
 This will map port 8080 inside the container as port 80 on the host and container port 8443 as host port 443. You can then go to `http://host-ip` or `https://host-ip` in a browser.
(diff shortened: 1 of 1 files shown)


md5:
cf. https://github.com/eclipse/jetty.project/commit/d02a5c3c8bbf3fe05b262125358677847b303688 in particular

md5:
I really don't understand what Travis is complaining about here. It seems to be saying that `README.md` was updated, but I don't see that in these commits.

md5:
Looks like I made the build happy with some rebasing. :+1:

yosifkit:
LGTM

jetty/content.md
```diff
@@ -8,19 +8,21 @@ Jetty is a pure Java-based HTTP (Web) server and Java Servlet container. While W
...
-You can test it by visiting `http://container-ip:8080` in a browser or, if you need access outside the host, on port 8888:
+You can test it by visiting `http://container-ip:8080` or `https://container-ip:8443/` in a browser. To expose your Jetty server to outside requests, use a port mapping as follows:
 
 ```console
-$ docker run -d -p 8888:8080 %%REPO%%:9
+$ docker run -d -p 8888:8080 %%REPO%%
 ```
 
-You can then go to `http://localhost:8888` or `http://host-ip:8888` in a browser.
+This will map port 8080 inside the container as port 80 on the host and container port 8443 as host port 443. You can then go to `http://host-ip` or `https://host-ip` in a browser.
```
tianon: Uh, the `run` line above is still using `8888` on the host. :smile:

md5: Fixed in https://github.com/appropriate/docker-library-docs/commit/709917df26cb21c7307a5abfd449e7839aaa8b1b

I think this might have been something that got messed up while resolving a merge conflict.

tianon:
LGTM :+1:

PR #354, opened 2015-10-02
Docs for sourcemage official image

vaygr:
Request for docs for sourcemage official image [pull request](https://github.com/docker-library/official-images/pull/1090)


vaygr:
Official image is already there, waiting for this PR for a nice page at the Docker Hub.

tianon:
LGTM

tianon:
Whoa one sec, just noticed that `--privileged`.  That's going to need some explanation. :wink:

tianon:
There's got to be a way to make the package manager work with less than full `--privileged`. :confused:

tianon:
```
CastFS failed the sanity check, staging disabled!
```

Looks like it does work fine without `--privileged`, it just doesn't get the full benefit of "CastFS".

tianon:
(Which I think is more OK in the case of Docker than in the normal desktop case since your installation ought to be happening in throw-away containers anyways, right?)

tianon:
I successfully got CastFS to work with just `--device /dev/fuse --cap-add SYS_ADMIN`, which is better than `--privileged` (which gives full privileges to the entire host device hierarchy, for example), but still isn't great since `SYS_ADMIN` controls a lot more than just whether we can `mount`.  Since it works without, I think I'd rather we not include it by default, and have CastFS notes in a separate section noting that there are security implications to granting these capabilities/privileges to the container.

tianon:
This is also really important to note because in a `Dockerfile`, we have no way of adding devices or extra capabilities, so it's great that the following `Dockerfile` actually works: (just tested it! :+1:)
```Dockerfile
FROM sourcemage
RUN cast tmux
```

vaygr:
Got it.

Please see my latest commit which takes care of that.

yosifkit:
LGTM

tianon:
Awesome, thanks! :+1: 

re-LGTM

tianon:
:balloon:

PR #357, opened 2015-10-03
celery workers also need the database connection

feldi88:
If the database is not linked, the workers do not find the tables.


yosifkit:
LGTM

PR #359, opened 2015-10-05
Move the theoretical best-effort support line from Docker 1.0 up to Docker 1.6

tianon:
Also, add a link to the Docker docs for installation.


psftw:
sounds reasonable to me based on a quick look at the data.  Pre-1.5 is a blip, and even 1.5 clients are not hitting hub that much these days.

tianon:
So maybe we could kick up to 1.6?

psftw:
Sure.  Labels anyone?  *ducks and hides*

tianon:
Done!

yosifkit:
LGTM

PR #360, opened 2015-10-05
Should state what that image actually does before explaining what this is for.

ekevoo:
I am new to Docker, but not to Python.

I was reading the description and I was completely confused as to how it could help in any way to get a new project off the ground. This is probably not the best phrasing, but at least should be helpful to people with that background, looking for how Docker could be helpful.

Thanks.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -81,7 +81,7 @@ This is the defacto image. If you are unsure about what your needs are, you prob
 
 ## `python:onbuild`
 
-This image makes building derivative images easier. For most use cases, creating a `Dockerfile` in the base of your project directory with the line `FROM python:onbuild` will be enough to create a stand-alone image for your project.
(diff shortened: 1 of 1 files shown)


yosifkit:
Seems good to me, but readme.md is [generated](https://github.com/docker-library/docs#image-namereadmemd), so can you put it in content.md?

ekevoo:
I noticed from the integration error message… But content.md seems to not have any mention of this at all. I'm confused! :(

yosifkit:
Oh, silly me. This part is [generated](https://github.com/docker-library/docs/blob/master/.template-helpers/variant-onbuild.md) generically for all images that have an onbuild variant.  We should be able to customize it by copying that `variant-onbuild.md` into the python directory and adding in the changes.

ekevoo:
Okay, I'll generate a new pull request some other time. Thanks for the pointers.

PR #361, opened 2015-10-08
Explain the python:onbuild variant.

ekevoo:
Give more details on how the onbuild variant of Python is different for people who are more familiar with Python than with Docker. More details in PR #360.


yosifkit:
LGTM

PR #362, opened 2015-10-08
Rails, no need to bundle install in the throw-away container

yosifkit:
Fixes https://github.com/docker-library/official-images/issues/1104


tianon:
LGTM

PR #364, opened 2015-10-14
Add README config changes to enable SSL

stav09:
README.md missing additional httpd.conf changes to enable modules required to support SSL

### DIFF
--- a/httpd/README.md
+++ b/httpd/README.md
@@ -52,7 +52,12 @@ COPY ./my-httpd.conf /usr/local/apache2/conf/httpd.conf
 
 #### SSL/HTTPS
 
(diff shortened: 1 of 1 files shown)


stav09:
I'm not sure how a valid README file update has broken the CI build...?

yosifkit:
The CI failing is because it is complaining that you changed `readme.md` because that file is generated.  Could you put the change into `content.md` instead?

Also, all the markdown is run through tianon's fork of `markdownfmt`,  so you'll need a newline after the paragraph and before the list.  Just run `./markdownfmt.sh -d httpd/content.md` and it will give a diff if there are changes needed.

Thanks for the contribution!

stav09:
Ah! No problems, will make the change. Thanks

tianon:
Doing an "old/stale/stalled PRs" pass and this one's in the list, sorry!  I'm going to close, since this looks like it was forgotten. :heart: :heart:

PR #367, opened 2015-10-16
latest gcc source updated and  COPY command is replaced by VOLUME to …

trailblazerakash:
…avoid extra memory wastage

### DIFF
--- a/gcc/README.md
+++ b/gcc/README.md
@@ -24,8 +24,8 @@ The GNU Compiler Collection (GCC) is a compiler system produced by the GNU Proje
 The most straightforward way to use this image is to use a gcc container as both the build and runtime environment. In your `Dockerfile`, writing something along the lines of the following will compile and run your project:
 
 ```dockerfile
-FROM gcc:4.9
-COPY . /usr/src/myapp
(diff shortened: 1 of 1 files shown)


gcc/README.md
```diff
@@ -24,8 +24,8 @@ The GNU Compiler Collection (GCC) is a compiler system produced by the GNU Proje
 The most straightforward way to use this image is to use a gcc container as both the build and runtime environment. In your `Dockerfile`, writing something along the lines of the following will compile and run your project:
 
 ```dockerfile
-FROM gcc:4.9
-COPY . /usr/src/myapp
+FROM gcc
+VOLUME  /usr/src/myapp
```
yosifkit: This would actually make [line 30](https://github.com/docker-library/docs/pull/367/files#diff-23eb9253b3d9879959b5bd03d51c15a8R30) fail since there would be no code to compile in the image.  One could just bind mount the code and run the compile in the container, like [line 46](https://github.com/docker-library/docs/pull/367/files#diff-23eb9253b3d9879959b5bd03d51c15a8R46).

PR #375, opened 2015-10-30
MariaDB: Added section about how to create database dumps.

ypid:
Related to: https://github.com/docker-library/postgres/issues/83


yosifkit:
I like it! :heart: LTGM

tianon:
Doesn't this do strange things with `-t` in there while redirecting?

PR #376, opened 2015-11-01
Link postgres as needed in sentry docs.

jmcarp:
Celery workers need to be linked to postgres; celery beat doesn't.

cc @mattrobenolt

### DIFF
--- a/sentry/content.md
+++ b/sentry/content.md
@@ -49,15 +49,15 @@ Sentry is a realtime event logging and aggregation platform. It specializes in m
 	-	using the celery image:
 
 		```console
-		$ docker run -d --name celery-beat --link some-postgres:postgres --link some-redis:redis  -e CELERY_BROKER_URL=redis://redis celery celery beat
-		$ docker run -d --name celery-worker1  --link some-postgres:postgres --link some-redis:redis  -e CELERY_BROKER_URL=redis://redis celery
+		$ docker run -d --name celery-beat --link some-postgres:postgres --link some-redis:redis -e CELERY_BROKER_URL=redis://redis celery celery beat
+		$ docker run -d --name celery-worker1 --link some-postgres:postgres --link some-redis:redis -e CELERY_BROKER_URL=redis://redis celery
 		```
 
 	-	using the celery bundled with sentry
 
 		```console
-		$ docker run -d --name sentry-celery-beat --link some-redis:redis sentry sentry celery beat
-		$ docker run -d --name sentry-celery1 --link some-redis:redis sentry sentry celery worker
+		$ docker run -d --name sentry-celery-beat --link some-postgres:postgres --link some-redis:redis sentry sentry celery beat
+		$ docker run -d --name sentry-celery1 --link some-postgres:postgres --link some-redis:redis sentry sentry celery worker
 		```
 
 ### port mapping


mattrobenolt:
I think we should link them both. I'm not entirely sure if we can confidently say that celery beat will never need access to Postgres.

jmcarp:
Unless sentry starts using postgres as a message broker, I'm not sure why beat would ever need postgres, but you'd know better than I would, and there's no harm in linking. Patch updated!

mattrobenolt:
That's _likely_ true, but I'm not 100% confident what all celery beat does since I don't build celery. :) So just playing it safe.

Thanks!

mattrobenolt:
Oh wait, I don't even have the ability to merge. :stuck_out_tongue_closed_eyes:

yosifkit:
Thanks for the fixes. (esp fixing the extra whitespace :heart::purple_heart::blue_heart:)

@mattrobenolt, does it work correctly with just the celery image or only the bundled one in the sentry image?  I'm not 100% confident that I ever saw it working with just the plain celery image.

mattrobenolt:
@yosifkit It'd require the one in the Sentry image. The plain celery image is more of an 'onbuild' thing, since Celery by itself doesn't do anything. That's analogous to say, the Django image. Celery is just a library and requires application code around it to function. :)

In our case, celery is just a dependency of Sentry and gets installed in our image.

sentry/content.md
```diff
@@ -49,15 +49,15 @@ Sentry is a realtime event logging and aggregation platform. It specializes in m
 	-	using the celery image:
 
 		```console
-		$ docker run -d --name celery-beat --link some-postgres:postgres --link some-redis:redis  -e CELERY_BROKER_URL=redis://redis celery celery beat
-		$ docker run -d --name celery-worker1  --link some-postgres:postgres --link some-redis:redis  -e CELERY_BROKER_URL=redis://redis celery
+		$ docker run -d --name celery-beat --link some-postgres:postgres --link some-redis:redis -e CELERY_BROKER_URL=redis://redis celery celery beat
```
mattrobenolt: Yeah, then this entire section should be removed since this 100% will not work with the vanilla celery image.

I didn't even realize this was here. :(

mattrobenolt:
@yosifkit How do you recommend we manage docs for `sentry` this way? Since I don't have commit privileges into `docker-library/docs`, it's a bit cumbersome to make trivial changes to make sure things keep up to date. And I don't _want_ to be pesky and overload with pull requests.

Or are you guys pretty diligent on keeping up with pull requests? (not implying I thought otherwise of course. :smile:)

I'm asking because we're working on cutting an 8.0 with our official docker support, so all of these docs are going to need to be revised and updated to reflect this.

yosifkit:
Feel free to PR away, and then poke us if we take too long.  Usually docs for current images are simple enough for us to merge quickly.  As long as you don't edit `readme.md` and run `./mardownfmt.sh -d sentry` so that there are no Travis issues, then we should be golden.

If you have any questions on anything here in the docs or about the official-images repo, feel free to ask.

Should we merge this, or wait and cut out the plain celery part?

mattrobenolt:
You can merge this now. I'll submit a PR to yank the vanilla celery references next.

mattrobenolt:
thx @yosifkit 

See #380

PR #380, opened 2015-11-04
sentry: remove reference to vanilla celery image

mattrobenolt:
This doesn't make sense and will never work.

See discussion on https://github.com/docker-library/docs/pull/376

/cc @yosifkit


yosifkit:
LGTM

PR #381, opened 2015-11-05
Update README.md

NoumanSaleem:
container fails to start when using documented `--addr` flag. Should be `--advertise`

### DIFF
--- a/swarm/README.md
+++ b/swarm/README.md
@@ -26,7 +26,7 @@ $ docker run --rm swarm create
 # on each of your nodes, start the swarm agent
 #  <node_ip> doesn't have to be public (eg. 192.168.0.X),
 #  as long as the swarm manager can access it.
-$ docker run -d swarm join --addr=<node_ip:2375> token://<cluster_id>
+$ docker run -d swarm join --advertise=<node_ip:2375> token://<cluster_id>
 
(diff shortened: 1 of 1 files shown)


yosifkit:
It is still part of the `--help` output and I am unable to reproduce failure with either flag.  None of the swarm guys are watching this repo, but you could try `#docker-distribution` channel on freenode IRC if you need help troubleshooting.
```console
$ docker run -it --rm swarm join --help
Usage: swarm join [OPTIONS] <discovery>

join a docker cluster

Arguments: 
   <discovery>    discovery service to use [$SWARM_DISCOVERY]
                   * token://<token>
                   * consul://<ip>/<path>
                   * etcd://<ip1>,<ip2>/<path>
                   * file://path/to/file
                   * zk://<ip1>,<ip2>/<path>
                   * [nodes://]<ip1>,<ip2>

Options:
   --advertise, --addr 							Address of the Docker Engine joining the cluster. Swarm manager(s) MUST be able to reach the Docker Engine at this address. [$SWARM_ADVERTISE]
   --heartbeat "20s"							period between each heartbeat
   --ttl "60s"								sets the expiration of an ephemeral node
   --discovery-opt [--discovery-opt option --discovery-opt option]	discovery options
```

PR #393, opened 2015-11-16
Bumping version of kibana to 4.2.1

tebriel:
Also updated dockerfile ref


tebriel:
Oh, didn't realize it was auto-generated.

PR #394, opened 2015-11-16
[ROS] Update for new Network functions in Docker v1.9

ruffsl:
Also adding an equivalent docker-compose example.

### DIFF
--- a/ros/content.md
+++ b/ros/content.md
@@ -61,7 +61,6 @@ The ROS runtime "graph" is a peer-to-peer network of processes (potentially dist
 
 ## Deployment example
 
-**NOTE:** This requires the experimental version of Docker for future networking features.  
 If we want our all ROS nodes to easily talk to each other, we'll can use a virtual network to connect the separate containers. In this short example, we'll create a virtual network, spin up a new container running `roscore` advertised as the `master` service on the new network, then spawn a message publisher and subscriber process as services on the same network.
 
 ### Build image
@@ -96,8 +95,8 @@ $ docker build --tag ros:ros-tutorials .
 > To create a container for the ROS master and advertise it's service:
 
 ```console
-$ docker run -it --rm\
-    --publish-service=master.foo \
+$ docker run -it --rm \
+    --net foo \
     --name master \
     ros:ros-tutorials \
     roscore
@@ -106,11 +105,11 @@ $ docker run -it --rm\
 > Now you can see that master is running and is ready manage our other ROS nodes. To add our `talker` node, we'll need to point the relevant environment variable to the master service:
 
 ```console
-$ docker run -it --rm\
-    --publish-service=talker.foo \
+$ docker run -it --rm \
+    --net foo \
+    --name talker \
     --env ROS_HOSTNAME=talker \
     --env ROS_MASTER_URI=http://master:11311 \
-    --name talker \
     ros:ros-tutorials \
     rosrun roscpp_tutorials talker
 ```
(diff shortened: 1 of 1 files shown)


yosifkit:
`markdownfmt` is sad :cry:.  Looks like you just need some newlines :smile:.

ruffsl:
Ok, think I've got it.

ruffsl:
Actually, @toddsampson, anything you think I should add?

ruffsl:
@yosifkit and @tianon, I wanted to ask if there was a better method of setting the `ROS_HOSTNAME` environment variable to that of the host name of the container itself? Such that I wouldn't have to explicitly specify the name of the container. I'm thinking this could be done by extending the entry point script to using the injected `HOSTNAME` env at runtime, but was wondering if there was a more docker-like method. On second thought, perhaps that'd be a useful addition, setting the `ROS_HOSTNAME` and `ROS_MASTER_URI` to use the `HOSTNAME` by default, rather than just `localhost`? Will the hostname change if the container changes networks?

tianon:
Can containers dynamically move between networks?

The hostname should be provided by the `--hostname` flag on run, and _should_ be unaffected by networks generally.  I think it'd totally make sense to have something simple like this in your entrypoint to default those variables sanely:
```bash
: ${ROS_HOSTNAME:=$HOSTNAME}
: ${ROS_MASTER_URI:=$ROS_HOSTNAME}
```
(adjusted of course if those values or that precedence ordering doesn't really make sense)

ruffsl:
Just gave that a shot in the `ros_entrypoint.sh`

```bash
#!/bin/bash
set -e

# setup ros environment
source "/opt/ros/$ROS_DISTRO/setup.bash"

# set ros hostname to container's hostname
: ${ROS_HOSTNAME:=$HOSTNAME}
: ${ROS_MASTER_URI:=http://$ROS_HOSTNAME:11311}

exec "$@"
```
But I'm realizing that in `/etc/hosts`, only the local container's `<container_id>` is listed, all other containers are only listed under the `<compose_folder>_<service_name>` or `<container_name>`, but never `<container_id>` . I suppose there is no way to enable that?

```
$ docker exec -it talker bash      
root@ab1a3ab71095:/# source ros_entrypoint.sh 
root@ab1a3ab71095:/# rostopic list
/chatter
/rosout
/rosout_agg
root@ab1a3ab71095:/# cat /etc/hosts 
172.19.0.4	ab1a3ab71095
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
172.19.0.2	listener
172.19.0.2	listener.rostutorials
172.19.0.3	master
172.19.0.3	master.rostutorials

$ docker exec -it listener bash
root@8c32438b2e60:/# source ros_entrypoint.sh
root@8c32438b2e60:/# rostopic list
/chatter
/rosout
/rosout_agg
root@8c32438b2e60:/# cat /etc/hosts
172.19.0.2	8c32438b2e60
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
172.19.0.3	master
172.19.0.3	master.rostutorials
172.19.0.4	talker
172.19.0.4	talker.rostutorials

```
So in light of this, setting the ros vars with to the hostname (or `<container_id>`) wont be of much use. As the master couldn't know the address that resolves to the hostname the listener subscribed with. We could have the script use `<container_name>` to set its own `ROS_HOSTNAME`, but 1) it doesn't exists in the container's own env, and 2) it's not listed in the containers own `/etc/hosts`. Hmm :thought_balloon:

ruffsl:
@tianon and @yosifkit , unless you have an opinion to improve this, I'm in favor of just simply keeping the `container_name` in the compose file as currently in this PR. It seems updating the entry-point script wont have much of an improving effect.

yosifkit:
Sounds reasonable to me! :+1:

tianon:
LGTM

ruffsl:
Thanks guys!

PR #400, opened 2015-11-21
Document correct COPY path for golang onbuild image

Starefossen:
This PR replaces the documented `/usr/src/app` path with `/usr/go/app` which is the correct path according to the [latest onbuild Dockerfile](https://github.com/docker-library/golang/blob/master/1.5/onbuild/Dockerfile#L3) for the golang image.

### DIFF
--- a/golang/content.md
+++ b/golang/content.md
@@ -16,7 +16,7 @@ The most straightforward way to use this image is to use a Go container as both 
 FROM golang:1.3-onbuild
 ```
 
-This image includes multiple `ONBUILD` triggers which should cover most applications. The build will `COPY . /usr/src/app`, `RUN go get -d -v`, and `RUN
+This image includes multiple `ONBUILD` triggers which should cover most applications. The build will `COPY . /usr/go/app`, `RUN go get -d -v`, and `RUN
 go install -v`.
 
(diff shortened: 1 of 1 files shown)


golang/content.md
```diff
@@ -16,7 +16,7 @@ The most straightforward way to use this image is to use a Go container as both
 FROM golang:1.3-onbuild
 ```
 
-This image includes multiple `ONBUILD` triggers which should cover most applications. The build will `COPY . /usr/src/app`, `RUN go get -d -v`, and `RUN
+This image includes multiple `ONBUILD` triggers which should cover most applications. The build will `COPY . /usr/go/app`, `RUN go get -d -v`, and `RUN
```
tianon: It's actually [`/go/src/app`](https://github.com/docker-library/golang/blob/7ddbd90e7e9c96ed9c7346f21cbf4142a80e37be/1.5/onbuild/Dockerfile#L3-L4), not `/usr/go/app`; want to update the PR, or would you rather I carry this change?  Thanks for pointing it out regardless! :+1:

Starefossen:
Derp! I thought I got it right 😝 I won't be able to update this for a couple of hours, so feel free to close this and push the correct change if you like.

PR #401, opened 2015-11-23
Fix headline, add blank else ## won't become a headline.

dothebart:
It seems as if ## headlines aren't rendered different to floating text at all, but ##headlines are even leaving the format in...


dothebart:
never mind ;-)

PR #402, opened 2015-11-23
Fix headline, add blank else ##foo  will leave the ## in the rendered…

dothebart:
… text. It doesn't become a headline with 2 * # however.


tianon:
LGTM

PR #409, opened 2015-11-27
Add nuxeo image documentation

dmetzler:
Hello,

We would like to provide an official Nuxeo Docker image.

Thx for reviewing.


Moghedrin:
LGTM, pending [#1223](https://github.com/docker-library/official-images/pull/1223).

tianon:
Do you think it's worth noting that the default uname/pword is `Administrator`/`Administrator` here so folks trying it out don't have to search for it buried in the main Nuxeo docs?

tianon:
LGTM :+1:

dmetzler:
Description seems to be empty in official image, is this normal ?

tianon:
Doh, looks like the machine doing that died; I've kicked off a temporary fix and so the description should be live within the next 30 minutes or so (possibly less).

PR #410, opened 2015-11-30
Update ghost

lwis:


yosifkit:
This part gets generated automatically from the `library/ghost` file in `official-images` by the @docker-library-bot.

PR #412, opened 2015-12-02
Fix instruction for without dockerfile example

rogeriopradoj:
Thanks to @wsilva! Without explicit port forwarding, the command did not work.


yosifkit:
LGTM

tianon:
Thanks!

rogeriopradoj:
My pleasure!

PR #413, opened 2015-12-02
variant-alpine: explain often missing tooling

ruudud:
As discussed in https://github.com/docker-library/golang/pull/69 -- try to explain that common tooling which is not strictly needed is often not included in Alpine images. (E.g. git in golang)

Closes docker-library/golang#69


yosifkit:
LGTM

tianon:
LGTM, thanks! :+1:

PR #421, opened 2015-12-12
Added info about additional Cassandra environment options

niclash:
The cassandra-env.sh allows for some additional environment variables to be set. Especially the memory settings, which are critical on kernels where Docker --memory constraint is not available.


tianon:
Couple of thoughts:

1. `REAMDE.md` is generated, so this additional text ought to be in `content.md` instead
2. is there any documentation from upstream we can link to so that it's clear these aren't "our" settings, but are actually official upstream knobs?

Moghedrin:
Closing based on the fact that it's been over a year with no further word.

PR #423, opened 2015-12-14
Let tag-details be more forgiving of a 404

tianon:


tianon:
This is for that window between when we merge to official-images and when the image is actually built and pushed.

Example output:

## `perl:6`

**does not exist** (yet?)

yosifkit:
LGTM

PR #426, opened 2015-12-15
Add --log-driver=none to irssi run command.

psftw:


irssi/content.md
```diff
@@ -18,26 +18,22 @@ On a Linux system, build and launch a container named `my-running-irssi` like th
...
 You omit `/etc/localtime` on Mac OS X because `boot2docker` doesn't use this file.
 
-Of course, you can name your image anything you like. In Docker 1.5 you can also use the `--read-only` mount flag. For example, on Linux:
-
-```console
-$ docker run -it --name my-running-irssi -e TERM -u $(id -u):$(id -g) \
-    --read-only -v $HOME/.irssi:/home/user/.irssi \
-    -v /etc/localtime:/etc/localtime \
-    irssi
-```
```
tianon: Why'd you axe this section?

psftw: what value does it add?

psftw:
attempt to fix CI by actually using markdownfmt.sh

psftw:
alright, one more time without README.md getting updated.  maybe I'll learn what I'm doing one day.

irssi/content.md
tianon: I think the idea was to point out that `~/.irssi` is the _only_ place irssi tries to mutate by default.

tianon:
LGTM

cc @jfrazelle

jfrazelle:
nice LGTM

PR #429, opened 2015-12-23
Added note for how to connect to mongo instance from Windows Command …

gavynriebau:
Running the command `docker run -it --link some-mongo:mongo --rm mongo sh -c 'exec mongo "$MONGO_PORT_27017_TCP_ADDR:$MONGO_PORT_27017_TCP_PORT/test"'` doesn't work from windows CMD prompt and results in the output:

> mongo: 1: mongo: Syntax error: Unterminated quoted string

This pull request includes a small note for Windows users to help them avoid this problem.

### DIFF
--- a/mongo/README.md
+++ b/mongo/README.md
@@ -39,6 +39,8 @@ $ docker run --name some-app --link some-mongo:mongo -d application-that-uses-mo
 
 ## ... or via `mongo`
 
+(Replace `'` with `"` if running in Windows CMD)
+
 ```console
 $ docker run -it --link some-mongo:mongo --rm mongo sh -c 'exec mongo "$MONGO_PORT_27017_TCP_ADDR:$MONGO_PORT_27017_TCP_PORT/test"'
 ```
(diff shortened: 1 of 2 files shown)


Moghedrin:
Just to make certain I understand this correct, in Windows I would need to type:
`$ docker run -it --link some-mongo:mongo --rm mongo sh -c "exec mongo "$MONGO_PORT_27017_TCP_ADDR:$MONGO_PORT_27017_TCP_PORT/test""`

That seems... funky.

In any case, I'm generally -1 against this. I don't feel comfortable adding this sort of note here and not everywhere in the docs that it's applicable, and I definitely don't think some funky Windows quote parsing behavior (if that's what it does) warrants a note everywhere it could come up. Rather, that's something that people working with Windows should be expected to know.

gavynriebau:
@Moghedrin: Yes, that's correct and to make matters even more confusing using single quotes work in powershell, just not the windows command prompt.

The difference in parsing is pretty small so you're probably right that windows users should just recognize it and adjust the command accordingly, I'll close the pull request.

PR #430, opened 2015-12-27
Explicitly mention you can't upgrade odoo easily

pierreozoux:
https://github.com/odoo/docker/issues/36

And yes I'm angry.
I took the decision to use odoo for my business one week before the upgrade.
And now I'm stuck at version8...
And yes, I read this paragraph about the possible upgrade, and yes, I thought it would be like any other free software out there... Easy update...

Maybe the language is not appropriate, but we have to find a way to make sure people are aware they can't upgrade!

@OdooTeam: I think this is really bad practice for free software! I stopped recommending odoo after my experience. I wish this software never existed. I lost so many brain time thinking with negative thoughts about you... I'm sad.

Maybe we should actually remove this software from Docker library? What would be the procedure?

### DIFF
--- a/odoo/README.md
+++ b/odoo/README.md
@@ -79,17 +79,14 @@ $ docker run -p 127.0.0.1:8071:8069 --name odoo3 --link db:db -t odoo
 
 Please note that for plain use of mails and reports functionalities, when the host and container ports differ (e.g. 8070 and 8069), one has to set, in Odoo, Settings->Parameters->System Parameters (requires technical features), web.base.url to the container port (e.g. 127.0.0.1:8069).
 
-# How to upgrade this image
+# Upgrade
 
-Suppose you created a database from an Odoo instance named old-odoo, and you want to access this database from a new Odoo instance named new-odoo, e.g. because you've just downloaded a newer Odoo image.
+Be advise that you can't upgrade this image to a newer version.
+For instance if you run the version 8, and you want to upgrade to version 9, you can't.
 
-By default, Odoo 8.0 uses a filestore (located at /var/lib/odoo/filestore/) for attachments. You should restore this filestore in your new Odoo instance by running
+The only way to upgrade for now is to pay [odoo services](https://upgrade.odoo.com/database/upload) and upload your data to their system (yes, it is a bit scary for the privacy of your business).
 
-```console
-$ docker run --volumes-from old-odoo -p 127.0.0.1:8070:8069 --name new-odoo --link db:db -t odoo
-```
-
-You can also simply prevent Odoo from using the filestore by setting the system parameter `ir_attachment.location` to `db-storage` in Settings->Parameters->System Parameters (requires technical features).
+If you are about to take the decision wether to use odoo or not for your business, please consifer this information that could greatly impact your business. Indeed, a non up to date software is dangerous for the security of your business also.
 
 # License
 


Yenthe666:
@pierreozoux, you're a bit overreacting in my opinion.. Ofcourse you can upgrade 8.0 to 9.0, but you'll need to pay for it. If I'm not mistaking Odoo will upgrade Odoo 8 to Odoo 9 if you buy an Enterprise contract (20€/user/month).
Do note that upgrading such a database isn't really that easy either. Odoo is open source and uses hundred of tables in a database, which tend to change every version. 

I do agree that upgrades should be made more easy and better but that being said it is a huge effort to get there. So there is room for growth but there is nobody that says you can't upgrade.

pierreozoux:
I waited 3 months before doing this PR, so I maybe overreact a bit, but
it is less than at the beginning :)

The price is something we can have different opinions about, I do agree
with that. (I'm not mad at their new pricing for some features that were
free before, I agree with this kind of business model).

But, even if I had the money, I would never send my database to this
company. I respect my users and their privacy.

I'm a sysadmin, the little I know about development is that when you
upgrade a schema, you also write a migration script. This is the
standard I know about software development (open or closed source).
I don't know how odoo is developing, but I don't see it different.

Just imagine one second WordPress announcing to the world that from now
on, you'll have to pay upgrade (with security fix) and you'll have to
upload your databases to their server. Just imagine Internet reaction to
that. (and if you have say 10 plugins, you might be close to 100 tables,
but they manage to handle it...)

I still can't understand why they call this F/LOSS... I think we should
add a 5th freedom, freedom to have security fixes!

I already said it, this is maybe the wrong phrasing, but I really think
we should warn the end user what he/she is about to do.
(Luckily I already warned 2 friends, that didn't use it because of that,
but there is a lot more people out there)

On 27-12-2015 16:45, Yenthe wrote:
> @pierreozoux, you're a bit overreacting in my opinion.. Ofcourse you can upgrade 8.0 to 9.0, but you'll need to pay for it. If I'm not mistaking Odoo will upgrade Odoo 8 to Odoo 9 if you buy an Enterprise contract (20€/user/month).
> Do note that upgrading such a database isn't really that easy either. Odoo is open source and uses hundred of tables in a database, which tend to change every version. 
> 
> I do agree that upgrades should be made more easy and better but that being said it is a huge effort to get there. So there is room for growth but there is nobody that says you can't upgrade.
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/docker-library/docs/pull/430#issuecomment-167425588
> 

-- 
I use PGP to protect our privacy, if you want to know more, you can
follow this
https://emailselfdefense.fsf.org/en/

If you have further questions, please do not hesitate to ask.
You can verify my public key here: https://keybase.io/pierreozoux

tianon:
I definitely think that this major version upgrade cost needs to be highlighted in the documentation, although this PR as-is is not what I'd like to see.

Do minor revisions upgrade without hassle, or does this apply to them as well?  Ideally this section should link to the place to get more information about the details (or provide the details directly), with the strong caveat that it is not free and a link to the fee structure and how to get the process going.

cc @aab-odoo

pierreozoux:
@tianon it just impacts major version upgrades: https://github.com/odoo/docker/issues/46

aab-odoo:
Hello @pierreozoux,

I may understand your frustration but let me clarify a few things anyway.

When I wrote this documentation, there was only one Odoo image (for
version 8). So when I spoke about upgrading, I definitely meant upgrading the
Odoo 8 version to a newer release (a new release of each version is built every
night, and we update our docker images on a regular basis to make them use
recent releases). Then, we added an image for Odoo 9, and at this point the doc
might have become unclear, sorry for that. I agree we should clarify it.

Nevertheless, even if our last version is 9, we are still maintaining versions
7 and 8. So when you say that you need to upgrade to version 9 to get security
fixes, you're wrong. Major versions contain new features, but bug fixes are
directly applied on the oldest maintained (or even no longer maintained for
critical security bugs) version containing the bug.

Migrating from a major version to another is definitely a complex process,
whatever your believe. I am not in position to discuss about our pricing, or
strategic position, but this is how it is and we have to accept it. Anyway, 
there is this [community project](https://github.com/OCA/OpenUpgrade) for the development of migration scripts.

So I will update the documentation to make it lucid, but I think that the
reasons of your anger are not really related to Docker, but rather to Odoo
policy.

Finaly, let me note that our motivations for creating a Docker image were to
provide a quick and easy access to Odoo for people wanting to discover and test
it. But this was definitely not to use it in production.

Regards

pierreozoux:
Thanks @aab-odoo for your deep insight.
I run everything on docker :) And I thank you of course for providing an image, this is really nice.

I think the most annoying part for me is not the price, but the fact you have to send your database to odoo compagny. This is crazy when you know we are on the post snowden era. People are hosting their software to avoid spying and economic intelligence, and you ask people to send all their data to your company. This is my point of view, but of course odoo is free to do what they want, and I'm free to not use it.

I'm happy that the documentation will be updated to make sure people are aware. (you can also specify that it shouldn't run in production but I don't see any reasons for that).

I'll close this PR when the doc will be updated :)

thanks everybody for your time, and sorry if I sounded angry, but I think my language was measured.

Bon bout d'an (happy piece of year) and all the best to everybody!

aab-odoo:
I understand your concerns. Note that we have an [anonymization module](https://github.com/odoo/odoo/tree/9.0/addons/anonymization) that allows you to hide confidential data before sending us your database for migration.

I'll PR today to update the doc.

Regards

PR #433, opened 2015-12-29
Handle completely uninitialized repo descriptions better by explicitly converting undef to "" (avoids warnings about "Use of uninitialized value $currentText")

tianon:


yosifkit:
LGTM

PR #434, opened 2015-12-29
Switch %%LOGO%% to use an explicit commit link so that logos update in the same way the rest of the content does

tianon:


yosifkit:
LGTM

PR #438, opened 2016-01-03
Add a link to a useful Hacker News thread and to the Alpine image page for pointing folks at how to install new packages

tianon:
See also some of the discussion following https://github.com/docker-library/python/pull/71#issuecomment-165909092 for why we need more information about what using this variant might mean for a potential user.


yosifkit:
LGTM

PR #440, opened 2016-01-04
Odoo: clarification of upgrading process

aab-odoo:
Hello,

This PR is related to #430

Regards

### DIFF
--- a/odoo/content.md
+++ b/odoo/content.md
@@ -72,6 +72,8 @@ Please note that for plain use of mails and reports functionalities, when the ho
 
 # How to upgrade this image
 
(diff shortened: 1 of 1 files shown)


tianon:
This is a great addition, but I think it also needs to reference both https://upgrade.odoo.com and https://github.com/OCA/OpenUpgrade in some way that makes it more clear that major version upgrades are non-trivial and that open-source/free upgrades may not necessarily even be possible unless/until implemented in the "Open Source Upgrade" repository.

tianon:
Actually, https://doc.therp.nl/openupgrade/ is probably better to link to since it actually talks about the important stuff (unlike that Git repo's README :smile:).

tianon:
ping @aab-odoo; have you had a chance to look at this again?

aab-odoo:
@tianon Sorry i've been quite busy lately. I just updated the PR.

yosifkit:
LGTM

tianon:
Awesome, thanks! :+1:

LGTM

PR #442, opened 2016-01-05
Fix link to official images on the hub

devth:


tianon:
Aw man, that sucks -- the other list was way better, and had decent sorting control. :disappointed: :cry:

Oh well; LGTM

PR #443, opened 2016-01-11
WIP: Feedback requested, what is a good way for users to add virtualhost files?

peoplemerge:
This patch has a trivial solution.  This first PR is intended to be open for feedback as a solution is discussed, so do not merge this first commit yet:

* patch would be needed to be installed - then later removed
* open to other ideas on how a user can change that line of config
* or suggest something else entirely?

### DIFF
--- a/nginx/README.md
+++ b/nginx/README.md
@@ -72,6 +72,28 @@ Then, build with `docker build -t some-custom-nginx .` and run:
 $ docker run --name some-nginx -d some-custom-nginx
 ```
 
+## virtualhost configuration
+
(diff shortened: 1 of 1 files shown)


peoplemerge:
possibly use `tr` instead of patch as this works

psftw:
Isn't this just convention?  I don't think nginx actually cares about folder structure for included config files, does it?

tianon:
Indeed, if you've got a configuration snippet you want included (such as a virtual host definition), adding it to `conf.d` should do the trick (or overwrite `nginx.conf` entirely). :+1:

PR #447, opened 2016-01-13
Update docs for sentry 8

mattrobenolt:
See also https://github.com/docker-library/official-images/pull/1338

/cc @tianon @yosifkit I assume the README.md file is autogenerated from these components? plz correct me if I'm wrong. :)


yosifkit:
LGTM

yosifkit:
(and yes, readme will get generated by the bot)

PR #454, opened 2016-01-15
Describe the updated Busybox variants in a little more detail

tianon:
See also https://github.com/docker-library/official-images/pull/1347 and https://github.com/docker-library/busybox/pull/7.

cc @jpetazzo


psftw:
Nice work on the image update and docs!

yosifkit:
LGTM

PR #455, opened 2016-01-15
Update details of websphere-liberty:beta

davidcurrie:
Updates for https://github.com/docker-library/official-images/pull/1351 that reflect that a) is ceased to be referred to as the v9 beta months ago and b) that as of the January beta it is created for a zip rather than an executable JAR.


tianon:
LGTM, merged by hand in 53676645728ca85ef31f4a9ba41babc7120c6a21 because I had to fix a CI issue and it wasn't worth round-tripping you over :+1:

PR #418, opened 2015-12-11
Adding an example about pecl extension

Grummfy:

PR #457, opened 2016-01-19
Add docs for PECL and "other" extensions installation in PHP

### LINKED ISSUE #323: Document docker-php-ext-enable
It looks like the `docker-php-ext-enable` script added in https://github.com/docker-library/php/pull/122 was never documented. The comments in the PR hint at standalone use of the script (`pecl install mongo && docker-php-ext-enable mongo`), so it should be documented here.

tianon:
Closes #418
Closes docker-library/php#170
Fixes #323


tianon:
@yosifkit good to go :+1:

yosifkit:
LGTM

PR #460, opened 2016-01-20
add Kong, Open-source Microservice & API Management layer

ahmadnassri:
goes hand in hand with https://github.com/docker-library/official-images/pull/1358


kong/content.md
```diff
@@ -0,0 +1,56 @@
...
+# How to use this image
+
+First, Kong requires a running Cassandra cluster before it starts. You can either use the [mashape/cassandra](https://github.com/Mashape/docker-cassandra) image, provision a test instance on [kongdb.org](http://kongdb.org) or use a cluster of your own.
+
+## 1. Link Kong to a Cassandra container
+
+Start a Cassandra container by doing so:
+
+```shell
+$ docker run -d -p 9042:9042 --name cassandra mashape/cassandra
```
thibaultCha: The `mashape/cassandra` image should be deprecated (ideally, removed) and the example should use the official `cassandra` image (which did not exist at that time), with the `2.2` tag.

thefosk:
I squashed the commits and a check that involves `TRAVIS_COMMIT_RANGE` is now failing - before squashing the tests on TravisCI were passing.

kong/content.md
```diff
@@ -0,0 +1,91 @@
...
+$ docker run -d --name kong \
+    -e "DATABASE=cassandra" \
+    --link kong-database:kong-database \
+    -p 8000:8000 \
+    -p 8443:8443 \
+    -p 8001:8001 \
+    -p 7946:7946 \
+    -p 7946:7946/udp \
+    --security-opt seccomp:unconfined \
+    mashape/kong
```
tianon: This should just be `kong` :wink:

Is `--security-opt seccomp:unconfined` actually required?  You've got a note explaining what to do if it errors out, but nothing here explaining why it's necessary or even recommended, which is much more interesting IMO. :smile:

kong/content.md
```diff
@@ -0,0 +1,91 @@
...
+$ docker run -d \
+    -v /path/to/your/kong/configuration/directory/:/etc/kong/ \
+    -p 8000:8000 \
+    -p 8443:8443 \
+    -p 8001:8001 \
+    -p 7946:7946 \
+    -p 7946:7946/udp \
+    --security-opt seccomp:unconfined \
+    --name kong \
+    mashape/kong
```
tianon: Same comment as above :smile:

thefosk:
@tianon PR updated with Kong 0.9 which doesn't require `--security-opt seccomp:unconfined`.

yosifkit:
LGTM

thefosk:
Let me know if you need anything else from my side.

PR #462, opened 2016-01-22
sudo and git are needed to build ffmepg

dmetzler:
Small enhancement to documentation


yosifkit:
I am always hesitant to recommend installing `sudo` in a container, esp when the process is already running as root, but it is up to you.  (My suggestion would be to just remove [`sudo` from your ffmeg repo](https://github.com/nuxeo/ffmpeg-nuxeo/blob/2aa906d33209374ff0e874e5fcbabca711a23fed/build-yasm.sh#L8) since you tell users to run the install scripts as root already.)

What do you think?

dmetzler:
Hello, sorry for the late answer....

The fact is that our ffmpeg build is common to different distribution packaging and it makes uses of sudo.

tianon:
Couldn't it be replaced by something like this? (to make the `sudo` usage conditional on whether or not we're already `root`/uid 0)
```bash
[ "$(id -u)" != 0 ] && sudo=sudo || sudo=
$sudo checkinstall --pkgname=yasm --pkgversion="1.2.0" --backup=no --deldoc=yes --default
```

dmetzler:
Ok, I'm doing a PR like this on ffmpeg-nuxeo.

dmetzler:
@yosifkit Can also be merged.

yosifkit:
Looks like the travis test has a few nitpicks for some extra newlines.

dmetzler:
@yosifkit  ok sorry, thought it was the docker image build that sometime randomly fails. It's ok now.

tianon:
LGTM

Issue #404, opened 2015-11-23
Modification of the "Supported Versions" text


Djelibeybi:
We need to be able to change the text added to the documentation under the "Supported Versions" heading. This is so that we can direct our customers to the appropriate support services based on whether they have a paid subscription or not.

Also, we need to be able to specify the currently supported distribution version of Docker, which may differ from the upstream release.

Djelibeybi:
Note that changing the `CONTENT.md` isn't suitable, as we'd still be left with a confusing "Supported Versions" block.

tianon:
Hmm, interesting requirements -- for the second half of that, you _should_ be able to add `oraclelinux/docker-versions.md` (which will override the contents of `.template-helpers/docker-versions.md`).

The first bit is a little more challenging, since I don't think the templating scripts as-is have a good mechanism for supporting that sort of customization.  Do you think you could whip up an example of what you'd like the output to be so we can evaluate better how best to implement/support it?

Djelibeybi:
Resolved via #464 which is sufficient for our purposes. Thanks for the tip to create `docker-versions.md`.

tianon:
Nice, glad that did it! :smile:

PR #464, opened 2016-01-22
Updated the Oracle Linux documentation

Djelibeybi:
Thanks to @tianon's tips in #404 (which is a great issue number), I've added `docker-versions.md` and rebuilt `README.md` using `update.sh`.

The new text provides more accurate support information for both Oracle support customers and users without a paid support subscription.


yosifkit:
LGTM

PR #466, opened 2016-01-23
Add example for mysqli

muuki88:
If you need a  MySQL connection inside your docker container

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -97,6 +97,19 @@ RUN apt-get update && apt-get install -y libmemcached-dev \
 	&& docker-php-ext-enable memcached
 ```
 
+### Mysqli extension
+
+If you need database access from your php app, you may need to install mysqli extensions.
+
+```dockerfile
+FROM php:5.6-apache
+# Install modules
+RUN apt-get update && apt-get install -y \
+        php5-mysql \
+        && docker-php-ext-install -j$(nproc) mysqli
+CMD ["apache2-foreground"]
+```
+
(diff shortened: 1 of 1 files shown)


php/content.md
```diff
@@ -97,6 +97,19 @@ RUN apt-get update && apt-get install -y libmemcached-dev \
...
+
+If you need database access from your php app, you may need to install mysqli extensions.
+
+```dockerfile
+FROM php:5.6-apache
+# Install modules
+RUN apt-get update && apt-get install -y \
+        php5-mysql \
+        && docker-php-ext-install -j$(nproc) mysqli
+CMD ["apache2-foreground"]
```
yosifkit: There is no need to `apt-get install php5-mysql`, php is already [compiled with `mysqlnd`](https://github.com/docker-library/php/blob/master/5.6/Dockerfile#L51), so you just need to turn it on.  That package is for the apt provided php, which is not what we use.  Just a `docker-php-ext-install mysqli` should do the trick.

tianon:
Since `mysqli` is a core extension, isn't this example really just a more specific example for the `PHP Core Extensions` section above `PECL`?

Perhaps it would be sufficient to add a blurb at the end of that section noting something like "Several common database-related extensions are `mysqli`, `pgsql`, and `pdo` (with `pdo_mysql`, `pdo_pgsql`, etc)."

Moghedrin:
No response, closing as stale.

PR #469, opened 2016-01-26
Add a note about how to remove the spurious stderr blurb in golang:onbuild

tianon:
Fixes docker-library/golang#41


yosifkit:
LGTM

PR #472, opened 2016-01-29
Mention Stack and fix inline Dockerfile.

psftw:
To go with https://github.com/docker-library/official-images/pull/1392


yosifkit:
LGTM

tianon:
LGTM

PR #478, opened 2016-02-02
Add docs for random and onetime root password

sbhw:
See PR docker-library/mysql#119 

IMO its unfortunate that there are different repositories: one containing the build definition, and this one containing the docker image README.md. Having this README.md in the build repository could help keeping the docs in sync with implementation. 

Additional change: add warning about root passwort getting published to env variable, see docker-library/mysql#88


yosifkit:
This will need to be in `content.md` because [`readme.md` is generated](https://github.com/docker-library/docs#image-namereadmemd).

Thanks for your PR.

cc @ltangvald on content, LGTM

ltangvald:
Contents look good to me

PR #482, opened 2016-02-07
Fix typo

ream88:


tianon:
Thanks for the heads up!  This needed to be over in `content.md` (since `README.md` is generated), so I've pushed the appropriate fix directly. :+1:

ream88:
Ah ok, thx anyways!

PR #486, opened 2016-02-11
Merge with standard arangodb docs

m0ppers:


arangodb/content.md
```diff
@@ -4,9 +4,7 @@ ArangoDB is a multi-model, open-source database with flexible data models for do
 
 The supported data models can be mixed in queries and allow ArangoDB to be the aggregation point for the data request you have in mind.
 
-> [arangodb.com](https://arangodb.com)
-
-%%LOGO%%
```
yosifkit: If you drop this then the logo will no longer be in the generated [`README.md`](https://github.com/docker-library/docs/blob/master/arangodb/README.md).

arangodb/content.md
```diff
@@ -4,9 +4,7 @@ ArangoDB is a multi-model, open-source database with flexible data models for do
...
+Dockerfile: [`Latest` (Dockerfile)](https://github.com/arangodb/arangodb-docker/blob/master/Dockerfile.templ)
```
yosifkit: I realize this is a sync from your other readme, but there are [links generated](https://github.com/docker-library/docs/blob/master/arangodb/README.md#supported-tags-and-respective-dockerfile-links) for each Dockerfile for each version.

arangodb/content.md
```diff
@@ -20,36 +18,29 @@ Joins and Transactions are key features for flexible, secure data designs, widel
 
 Furthermore, ArangoDB offers a microservice framework called [Foxx](https://www.arangodb.com/foxx) to build your own Rest API with a few lines of code.
 
-ArangoDB Documentation
-
--	[ArangoDB Documentation](https://www.arangodb.com/documentation)
--	[ArangoDB Tutorials](https://www.arangodb.com/tutorials)
+ArangoDB Documentation - [ArangoDB Documentation](https://www.arangodb.com/documentation) - [ArangoDB Tutorials](https://www.arangodb.com/tutorials)
```
yosifkit: I assume that you want to keep the bullet points, so I think this is an unwanted change.

arangodb/content.md
```diff
@@ -20,36 +18,29 @@ Joins and Transactions are key features for flexible, secure data designs, widel
...
 
-### Start an `ArangoDB` instance
+### Start an ArangoDB instance
 
 In order to start an ArangoDB instance run
 
-```console
-$ docker run -d --name arangodb-instance arangodb
-```
+	unix> docker run -e ARANGO_RANDOM_ROOT_PASSWORD=1 -d --name arangodb-instance -d arangodb/arangodb
```
yosifkit: Indented code blocks vs fenced will lose the color formatting based on the language specifier; either way is fine.

This should just be `arangodb` instead of `arangodb/arangodb`.  (this is applicable throughout the doc)

arangodb/content.md
```diff
@@ -20,36 +18,29 @@ Joins and Transactions are key features for flexible, secure data designs, widel
...
+Will create and launch the arangodb docker instance as background process. The Identifier of the process is printed. By default ArangoDB listen on port 8529 for request and the image includes`EXPOSE 8529`. If you link an application container it is automatically available in the linked container. See the following examples.
 
 In order to get the IP arango listens on run:
 
-```console
-$ docker inspect --format '{{ .NetworkSettings.IPAddress }}' arangodb-instance
-```
+	docker inspect --format '{{ .NetworkSettings.IPAddress }}' <IDENTIFIER>
+
+(where <IDENTIFIER> is the return string of the previous start command)
```
yosifkit: This could still be `arangodb-instance`, since that is the `--name` given, but either way is fine.

arangodb/content.md
```diff
@@ -65,65 +56,76 @@ These can be used to access the database.
...
+
+Generate a random root password when starting. The password will be printed to stdout (may be inspected later using `docker logs`)
+
+1.	ARANGO_NO_AUTH=1
+
+Disable authentication. Useful for testing.
+
+**WARNING** Doing so in production will expose all your data. Make sure that ArangoDB is not diretcly accessible from the internet!
+
+1.	ARANGO_ROOT_PASSWORD=somepassword
```
yosifkit: This will not be a continuous list unless you indent the paragraphs in between.

arangodb/content.md
```diff
@@ -65,65 +56,76 @@ These can be used to access the database.
...
-This will start a new container `some-%%REPO%%` where the ArangoDB instance uses the startup settings from your config file instead of the default one.
+And use this data container in your ArangoDB container.
 
-Note that users on host systems with SELinux enabled may see issues with this. The current workaround is to assign the relevant SELinux policy type to your new config file so that the container will be allowed to mount it:
+	unix> docker run -e ARANGO_RANDOM_ROOT_PASSWORD=1 --volumes-from arangodb-persist -p 8529:8529 arangodb
 
-```console
-$ chcon -Rt svirt_sandbox_file_t /my/custom
-```
+If want to save a few bytes you can alternatively use[tianon/true](https://registry.hub.docker.com/u/tianon/true/) or[progrium/busybox](https://registry.hub.docker.com/u/progrium/busybox/) for creating the volume only containers. Please note that you need to provide the used volumes in this case. For example
```
yosifkit: As much as I :heart: `tianon/true` we only recommend other official images; `busybox` and `alpine` are extremely small.

arangodb/content.md
```diff
@@ -65,65 +56,76 @@ These can be used to access the database.
...
 
-### Using a data container
+	unix> docker run -d --name arangodb-persist -v /var/lib/arangodb tianon/true true
 
-Alternatively you can create a container holding the data.
+## Building your own ArangoDB image
```
yosifkit: This section feels a little out of place, since this will be on the Docker Hub and not in the git repo.

m0ppers:
wow...thanks for the very helpful suggestions. I hope I got everything.

yosifkit:
Seems fine to me! :+1:

PR #487, opened 2016-02-11
Remove "-it" from Apache invocations of "docker run" (due to SIGWINCH killing Apache, by design)

tianon:
See https://bz.apache.org/bugzilla/show_bug.cgi?id=50669 for additional detail from upstream.

Fixes docker-library/php#64


yosifkit:
LGTM

PR #494, opened 2016-02-18
amend centos systemd document

wu0407:
Systemd requires the CAP_SYS_ADMIN capability. This means running docker
with --privileged.If not running with --privileged or --caps-add SYS_ADMIN,will have error:
`Failed to mount tmpfs at /run: Operation not permitted
[!!!!!!] Failed to mount API filesystems, freezing.`

Remove the systemd-update-utmp*.service in the /lib/systemd/system dir.Because some warn output:
`[ INFO ] Update UTMP about System Boot/Shutdown is not active.
[DEPEND] Dependency failed for Update UTMP about System Runlevel Changes.`


tianon:
cc @jperrin

yosifkit:
Closing old PR; feel free to comment if you feel this was closed in error.

PR #495, opened 2016-02-18
Add CouchDB documentation

klaemo:
Adds documentation for the official CouchDB image.

References: https://github.com/docker-library/official-images/pull/1288


couchdb/content.md
```diff
@@ -0,0 +1,92 @@
...
+and then build and run
+
+```console
+$ docker build -t you/awesome-couchdb .
+$ docker run -d -p 5984:5984 you/awesome-couchdb
+```
+
+## Logging
+
+By default containers run from this image only log to `stdout`. This means that the `/_log` endpoint is not available. You can enable logging to file in the [configuration](http://docs.couchdb.org/en/1.6.1/config/logging.html).
```
yosifkit: So, even though you set file logging to `/dev/null` in https://github.com/klaemo/docker-couchdb/blob/fe82d8781f99cefb4b5efbaa319b30fd4f80a73f/1.6.1/Dockerfile#L94 it still logs to `stdout`? Cool.

yosifkit:
LGTM

PR #498, opened 2016-02-22
Add a section about "--auth" to the mongo documentation

tianon:
Closes https://github.com/docker-library/mongo/pull/48


yosifkit:
LGTM

PR #503, opened 2016-02-26
Officially deprecate "rails"

tianon:
Given what it is (just a framework), and looking through https://github.com/search?l=dockerfile&q=%22FROM+rails%22&type=Code, I think it's time we finally cut the cord.

### DIFF
--- a/rails/deprecated.md
+++ b/rails/deprecated.md
@@ -1,0 +1,19 @@
+This image is officially deprecated in favor of [the standard `ruby` image](https://hub.docker.com/_/ruby/), and will receive no further updates after 2016-12-31 (Dec 31, 2016). Please adjust your usage accordingly.
+
+For most usages of this image, it was already not bringing in `rails` from this image, but actually from your project's `Gemfile`, so the only "value" being added here was the pre-installing of `nodejs`, `mysql-client`, `postgresql-client`, and `sqlite3` for various uses of the `rails` framework.
+
+For example, a `Dockerfile` similar to the following would be a good starting point for a Rails project using PostgreSQL:
+
+```dockerfile
+FROM ruby:2.3
+
+RUN apt-get update && apt-get install -y postgresql-client --no-install-recommends && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /usr/src/app
+COPY Gemfile* ./
+RUN bundle install
+COPY . .
+
+EXPOSE 3000
+CMD ["rails", "server", "-b", "0.0.0.0"]
+```


yosifkit:
Rails itself is included in the `Gemfile` of a project that uses rails so that most users might even be installing a different version of rails than the one included in `rails:latest`.  It is still useful for the first initialization of a rails project. We could help ease users over to the `ruby` image if we install the apt packages required for most rails installations (https://github.com/docker-library/rails/blob/db9122c0525e0a9f76ed47dbfb26e8e25ca0af70/Dockerfile#L7).

rails/deprecated.md
```diff
@@ -0,0 +1,19 @@
+This image is officially deprecated in favor of [the standard `ruby` image](https://hub.docker.com/_/ruby/), and will receive no further updates. Please adjust your usage accordingly.
```
yosifkit: Maybe we should give a leniency period rather than this abrupt end. How about an EOL of maybe 6 months from now?

tianon: Yeah, I was planning on doing so, but calling it out explicitly and giving it a specific EOL date probably makes sense. :+1:

tianon:
Yeah, I suppose it _is_ somewhat useful for initializing a new project...  So perhaps we just deprecate the `onbuild` version of it?  Even initializing a new project is pretty hacky, and requires funky bind-mount stuff, so I'm still neutral leaning towards -1 on keeping the image around just for that.

As for the deps, they're not actually universally required, so I think the fact that this text explicitly calls out "for a project using PostgreSQL" is probably good enough -- folks should be able to figure out that if they use MySQL, they need to include whatever their bindings need for that to work... :smile:  (but maybe those client bindings make some sense in `buildpack-deps` to a certain extent?)

tianon:
I've replaced:
```markdown
... and will receive no further updates.
```
with:
```markdown
and will receive no further updates after 2016-12-31 (Dec 31, 2016).
```

Happy to bikeshed on a different date, but I figured setting it somewhere in the far future should give a plenty big window for existing users (if we merge soon, that's over half a year / 6 months).  It seems a little long, but I didn't think this discussion was quite open and shut yet so it can serve as a concrete point to discuss around. :smile:

yosifkit:
LGTM

yosifkit:
I though about the possibility of adding the [sql clients](https://github.com/docker-library/rails/blob/81676c55499113778b097a9235bd449a754d94f4/Dockerfile#L7) from the rails image to buildpack-deps so that all language stacks can use them.  On `buildpack-deps:jessie` it is about 45MB to add them:
```console
$ docker images buildpack-deps
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
buildpack-deps      temp                031a9584969b        5 seconds ago       655.4 MB
buildpack-deps      latest              f9f7b5cfe736        10 days ago         610 MB
```

PR #504, opened 2016-02-29
clarify logo.png

psftw:
RE https://github.com/docker-library/official-images/pull/1342#issuecomment-189841143


tianon:
Arg, that rug flying out from under us.

LGTM

PR #513, opened 2016-03-02
Update README.md

Jacqs:
update on behalf of Dave Currie


yosifkit:
This file is generated from your `content.md`.  It looks like this part specifically comes from the main [`template.md`](https://github.com/docker-library/docs/blob/4b07201b836a691740049c50eb8716dd3780af97/.template-helpers/template.md).  Feel free to adjust there instead.

PR #132, opened 2014-12-31
Add memcached docs

yosifkit:
ping @tianon, @Moghedrin

PR #514, opened 2016-03-04
Document POSTGRES_INITDB_ARGS env var

knickers:
This documents the POSTGRES_INITDB_ARGS environment variable added in #132.


yosifkit:
LGTM (pending the actual feature merging to official images)

PR #515, opened 2016-03-09
Updated Oracle Linux docs with Yum Server name change

Djelibeybi:
Also switch to HTTPS for the main Oracle Linux home page.


yosifkit:
LGTM

(ignoring Travis since it is just reminding of the change to `README` which is the same as `content`)

PR #520, opened 2016-03-17
Switch busybox's logo from "buildroot" to the official logo (especially now that we don't actually use buildroot directly)

tianon:


yosifkit:
LGTM

psftw:
Checking internally how these logos are being deployed to Hub...

PR #522, opened 2016-03-17
Adjust a couple URLs to clarify the process slightly more

tianon:
As discussed in https://github.com/tianon/docker-brew-ubuntu-core/issues/41#issuecomment-184967383


yosifkit:
LGTM

PR #523, opened 2016-03-17
Update/remove golang:cross references

tianon:
Thanks for pointing this out, @fboyer! :heart:


tianon:
(see also https://github.com/docker-library/golang/issues/63)

yosifkit:
LGTM

PR #532, opened 2016-03-24
Adding an example compose version 2 file

twitherspoon:
Environment variables will only be populated if you’re using the legacy version 1 Compose file format. https://docs.docker.com/compose/link-env-deprecated/

You must set additional environment variables to launch the Wordpress docker container in version 2.

Not sure how to add these instructions into the README.md and content.md files.  If you provided steps I will add those to the documentation as well.


tianon:
Looks like we need to update the wordpress entrypoint to account for this change somehow -- perhaps by defaulting to `: ${WORDPRESS_DB_HOST:=mysql}` instead of trying to auto-detect, and hope that the later code (or WordPress itself) will capture the failure to specify. :disappointed:

PR #535, opened 2016-03-29
Add a note about the new debian/eol image

tianon:
cc @paultag

### DIFF
--- a/debian/content.md
+++ b/debian/content.md
@@ -15,6 +15,8 @@ http://httpredir.debian.org/debian testing main`).
 
 The mirror of choice for these images is [the httpredir.debian.org redirecting mirror](http://httpredir.debian.org) so that it's as close to optimal as possible, regardless of location or connection. See the [httpredir homepage](http://httpredir.debian.org) for more information.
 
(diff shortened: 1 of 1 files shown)


tianon:
I've updated the `debian/eol` image description with some color about the available Debian releases too, courtesy of the `debian-history` package. :smile:

yosifkit:
As much as this is a great idea I have to bring up the unwritten rule that we don't recommend non-official images within official images. :disappointed:  But if you have a compelling exception that can be uniformly applied...

tianon:
> `gazebo/content.md:52`:The OSRF's Docker Hub organization profile contains a Gazebo repo at [osrf/gazebo](https://registry.hub.docker.com/u/osrf/gazebo/) which is based on this repo but includes additional tags for these advanced use cases.

> `ros/content.md:42`:The rest of the common meta-packages such as `desktop` and `desktop-full` are hosted on automatic build repos under OSRF's Docker Hub profile [here](https://registry.hub.docker.com/u/osrf/ros).

So we _do_ kind of have an unwritten exception in place for "additional functionality still in the spirit of the original".

yosifkit:
That seems like a valid exception.  I thought we had something for weekly jenkins but can't find anything.

LGTM

PR #546, opened 2016-04-12
make sample start line generic

ganncamp:
Make it easier for copy/pasters to run latest version of SonarQube platform.


ganncamp:
this file turns out to be auto-generated

PR #548, opened 2016-04-13
Adding docker documentation for the TICK stack

jsternberg:
docker-library/official-images#1583


psftw:
Sent over https://github.com/influxdata/docker-docs/pull/2

kapacitor/content.md
```diff
@@ -0,0 +1,133 @@
+# Kapacitor
+
+Kapacitor is an open source data processing engine written in Go. It can process both stream and batch data.
```
psftw: Need a bit more context here

influxdb/content.md
```diff
@@ -0,0 +1,123 @@
...
+Find more about API Endpoints & Ports [here](https://docs.influxdata.com/influxdb/latest/concepts/api/).
+
+### Configuration
+
+InfluxDB can be either configured from a config file or using environment variables. To mount a configuration file and use it with the server, you can use this command:
+
+Generate the default configuration file:
+
+```console
+$ docker run --rm influxdb influxd config > influxdb.conf
```
psftw: This doesn't work anymore with the latest changes, but I think it should.

kapacitor/content.md
```diff
@@ -0,0 +1,133 @@
...
+
+```console
+docker run -p 9092:9092 \
+    -v kapacitor:/var/lib/kapacitor \
+    kapacitor
+```
+
+### Configuration
+
+Kapacitor can be either configured from a config file or using environment variables. To mount a configuration file and use it with the server, you can use this command:
```
psftw: reword --  "use this command:" followed by a set of instructions/commands

kapacitor/content.md
```diff
@@ -0,0 +1,133 @@
...
+
+Generate the default configuration file:
+
+```console
+$ docker run --rm kapacitor kapacitord config > kapacitor.conf
```
psftw: This doesn't work with the latest changes, but I think it should.

kapacitor/content.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+Modify the default configuration, which will now be available under `$PWD`. Then start the Kapacitor container.
+
+```console
+$ docker run -p 9092:9092 \
```
psftw: double-check this command

kapacitor/content.md
```diff
@@ -0,0 +1,133 @@
...
+KAPACITOR_REPORTING_ENABLED=false
+```
+
+Find more about configuring Kapacitor [here](https://docs.influxdata.com/kapacitor/latest/introduction/installation/)
+
+### Exposed Ports
+
+-	9092 TCP -- HTTP API endpoint
+
+#### Subscriptions
```
psftw: I found Kapacitor to be the more confusing component when initially looking at TICK.  I don't find this updated section to be really helpful.  It suggests starting influxdb after Kapacitor which seems backwards, and it also uses older style linking vs. the built in service discovery which is inconsistent with the Telegraf docs.  The syntax for environment variable substitution when lists are involved could be explained as it isn't intuitive yet the URLS setting seems to be pretty important.

kapacitor/content.md
```diff
@@ -0,0 +1,133 @@
...
+
+# Start the InfluxDB container linked to Kapacitor
+docker run -p 8083:8083 -p 8086:8086 \
+    --name influxdb \
+    --link=kapacitor \
+    -v influxdb:/var/lib/influxdb \
+    influxdb
+```
+
+Modify `$HOSTNAME` with a valid hostname for connecting to InfluxDB.
```
psftw: Not very helpful instructions?

kapacitor/content.md
```diff
@@ -0,0 +1,133 @@
...
+Run another container linked to the `kapacitor` container for using the client. Set the env `KAPACITOR_URL` so the client knows how to connect to Kapacitor. Mount in your current directory for accessing TICKscript files.
+
+```console
+$ docker run --rm --link=kapacitor \
+    -e KAPACITOR_URL=http://kapacitor:9092 \
+    -v $PWD:/root -w=/root -it \
+    kapacitor bash
+```
+
+Then from within the container you can use the `kapacitor` command to interact with the daemon. Assuming you have Telegraf + InfluxDB + Kapacitor all hooked up, then create a file `cpu_alert.tick`. If not then see the [docker-compose](https://docs.docker.com/compose/) [TICK stack](https://github.com/influxdata/TICK-docker) environment for easy setup.
```
psftw: This link to influxdata/TICK-docker has nothing to do with docker-compose.

kapacitor/content.md
```diff
@@ -0,0 +1,133 @@
...
+
+Kapacitor can be either configured from a config file or using environment variables. To mount a configuration file and use it with the server, you can use this command:
+
+Generate the default configuration file:
+
+```console
+$ docker run --rm kapacitor kapacitord config > kapacitor.conf
+```
+
+Modify the default configuration, which will now be available under `$PWD`. Then start the Kapacitor container.
```
psftw: An example excerpt of a config file as part of the instructions for running kapacitor would be preferred over the later instructions to configure it manually via bash-in-container.

influxdb/content.md
```diff
@@ -0,0 +1,123 @@
...
+Start the container:
+
+```console
+$ docker run --name=influxdb -d -p 8083:8083 -p 8086:8086 influxdb
+```
+
+Run the influx client in another container:
+
+```console
+$ docker run --rm --link=influxdb -it influxdb influx -host influxdb
```
jsternberg: This only sometimes works due to a bug in the library that takes care of the line input that I'm fixing in peterh/liner#75 or we'll fork the dependency so we can get the fix. There's also a second bug that causes this to occasionally enter an infinite loop and I'll submit that bugfix to influxdb after the other PR gets merged. We'll definitely have this fix in for 0.13.

influxdb/content.md
jsternberg: Works again.

kapacitor/content.md
jsternberg: Fixed this so it works again.

kapacitor/content.md
jsternberg: When you say built in service discovery, do you mean `--net=container:influxdb` or something else?

jsternberg:
I updated the Kapacitor documentation, squashed the commits, and rebased on the latest. I also removed the references to the TICK docker-compose file since it simplified the README.

There's an issue with the influx binary for 0.12 so that won't work properly for now. The issue will be fixed in 0.13, but we would still like it if we could get 0.12 merged in either way.

kapacitor/content.md
tianon: I believe @psftw is referring to `docker network` (which is what is slowly replacing `--link`); see https://docs.docker.com/engine/userguide/networking/dockernetworks/

jsternberg: I've updated the documentation to use `docker network` and remove all references to `--link`.

psftw:
Thanks.  Latest changes look good.

tianon:
LGTM

yosifkit:
LGTM

PR #552, opened 2016-04-18
Remove reference to knickers/mongo-express

knickers:
This removes a reference to knickers/mongo-express I had missed in an example, and changes it to the official mongo-express.


tianon:
Man, shame on you, but even more shame on us for not noticing.  Thanks for fixing it!

LGTM :+1:

PR #557, opened 2016-04-26
Fix typo in README.md

anthonycollini:


tianon:
Thanks! :metal:

LGTM

tianon:
Since this affects just about every `docker-library/` repo, I'm going to commit all those fixes directly. :+1:

tianon:
Now we have a fun list of references of all the places this could've been caught sooner. :smile:

PR #564, opened 2016-05-07
Add spiped official image

TimWolla:
see docker-library/official-images#1714

### DIFF
--- a/spiped/README-short.txt
+++ b/spiped/README-short.txt
@@ -1,0 +1,1 @@
+Spiped is a utility for creating symmetrically encrypted and authenticated pipes between sockets.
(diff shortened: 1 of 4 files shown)


TimWolla:
I just force pushed to add TimWolla/docker-spiped#7.

tianon:
Content LGTM; I'm likely to just carry this for you at this point, especially given how long it's been. :+1:

(unless you really have some pressing desire to update it yourself -- either way I'd like to get `spiped` merged today if @yosifkit is as happy with the Dockerization as I am :+1:)

yosifkit:
content LGTM

PR #571, opened 2016-05-12
Updating the TICK stack documentation

jsternberg:
Fixing some small inconsistencies between them so now the documentation
is more uniform and all of the commands work correctly and are formatted
the same.


yosifkit:
LGTM

jsternberg:
I updated a single line removing the reference to the Raft port that was removed in 0.12.

yosifkit:
LGTM

tianon:
:+1:

PR #572, opened 2016-05-13
Remove excessive complexity from update.sh

yosifkit:
- remove `otherRepos` array from update.sh to be a file within each image folder
  - removes the need to edit update.sh on new images; keeps an image self-contained within its folder
- drop `mailing-list` support since we already have `user-feedback.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,16 +12,23 @@ All Markdown files here are run through [tianon's fork of `markdownfmt`](https:/
 -	create a `README-short.txt` (required, 100 char max)
 -	create a `content.md` (required)
 -	create a `license.md` (required)
+-	create a `github-repo` (required)
 -	add a `logo.png` (recommended)
--	edit `update.sh` as needed (see below)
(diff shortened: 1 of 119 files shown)


README.md
```diff
@@ -12,16 +12,23 @@ All Markdown files here are run through [tianon's fork of `markdownfmt`](https:/
 -	create a `README-short.txt` (required, 100 char max)
 -	create a `content.md` (required)
 -	create a `license.md` (required)
+-	create a `github-repo` (required)
```
tianon: I think we should probably either add `.txt` on this file, or remove `.txt` on `README-short` :cry:

tianon: Well, I guess not.  This is sane.

tianon: Because `README-short` is explicitly "text", where this is a URL, so it's just a bit of extra metadata.

tianon:
LGTM, although we should add a test for this file to `.travis.yml` too :smile:

yosifkit:
Travis added and passes.

tianon:
Nice, LGTM :+1:

PR #574, opened 2016-05-15
Update http links to https

bwalding:
Checked that HTTPS was working on links I updated (was somewhat surprised they were all functional!)
Also changed example `mycompany.com` to `mycompany.example.com` to make it non-routable

@reviewbybees


tianon:
This needs to be moved to `content.md`; `README.md` is generated. :+1:

PR #576, opened 2016-05-17
A a new entry for Silverpeas

mmoqui:
see PR https://github.com/docker-library/official-images/pull/1729


silverpeas/content.md
```diff
@@ -0,0 +1,179 @@
...
+-	the Oracle database system.
+
+The Silverpeas images support actually only the two first database systems; because of the non-free licensing issues with the Oracle JDBC drivers, Silverpeas cannot include these drivers by default and consequently it cannot use transparently an Oracle database system as a persistence backend.
+
+For the same reasons, the Docker images of Silverpeas aren't shipped with the Oracle JVM but with the default OpenJDK. Silverpeas uses the Wildfly application server as runtime.
+
+The Silverpeas images use the following environment variables to set the database access parameters:
+
+-	`DB_SERVERTYPE` to specify the database system to use with Silverpeas: POSTGRESQL for PostgreSQL, MSSQL for Microsoft SQLServer, ORACLE for Oracle. By default, it is set to POSTGRESQL.
+-	`DB_SERVER` to specify the IP address or the name of the host on which the database system is running. By default, it is set to `localhost`.
```
tianon: It's probably worth considering changing this default from `localhost` to `db` or `database` or something so that either Docker networks or container linking (`--link some-postgres:database`) can work out-of-the-box (especially given that looking for `localhost` in a container doesn't make much sense).

silverpeas/content.md
```diff
@@ -0,0 +1,179 @@
...
+These environment variables can be also defined as properties into the Silverpeas global configuration file `config.properties` (see below).
+
+## Start a Silverpeas instance with a database from a Docker container
+
+In [Docker Hub](https://hub.docker.com/), no Docker images of Microsoft SQLServer are currently available, but you will find a lot of images of PostgreSQL. For example, with an [official PostgreSQL docker image](https://hub.docker.com/_/postgres/), you can start a PostgreSQL instance initialized with a superuser `postgres` with as password `mysecretpassword`:
+
+	$ docker run --name postgresql -d \
+	    -e POSTGRES_PASSWORD="mysecretpassword" \
+	    -v /var/lib/postgresql/data:/var/lib/postgresql/data \
+	    postgres
```
tianon: I think this should probably be an explicit `postgres:9.6` (instead of relying on `:latest).

silverpeas/content.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Start a Silverpeas instance with the default configuration
+
+Finally, a Silverpeas instance can be started by specifying the required database access parameters with the environment variables. In the example, the database is named `Silverpeas` and the priviledged user is `silverpeas` with as password `thesilverpeaspassword`:
+
+	$ docker run --name silverpeas -p 8080:8000 -d \
+	    -e DB_SERVERTYPE="POSTGRESQL" \
+	    -e DB_SERVER="database" \
+	    -e DB_NAME="Silverpeas" \
+	    -e DB_USER="silverpeas" \
```
tianon: The default user for the `postgres` image is `postgres`, isn't it?  Either this should be `postgres` or the above `docker run` should include `-e POSTGRES_USER=silverpeas`, right? (maybe also `-e POSTGRES_DB=Silverpeas` for good measure?)

silverpeas/content.md
tianon: Do you think it'd be good to switch this bind-mount to a named volume like is done below for the `silverpeas` application container instead of a bind mount? (love the idea of using named volumes in the docs :+1:)

silverpeas/content.md
```diff
@@ -0,0 +1,179 @@
...
+	    -v /etc/silverpeas/config.properties:/opt/silverpeas/configuration/properties \
+	    silverpeas \
+	    /bin/true
+
+Then to mount the volumes in the Silverpeas container:
+
+	$ docker run --name silverpeas -p 8080:8000 -d \
+	    --link postgresql:database \
+	    --volumes-from silverpeas-store \
+	    silverpeas
```
tianon: What's the value of using a data container over just using an explicit `docker volume create`?

silverpeas/content.md
```diff
@@ -0,0 +1,179 @@
...
+
+# Document conversion
+
+Some features in Silverpeas (export, preview, content visualization, ...) requires a document converter. The document conversion is performed in Silverpeas by the program LibreOffice running as a daemon. So, in order to enable and to use these features, you have first to use a Data Volume Container to store all the Silverpeas data and second to run a container embbeding a LibreOffice program running as a daemon. There is no official Docker images of LibreOffice but DockerHub hosts some of unofficial images of it ([xcgd/libreoffice](https://hub.docker.com/r/xcgd/libreoffice/) for example).
+
+Once a Data Volume Container created for Silverpeas as explained in the section above, you have to link it with the Docker image running LibreOffice as a daemon in order the program have access the documents to convert:
+
+	$ docker run --name libreoffice -d \
+	    --volumes-from silverpeas-store \
+	    xcgd/libreoffice
```
tianon: The Silverpeas image still contains libreoffice, right?  Could that be invoked here instead of depending on a non-official image? (https://github.com/docker-library/official-images/blob/20311d4ee3ecf4f832e405089f06ce83dd1972e5/README.md#repeatability)

ie, something like:
```console
$ docker run --name libreoffice -d ... silverpeas libreoffice --nologo --norestore --invisible --headless --accept='socket,host=0,port=8997,tcpNoDelay=1;urp;'
```

silverpeas/content.md
mmoqui: Done.

silverpeas/content.md
mmoqui: Done.

silverpeas/content.md
mmoqui: In the example given in the document, PostgresSQL is setup with the usual superuser account 'postgres'. The container can be used for any application other than Silverpeas and the instantiation of the container is just given as an example. For Silverpeas, it is recommended then to create a dedicated database and to create a user account with administrative right for that database.

silverpeas/content.md
mmoqui: The Data Container will contain in a single unit all the volumes into which Silverpeas store its data and activity. Those volumes are relative to each other at a given time of use of Silverpeas (they can be a snapshot of the activity of Silverpeas). It is then in my own point of view more easy to move and manage a data container with several volumes. (In my point of view, creating a docker volume is useful only for containers sharing and storing persistent data into a single volume.)

mmoqui: Hum, any remarks about my response?

mmoqui:
Any remarks about my answers?

silverpeas/content.md
tianon: I'm not sure how to feel about this -- I don't think we have many examples of using data containers in the official images, but it should be mostly harmless.  :+1:

silverpeas/content.md
```diff
@@ -0,0 +1,180 @@
...
+
+	$ docker run --name postgresql -d \
+	    -e POSTGRES_PASSWORD="mysecretpassword" \
+	    -v postgresql-data:/var/lib/postgresql/data \
+	    postgres:9.6
+
+We recommend strongly to mount the directory with the database file on the host so the data won't be lost when upgrading PostgreSQL to a newer version (a Data Volume Container can be used instead). For any information how to start a PostgreSQL container, you can refer its [documentation]((https://hub.docker.com/_/postgres/).
+
+Once the database system is running, a database for Silverpeas has to be created and a user with administrative rights on this database should be added; it is recommended to create a dedicated user account in the database for each application and therefore for Silverpeas. In this document, and by default, a database `Silverpeas` and a user `silverpeas` for that database are created.
```
yosifkit: Wouldn't it be easier for users to just use `-e POSTGRES_USER=silverpeas` and `-e POSTGRES_DB=Silverpeas` on the postres container creation?  When running a postgres container, most users create a new container for each application that needs a database, rather than having one database container for all of them.

silverpeas/content.md
```diff
@@ -0,0 +1,180 @@
...
+
+By default, some volumes are created inside the container, so that we can access them in the host.(Refers the [Docker Documentation](https://docs.docker.com/engine/tutorials/dockervolumes/#locating-a-volume) to locate them.) Among them `/opt/silverpeas/log` and `/opt/silverpeas/data`: the first volume contains the logs produced by Silverpeas whereas the second volume contains all the data that are created and managed by the users in Silverpeas. Because the latter has already a directories structure created at image creation, a host directory cannot be mounted into the container at `opt/silverpeas/data` without losing the volume's content (the mount point overlays the pre-existing content of the volume). In our example, in order to easily locate the two volumes, we label them explicitly with respectively the labels `silverpeas-log` and `silverpeas-data`. (Using a [Data Volume Container](https://docs.docker.com/engine/userguide/containers/dockervolumes/) to map `/opt/silverpeas/log` and `/opt/silverpeas/data` is a better solution.)
```
yosifkit: Should there be any mention of the default user/password of `SilverAdmin` or navigating to `http://localhost:8080/silverpeas` to view it?

silverpeas/content.md
mmoqui: For a development or a testing purpose, and for which the security isn't important, yes it is just easier.

For a production environment, running a RDBMS with a single database for each application they have seams to me an overloaded approach in which they don't use all the benefit of such a RDBMS. 
Besides, the credentials passed through the `POSTGRES_USER` and `POSTGRES_DB` environment variables are those of the RDBMS administrator. And in a pure security concern, even for a single application, it is a bad idea to use such an account with an application. It is strongly recommended to create a user account for each application and to grant to this user account only the administrative rights on the database(s) used by the application (and not on the whole RDBMS).

silverpeas/content.md
mmoqui: Yes. I'm going to update doc with this information

mmoqui: Done

PR #581, opened 2016-05-19
Update registry repo github url

yosifkit:
This is still pointing to the old `docker/registry` repo.

@RichardScothern, is `docker/distribution-library-image` the correct repo for users to file issues that they have with the registry image or would `https://github.com/docker/distribution` be better?


RichardScothern:
This change looks good - it matches other projects with similar setups (e.g. swarm)

PR #593, opened 2016-06-06
Update "generate-dockerfile-links-partial.sh" to utilize new bashbrew features for parsing manifest files

tianon:


tianon:
cc @yosifkit

yosifkit:
LGTM

PR #595, opened 2016-06-07
Odoo: force postgres version in docker command line

odony:
The odoo docker image is based on debian:jessie, which only has
postgresql-client-9.4, while the official postgres docker
has 9.5 server. This causes a version mismatch and errors when using the
postgres command-line tools, such as `pg_dump`.

Thanks!

Fixes odoo/odoo#12275
Fixes odoo/docker#49
Fixes odoo/docker#54
Fixes odoo/docker#59


yosifkit:
LGTM

PR #596, opened 2016-06-08
Switch the github-repo for influxdb, kapacitor, and telegraf

jsternberg:
Removed the `docker exec` example since it doesn't appear to work because of
docker/docker#8755 and added a clarification to how to use the admin interface.


yosifkit:
LGTM pending https://github.com/docker-library/official-images/pull/1823

PR #604, opened 2016-06-16
Fix rethinkdb/rethinkdb-dockerfiles#31

stuartpb:
The move to the official RethinkDB organization wasn't previously reflected in the docs.


tianon:
Whoops, good catch! :+1:

PR #606, opened 2016-06-17
Added very important notice about security

COLABORATI:
I believe it would be better to completely dump this image, as it leads many not so experienced people into a really evil trap, but a warning on top is the best thing you can do right now.


tianon:
The environment variables exist for the convenience of pre-configuring the database.  Their use is optional (save for one of `MYSQL_ROOT_PASSWORD`, `MYSQL_ALLOW_EMPTY_PASSWORD`, and/or `MYSQL_RANDOM_ROOT_PASSWORD`), and I think it would indeed be appropriate to encourage production deployments to either set them on an "initialization" container, and then stop the container and re-use the volume in a new container which does not include the variables, or to change the credentials entirely once the container is running.

The fact that they are shared through to explicitly linked containers (via `--link`) is both standard Docker behavior (https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/#environment-variables), and deprecated thanks to improvements in Docker networking (https://docs.docker.com/engine/userguide/networking/dockernetworks/) where environment variables are _not_ shared.

PR #607, opened 2016-06-17
Fix typo in rabbitmq: wiht -> with

josephfrazier:


tianon:
Nice catch, LGTM :+1:

PR #608, opened 2016-06-17
ROS: Update for version 2 of docker-compose

agakshat:
Changed the docker-compose part to reflect the new version 2 of docker-compose. Some changes like [docker-compose "--x-networking" up -d] were not running at all, while others would soon be deprecated.


tianon:
Seems sane to me!

cc @ruffsl

ruffsl:
Ah, I've been meaning to get around to updating that, Thanks @agakshat ! LGTM

PR #611, opened 2016-06-20
use google's dns as well

hutchic:
need a second dns specified otherwise we can't resolve the interenet

### DIFF
--- a/consul/content.md
+++ b/consul/content.md
@@ -173,7 +173,7 @@ $ docker run -d --net=host -p 53:8600/tcp -p 53:8600/udp consul agent -bind=<bri
 Now start another container and point it at Consul's DNS, using the bridge address of the host:
 
 ```console
-$ docker run -i --dns=<bridge ip> -t ubuntu sh -c "apt-get update && apt-get install -y dnsutils && dig consul.service.consul"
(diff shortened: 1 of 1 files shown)


tianon:
cc @slackpad

slackpad:
Hi @hutchic this might be better done by adding a [`-recusor=8.8.8.8`](https://www.consul.io/docs/agent/options.html#_recursor) example. I think having multiple servers defined but not equivalent (one Internet and one .consul) would possibly result in multiple queries during normal operation, which isn't super ideal.

hutchic:
@slackpad sure that would work I think that's what `progrium/consul` does. As currently coded / documented people aren't going to have a fun time following the docs verbatim

tianon:
@hutchic did you want to update your PR with that change? :smile: :pray:

hutchic:
I'm doubtful I will. I'm not actually using the container and if I was I'd just use the change from this PR over patching the actual image.

The way I see it this documentation change makes the docs correct / work now. I'd humbly suggest merge this change and make an issue on consul repository to add a `-recursor=8.8.8.8` at which time nuke this change from the docs

slackpad:
I'll take a swing at adding a recursor config and testing it out.

hutchic:
@slackpad just following up. If you added the recursor config we can probably close this PR

PR #621, opened 2016-06-28
Add Clear Linux Project documentation

bryteise:
Create documentation to go with https://github.com/docker-library/official-images/pull/1721.


clearlinux/README-short.txt
```diff
@@ -0,0 +1 @@
+Official docker build of Clear Linux OS for Intel Architecture
```
tianon: The image itself isn't Intel-only, is it?

tianon: (for context, this is the blurb that'll show up in `docker search` and at the top of the Hub description page)

tianon:
I think this is a good start, but I'd love it if we could add some examples of how to use the image a little bit (for example, a simple `Dockerfile` showing how to `FROM clearlinux:base` and install a few packages or something similarly trivial).

tianon:
Ok, managed to find https://clearlinux.org/documentation/bundles_overview.html, which IMO would be good to link to. :+1:

Here's essentially the kind of simple `Dockerfile` I'm thinking would be worthwhile to add:
```dockerfile
FROM clearlinux:base
RUN swupd bundle-add editors
```
(which I've just verified works well)

clearlinux/README-short.txt
bryteise: It's the official name we use (either OS or Project, in this case OS made more sense to me).

bryteise:
I'll start working on flushing out the documentation then, and thanks for the test!

yosifkit:
LGTM

PR #622, opened 2016-06-28
Add vanity DockerCon variations of hello-world

tianon:
See also https://github.com/docker-library/official-images/pull/1899 and https://github.com/docker-library/hello-world/pull/16.


yosifkit:
LGTM

PR #624, opened 2016-06-30
Adjust persistent volume path

m0ppers:
Paths have changed with the new version..

see https://github.com/docker-library/official-images/pull/1893


tianon:
Nice :+1:

PR #628, opened 2016-07-05
Update README.md

tomreay:
The readme currently says that to generate a new secret key you should run:

```
$ docker run --rm sentry generate-secret-key
```

But it seems this has changed, the following works for me

```
$ docker run --rm sentry config generate-secret-key
```


tianon:
cc @mattrobenolt

mattrobenolt:
:+1:

PR #632, opened 2016-07-06
Init docs repo for Eggdrop

vanosg:


eggdrop/content.md
```diff
@@ -0,0 +1,37 @@
...
+
+%%LOGO%%
+
+# How to use this image
+
+## First Run
+
+To run this container the first time, you'll need to pass in, at minimum, a nickname and server via Environmental Variables. At minimum, a docker run command similar to
+
+	docker run -ti -e NICK=FooBot -e SERVER=irc.freenode.net eggheads/eggdrop:latest
```
yosifkit: `s!eggheads/eggdrop:latest!eggdrop:latest!`

Might want to mention that `-i` is required even when running with `-d` on the `docker run`, since the process will exit if there is no `stdin`.  I am not sure if `-t` is required or useful as well on a background eggdrop container.

yosifkit:
LGTM

tianon:
LGTM

PR #636, opened 2016-07-07
Upgrade owncloud's docker-compose.yml to version 2

tianon:
Closes #619 (carried)


tianon:
Another case where it's slightly easier to review via https://github.com/docker-library/docs/pull/636/files?w=1

yosifkit:
LGTM

PR #637, opened 2016-07-11
typo directory name

m0ppers:


arangodb/content.md
```diff
@@ -100,7 +100,7 @@ unix> docker run -e ARANGO_RANDOM_ROOT_PASSWORD=1 arangodb arangod --help
 
 ## Persistent Data
 
-ArangoDB use the volume `/var/lib/arangodb3` as database directory to store the collection data and the volume `/var/lib/arangodb-apps3` as apps directory to store any extensions. These directories are marked as docker volumes.
+ArangoDB use the volume `/var/lib/arangodb3` as database directory to store the collection data and the volume `/var/lib/arangodb3-apps` as apps directory to store any extensions. These directories are marked as docker volumes.
```
tianon: If this is wrong, the `Dockerfile`'s `VOLUME` definition is as well: https://github.com/arangodb/arangodb-docker/blob/fc8a624249c8916376b863e58ab2584bed3715d5/jessie/3.0.1/Dockerfile#L46
```dockerfile
VOLUME ["/var/lib/arangodb3", "/var/lib/arangodb-apps3"]
```

m0ppers:
yes...it is....I changed that as well for the upcoming release....internally arangodb was always using `/var/lib/arangodb3-apps`. So whenever someone really used that directory it resulted in immediate failure.

Here is the original issue: https://github.com/arangodb/arangodb-docker/issues/25

And the fix:

https://github.com/arangodb/arangodb-docker/commit/e0e217da1be0febc271f26672c2e33c7ecd365d3

I don't know but I think backporting it to the older releases is probably a bad idea isn't it?

btw thanks for your _VERY_ thorough work. It is very appreciated :)

yosifkit:
LGTM

yosifkit:
No need to worry about backporting the fixed volume to other versions of 3.0; users should be able to just add a new `-v` argument and it will work fine.

PR #638, opened 2016-07-13
Add explanation to php for `docker-php-source`

yosifkit:
Depends on https://github.com/docker-library/official-images/pull/1942

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -47,7 +47,7 @@ FROM php:5.6-apache
 COPY src/ /var/www/html/
 ```
 
-Where `src/` is the directory containing all your php code. Then, run the commands to build and run the Docker image:
+Where `src/` is the directory containing all your PHP code. Then, run the commands to build and run the Docker image:
 
 ```console
 $ docker build -t my-php-app .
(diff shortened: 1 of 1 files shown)


php/content.md
```diff
@@ -68,6 +68,15 @@ Where `src/` is the directory containing all your php code and `config/` contain
 
 We provide the helper scripts `docker-php-ext-configure`, `docker-php-ext-install`, and `docker-php-ext-enable` to more easily install PHP extensions.
 
+In order to keep the images smaller, the php source is kept in a compressed tar file. To facilitate linking of the php source with any extension, we also provide the helper script `docker-php-source` to easily extract the tar or delete the extracted source. Note: if you do use `docker-php-source` to extract the source, be sure to delete it in the same layer of the docker image.
```
tianon: s/`the php source`/`PHP's source`/g

php/content.md
```diff
@@ -68,6 +68,15 @@ Where `src/` is the directory containing all your php code and `config/` contain
...
+
+```Dockerfile
+FROM php:5-apache
```
tianon: This should be `5.6` like all the rest so that we encourage users to use `X.Y`. :+1:

tianon: :astonished: the `fpm` example below is using `:5` too :disappointed:

yosifkit: :open_mouth: you found where I took it from.

tianon:
LGTM

yosifkit:
Your changes look good to me. :+1:

PR #640, opened 2016-07-14
Add explanation to php for `docker-php-composer-install`

mathroc:
Depends on docker-library/php#260 making it into docker-library/official-images


yosifkit:
Closing since companion PR was closed. Discussion on installing composer in php image by default is continued in https://github.com/docker-library/php/issues/344.

PR #653, opened 2016-08-09
Remove broken imagelayers badges

tianon:
It would appear that CenturyLinkLabs has no interest in updating ImageLayers for v2 images, and thus no longer loads our images. :cry:


yosifkit:
:sob: LGTM

PR #655, opened 2016-08-09
Store logo 'requirements'.

psftw:


tianon:
Ouch, that's really low resolution.

psftw:
That's what I'm told is required by Store, and I'm hoping to get them to support this repository as the source of truth for logo content in an automated way vs. the cluster that exists today.  If you want to debate the merits of image size, you know who to CC here ;-).

tianon:
I'm happy to auto-generate a `logo-256.png` for every image, if that'd be helpful. :+1:

psftw:
OK, closing this as I was mislead to believe that 256x256 was a requirement, and I'm hearing Hub won't receive any support in this regard anyway, so there is no point in adding requirements on the front end.

tianon:
Fair enough, thanks @psftw! :heart:

PR #659, opened 2016-08-11
Update Ubuntu description to remove "ubuntu:12.04" and add "ubuntu:16.04" in the example "/etc/apt/sources.list" content section

tianon:


yosifkit:
LGTM

PR #660, opened 2016-08-11
Add new explicit "openjdk" docs (and deprecate "java")

tianon:
https://github.com/docker-library/official-images/pull/2046


yosifkit:
LGTM

PR #666, opened 2016-08-17
Add "variant-windowsservercore" for discussing our new "windowsservercore" variants and how to use them

### LINKED ISSUE #665: Discuss "windowsservercore" variants
We added `windowsservercore` to `golang`, but we don't discuss it at all, and that should be fixed. :smile:

tianon:
Fixes #665

image

### DIFF
--- a/.template-helpers/variant-windowsservercore.md
+++ b/.template-helpers/variant-windowsservercore.md
@@ -1,0 +1,8 @@
+## `%%REPO%%:windowsservercore`
+
+This image is based on [Windows Server Core (`microsoft/windowsservercore`)](https://hub.docker.com/r/microsoft/windowsservercore/). As such, it only works in places which that image does, such as Windows 10 Professional/Enterprise (Anniversary Edition) or Windows Server 2016.
+
+For information about how to get Docker running on Windows, please see the relevant "Quick Start" guide provided by Microsoft:
+
+-	[Windows Server Quick Start](https://msdn.microsoft.com/en-us/virtualization/windowscontainers/quick_start/quick_start_windows_server)
+-	[Windows 10 Quick Start](https://msdn.microsoft.com/en-us/virtualization/windowscontainers/quick_start/quick_start_windows_10)


tianon:
This is obviously just a start, but we gotta start somewhere. :smile:

(`update.sh` will automatically add this file's contents to any repo which includes a `windowsservercore` variant)

cc @taylorb-microsoft @jhowardmsft -- is this content OK?  Are those links good?

.template-helpers/variant-windowsservercore.md
```diff
@@ -0,0 +1,8 @@
+## `%%REPO%%:windowsservercore`
+
+This image is based on [Windows Server Core (`microsoft/windowsservercore`)](https://hub.docker.com/r/microsoft/windowsservercore/). As such, it only works in places which that image does, such as Windows 10 (with insider previews) or Windows Server 2016.
+
+For information about how to get "Docker on Windows" installed and configured, please see the relevant "Quick Start" guide provided by Microsoft:
```
jhowardmsft: Nit but might be a tiny bit clearer (as I always end up reading it as `Docker for Windows` which is obviously something quite different) if you put ...'about how to get Docker running on Windows, please see the relevant'...

@neilpeterson is our docs guy, but I think those links are OK :innocent:

jhowardmsft: Doh, and maybe @neilpeterson can arrange FW-equivalent links?

taylorb-microsoft:
Awesome sauce
Windows 10 shipped already so you can remove the caveat of 'Insider previews'

tianon:
@taylorb-microsoft even the `Containers` feature is shipped?  The "Quick Start" guide still mentions "One physical computer system running a Windows 10 Insiders release." and "Windows 10 (insiders build 14372 and up)"

.template-helpers/variant-windowsservercore.md
tianon: Indeed, that makes sense, but I'll wait to give @neilpeterson a chance to comment as well before I update. :heart:

(These guides are simple and awesome, so I'm glad they're a good place to link! :smile:)

tianon:
Also, IIRC, @yosifkit tried to do `Enable-WindowsOptionalFeature -Online -FeatureName containers -All` the other day on his stock Windows 10 install and it errored out (presumably because he isn't running an insiders build)

.template-helpers/variant-windowsservercore.md
jhowardmsft: Perhaps I should read them one day :smiling_imp:

tianon:
(Oh, and in case you hadn't seen yet, https://hub.docker.com/_/golang/ does have `windowsservercore` tags now :metal:)

.template-helpers/variant-windowsservercore.md
tianon: Hah!  They download Docker directly from https://get.docker.com now instead of preview builds via aka.ms, and they're actually a lot shorter than they appear at a glance (something like 4-5 actual commands to run). :+1:

tianon:
(Also, https://github.com/docker-library/official-images/pull/2067 is a new one today :smile:)

neilpeterson:
I've updated our docs to reference Windows 10 Anniversary Edition.

@tianon - Other than Taylor's suggestion, this looks good to me, links are good. Thanks for promoting the docs.

neilp

tianon:
:tada: :balloon: thanks!  Updated!

(The later part of that doc still mentions "One physical computer system running a Windows 10 Insiders release." but the main paragraph is indeed updated. :+1:)

yosifkit:
LGTM

PR #669, opened 2016-08-22
Adjust "httpd" examples to use "-d" ("caught SIGWINCH, shutting down gracefully" on terminal resize)

tianon:
See https://github.com/docker-library/httpd/issues/9.


tianon:
Arg, this isn't actually enough.  Signals are still forwarded even without `-t`.

tianon:
Updated to just use `-d` instead.

yosifkit:
LGTM

PR #670, opened 2016-08-22
Note where the Ubuntu tarballs come from and what "latest" points to

### LINKED ISSUE #382: tag latest
When will the tag latest on docker hub refer to more recent versions of ubuntu, instead of refering to 14.04? Or will it always refering to the latest LTS release?

tianon:
Closes #382


yosifkit:
LGTM

PR #672, opened 2016-08-22
Add a blurb about using "mysql" to connect to non-Docker or remote instances

tianon:
Closes https://github.com/docker-library/mysql/issues/137

cc @ltangvald


ltangvald:
LGTM :)

PR #674, opened 2016-08-24
Update README.md

huntc:
When mounting a volume for HAProxy's configuration, make the entire haproxy configuration directory available. This way, if the `haproxy.cfg` file is re-written, HAProxy can pick up the new file.


huntc:
Closing given https://github.com/docker-library/docs/pull/675

PR #675, opened 2016-08-24
Mount the /etc/haproxy dir instead of the file

huntc:
When mounting a volume for HAProxy's configuration, make the entire haproxy configuration directory available. This way, if the haproxy.cfg file is re-written, HAProxy can pick up the new file.

### DIFF
--- a/haproxy/content.md
+++ b/haproxy/content.md
@@ -44,7 +44,7 @@ $ docker run -d --name my-running-haproxy my-haproxy
 ## Directly via bind mount
 
 ```console
-$ docker run -d --name my-running-haproxy -v /path/to/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro haproxy:1.5
+$ docker run -d --name my-running-haproxy -v /path/to/etc/haproxy:/usr/local/etc/haproxy:ro haproxy:1.5
 ```
 
 ### Reloading config
--- a/haproxy/content.md
+++ b/haproxy/content.md
@@ -43,6 +43,8 @@ $ docker run -d --name my-running-haproxy my-haproxy
 
 ## Directly via bind mount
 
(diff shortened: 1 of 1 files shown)


yosifkit:
I am unsure on this. While I would normally recommend mounting a folder so that you can edit the file locally and see the changes reflected in the container, doing so here masks a folder of default error files that a user might reply upon.

```console 
$ docker run -it --rm haproxy ls -l /usr/local/etc/haproxy/
total 4
drwxr-sr-x 2 root staff 4096 Aug 15 22:41 errors
$ docker run -it --rm haproxy ls -l /usr/local/etc/haproxy/errors
total 32
-rw-r--r-- 4 root staff 188 Aug 15 22:41 400.http
-rw-r--r-- 4 root staff 189 Aug 15 22:41 403.http
-rw-r--r-- 4 root staff 213 Aug 15 22:41 408.http
-rw-r--r-- 4 root staff 186 Aug 15 22:41 500.http
-rw-r--r-- 4 root staff 205 Aug 15 22:41 502.http
-rw-r--r-- 4 root staff 213 Aug 15 22:41 503.http
-rw-r--r-- 4 root staff 195 Aug 15 22:41 504.http
-rw-r--r-- 4 root staff 301 Aug 15 22:41 README
```

huntc:
As it is right now caused us to believe that there was a problem with this container. It is quite normal to rewrite the config when updating HAProxy. In fact this is the only way HAProxy can be informed of new configuration - 1. Write config file 2. Send SIGHUP.

Perhaps we should also make the user aware that they need to populate the host /etc/haproxy folder in its entirety.

yosifkit:
> Perhaps we should also make the user aware that they need to populate the host /etc/haproxy folder in its entirety.

That seems valid. :+1:

yosifkit:
LGTM, cc @tianon

tianon:
There was a conflict so I carried this change (with attribution) directly in https://github.com/docker-library/docs/commit/bf40b859df81689b44b4aeb4eb96e123633696ea. :+1:

PR #677, opened 2016-08-26
hello.asm->hello.c rename

wulas:
according to the new structure of repo - new filename is hello.c


yosifkit:
LGTM

tianon:
Arg, nice catch! :+1:

PR #679, opened 2016-08-31
deprecate glassfish

psftw:
short and sweet?  not sure what else to add since there isn't a great path forward for users.

context: https://github.com/docker-library/official-images/pull/1334


tianon:
Given that `glassfish/server` is an upstream initiative, I think we should probably link to https://github.com/docker-library/official-images/pull/1334#issuecomment-243676135 for context and recommend users move to https://hub.docker.com/r/glassfish/server/ directly.

psftw:
Good point, will tweak it, though I can't fully endorse that image.  FWIW, about 1-3 dozen unique users accessing this in an average week.

psftw:
OK, should be good now.

yosifkit:
LGTM

tianon:
Thanks @psftw :heart:

PR #682, opened 2016-09-01
Source Mage: updated urls to official domain

vaygr:
We finally moved our offsite to the origin domain, updating it for Docker Hub.


yosifkit:
LGTM

PR #685, opened 2016-09-07
Update 2 sr3fp11

dinogun:
Bump up the version to 1.8.0_sr3fp11, musl glibc should be musl libc.


tianon:
Nice, LGTM! :+1:

dinogun:
Thanks @tianon for the quick merge !

PR #688, opened 2016-09-12
Remove unused email alias.

psftw:
Also removed from the private repo behind docs.docker.com (though not published yet at the time of writing).


tianon:
LGTM

yosifkit:
LGTM

PR #690, opened 2016-09-13
Added the "-it" options to some docs examples for jenkins

lucapalano:
When I use the **"docker run ..."** command I always add the **"-it"** options. I found them particular useful because I'll be able to use **ctrl+p+q** to detach and **"docker attach ..."** to reattach. So, I think that user will find this tip useful and will avoid the use of **ctrl+c** that causes the stop of the entire container.

### DIFF
--- a/jenkins/content.md
+++ b/jenkins/content.md
@@ -11,13 +11,13 @@ For weekly releases check out [`jenkinsci/jenkins`](https://hub.docker.com/r/jen
 # How to use this image
 
 ```console
-docker run -p 8080:8080 -p 50000:50000 jenkins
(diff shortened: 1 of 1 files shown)


yosifkit:
Ping Jenkins image maintainer: @carlossg.

carlossg:
we build the hub doc from the image readme, so it would have to be at https://github.com/jenkinsci/docker/blob/master/README.md
although I think it is more of docker knowledge than something to add to each image

yosifkit:
Closing so that changes can be added over in `jenkinsci/docker` first.

PR #693, opened 2016-09-16
Add an official image for electrode

caoyangs:


caoyangs:
Can someone review this please?

yosifkit:
Really sorry for the slow reply. :disappointed:

I don't see a [PR to official-images](https://github.com/docker-library/official-images/pulls?utf8=%E2%9C%93&q=is%3Apr%20electrode), but would like to understand what `electrode` is before I suggest you work more on another PR.  How would you classify electrode? Is it like rails and django or more like groovy or WordPress.  Would a user be likely to do `FROM electrode` for their project or even `docker run -d electrode` and use it as a service?

PR #697, opened 2016-09-20
Add Jenkins "update.sh" badges to our docker-library repo README stubs for better community visibility

tianon:
cc @yosifkit

PR #698, opened 2016-09-20
Add README link/badge for our Jenkins "update.sh and push.sh" job

tianon:
cc @yosifkit


tianon:
See also https://github.com/docker-library/docs/pull/697.

PR #699, opened 2016-09-23
Adding Docs for Eclipse Mosquitto Official Image

daudetCA:


eclipse-mosquitto/content.md
```diff
@@ -0,0 +1,42 @@
+# What is Eclipse Mosquitto?
+
+Eclipse Mosquitto is an open source implementation of a server for version 3.1 and 3.1.1 of the MQTT protocol. Main homepage: http://mosquitto.org/
+
+%%LOGO%%
+
+# How to use this image
+
+##Mount Points
```
tianon: I think this ought to have a single space between `##` and `Mount` (same with other headers below)

eclipse-mosquitto/content.md
```diff
@@ -0,0 +1,42 @@
...
+
+Three mount points have been created in the image to be used for configuration, persistent storage and logs.
```
tianon: Calling these "mount points" might be a little confusing for users -- in my mind, "mount points" would imply that these are defined as volumes, but they're not.

Perhaps just calling them "directories" or "locations" would be better?

eclipse-mosquitto/content.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Note**: If a volume is used, the data will persist between containers.
+
+##Run
+
+Run a container using the new image:
+
+	docker run -it -p 1883:1883 -p 9001:9001 -v mosquitto.conf:/mosquitto/data/mosquitto.conf -v /mosquitto/data -v /mosquitto/log eclipse-mosquitto
+
+:boom: if the mosquitto configuration (mosquitto.conf) was modified to use non-default ports, the docker run command will need to be updated to expose the ports that have been configured.
```
tianon: Heh, this `:boom:` likely renders in GitHub, but won't render on Docker Hub, unfortunately. :disappointed:

eclipse-mosquitto/content.md
daudetCA: Good point. I've changed it to `Directories`.

eclipse-mosquitto/content.md
daudetCA: Oh ok...made it a note instead.

PR #700, opened 2016-09-25
Adds Nextcloud

pierreozoux:


tianon:
Content seems reasonable, but Travis has some formatting tweaks you'll need to make.

(An easy way to apply those suggestions directly would be to use the `markdownfmt.sh` script in this repo, ie `./markdownfmt.sh -w nextcloud/content.md`)

pierreozoux:
Thanks for feedback, I think I fixed it (still failing, but about the README, so it is a safe fail? Right?).

Let me know what are the nest steps to move this forward :)

nextcloud/content.md
```diff
@@ -0,0 +1,52 @@
...
+# How to use this image
+
+This image is designed to be used in a micro-service environment. It consists of the Nextcloud installation in an [php-fpm](https://hub.docker.com/_/php/) container. To use this image it must be combined with any webserver that can proxy the http requests to the FastCGI-port of the container.
+
+## Start Nextcloud
+
+Starting Nextcloud php-fpm instance listening on port 9000 is as easy as the following:
+
+```console
+$ docker run -d indiehosters/nextcloud
```
yosifkit: This should be just `nextcloud` (ie, rm the `indiehosters/`).

nextcloud/content.md
```diff
@@ -0,0 +1,52 @@
...
+
+## ... via [`docker-compose`](https://github.com/docker/compose)
+
+The recommended minimal setup is using this image in combination with two containers: A database container and a reverse proxy for the http connection to the service. A working example can be found at [IndieHosters/Nextcloud](https://github.com/indiehosters/nextcloud).
+
+If you want to access your Nextcloud from the internet we recommend configuring your reverse proxy to use encryption (for example via [let's Encrypt](https://letsencrypt.org/))
+
+## Update to a newer version
+
+To update your Nextcloud version you simply have to pull and start the new container.
```
yosifkit: This would only work when using a persistent `/var/www/html/data`, right?  That or the user would have to delete `version.php` in their `/var/www/html` folder, since that would be the only way for the `tar -c` to happen.  Or maybe I am missing something.

nextcloud/content.md
pierreozoux: corrected :)

nextcloud/content.md
pierreozoux: Yes indeed. Everything is the same as ownCloud, in fact, I just added this update section, but now it is linked to the "Persistent data" section.

pierreozoux:
Do you think we can get this merged in 2016? :P

yosifkit:
Looks like markdownfmt wants a change here:
```diff
-To update your Nextcloud version you simply have to pull and start the new container.
-(if you mounted the full `/var/www/html`, you need to delete `version.php`. See [persistent data](#persistent-data) for mor information.)
+To update your Nextcloud version you simply have to pull and start the new container. (if you mounted the full `/var/www/html`, you need to delete `version.php`. See [persistent data](#persistent-data) for mor information.)
```
Or if you want them separate paragraphs, then they need a newline between them. Also, `s/mor/more/`.

pierreozoux:
Ok fixed.

PR #708, opened 2016-09-30
Add docker-compose file for redmine

yosifkit:
```
$ docker-compose up
..... [much truncate]
redmine_1  | [2016-09-30 23:28:56] INFO  WEBrick 1.3.1
redmine_1  | [2016-09-30 23:28:56] INFO  ruby 2.2.5 (2016-04-26) [x86_64-linux]
redmine_1  | [2016-09-30 23:28:56] INFO  WEBrick::HTTPServer#start: pid=1 port=3000
```

### DIFF
--- a/redmine/docker-compose.yml
+++ b/redmine/docker-compose.yml
@@ -5,15 +5,17 @@ services:
   redmine:
     image: redmine
     ports:
-      - 8080:80
+      - 8080:3000
     environment:
-      REDMINE_DB_MYSQL: mysql
+      REDMINE_DB_MYSQL: db
       REDMINE_DB_PASSWORD: example
     depends_on:
-      - mysql
+      - db
+    restart: always
 
-  mysql:
+  db:
     image: mariadb
     environment:
       MYSQL_ROOT_PASSWORD: example
       MYSQL_DATABASE: redmine
+    restart: always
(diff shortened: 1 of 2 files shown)


redmine/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: '2'
+
+services:
+
+  redmine:
+    image: redmine
+    ports:
+      - 8080:80
```
tianon: Uhhh, I think this should be 3000: https://github.com/docker-library/redmine/blob/48c809da106963f2f373ba0250a03e9b6daa8bd3/Dockerfile.template#L89

redmine/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    environment:
+      REDMINE_DB_MYSQL: mysql
```
tianon: Do you think it'd be smart to change this name to `db` or something so that it's more clear that this is a hostname?  I probably have updated that script myself and was still scratching my head at what this was for a bit. :innocent:

redmine/docker-compose.yml
yosifkit: You think I would have noticed the output as well: `WEBrick::HTTPServer#start: pid=1 port=3000`

redmine/docker-compose.yml
yosifkit: That seems like a good idea.

PR #711, opened 2016-10-03
Fix links to Neo4j documentation

spacecowboy:
Both will be 404 soon (if not already).


tianon:
LGTM

(Travis failure is due to `README.md` having changes -- they match the `content.md` changes, however.)

spacecowboy:
@tianon thanks. reading the instructions,

> Note: do not actually commit the README.md file; it is automatically generated/committed before being uploaded to Docker Hub.

Would it make sense for us to untrack the generated README file to avoid CI failures in the future?

PR #714, opened 2016-10-07
Update content.md

liamawhite:
@davidcurrie


tianon:
This seems sane to me -- @davidcurrie what're your thoughts?

davidcurrie:
LGTM (but then I was stood behind him when he wrote it!)

tianon:
@davidcurrie hah! Makes sense! Thanks for confirming :smile:

PR #718, opened 2016-10-11
Deprecate "hipache" due to upstream inactivity

tianon:
cc @dmp42 @samalba -- any objections to deprecating the `hipache` official image?

We don't plan to remove the artifacts (in case folks are still using them), but it does seem prudent to let our users know the status of the project.

If there aren't objections to the deprecation, then I'd appreciate suggestions for improvement on the wording of the deprecation notice, especially whether the "potentially similar solutions" list is even in the right ballpark. :innocent: :heart:


samalba:
I don't have any objection. For the deprecation notice, let's just be transparent and mention the development of the project stopped. I can update the README of the main project as well.

For alternatives, I guess vulcand is the closest one.

tianon:
Nice, updated the PR to list `vulcand` first (specifically, the `mailgun/vulcand` image since that appears to be maintained by upstream).

I think a note in the upstream hipache README would be a nice touch. :+1:

jpetazzo:
Note: we could also mention traefik since it is similar to vulcand (and more actively maintained).

tianon:
@jpetazzo indeed, as-is it lists traefik after vulcand :smile: :+1:

PR #719, opened 2016-10-12
Add mention of password in the run command for Neo4j Image

callicles:
The neo4j image is able to change the password of the neo4j user on start up. The syntax is not very clear from the documentation.

### DIFF
--- a/neo4j/content.md
+++ b/neo4j/content.md
@@ -14,6 +14,7 @@ You can start a Neo4j container like this:
 docker run \
     --publish=7474:7474 --publish=7687:7687 \
     --volume=$HOME/neo4j/data:/data \
+    -e NEO4J_AUTH=neo4j/<password> \
     neo4j
 ```
 


yosifkit:
cc @benbc, @spacecowboy, and @srbaker, maintainers of the neo4j image.

spacecowboy:
@callicles please raise this issue on https://github.com/neo4j/docker-neo4j

callicles:
@spacecowboy Should I make a pull request there ?

spacecowboy:
@callicles I think we would prefer it if you outlined your suggested fix in an issue [here](https://github.com/neo4j/docker-neo4j) (the docker docs being here makes it kind of special).

But I can mention that we intentionally do not mention a lot of detail and instead link to our documentation: http://neo4j.com/docs/operations-manual/current/deployment/single-instance/docker/

There are several other capabilities in the container which are not mentioned in the readme for the same reason.

benbc:
@yosifkit It seems that the problem is really that the Docker store page for Neo4j has an out-of-date version of the README: https://store.docker.com/images/5f0be9a7-f5d7-4974-9c0e-78c33484e79c. That has been updated on the Hub page (https://hub.docker.com/_/neo4j/).

yosifkit:
As for the Docker Store page content, we are not directly involved: https://github.com/docker-library/official-images/pull/2232#issuecomment-252764305.

spacecowboy:
The docker store page has now been updated so I think this can be closed @yosifkit

PR #722, opened 2016-10-13
Add Adminer official image

TimWolla:
docker-library/official-images#2252


adminer/content.md
```diff
@@ -0,0 +1,27 @@
+# Adminer
+
+## What is Adminer?
+
+Adminer (formerly phpMinAdmin) is a full-featured database management tool written in PHP. Conversely to phpMyAdmin, it consist of a single file ready to deploy to the target server. Adminer is available for MySQL, PostgreSQL, SQLite, MS SQL, Oracle, Firebird, SimpleDB, Elasticsearch and MongoDB.
```
tianon: This is a pretty impressive list of supported databases!  I noticed that the image only installs `pdo_mysql`, `pdo_pgsql`, and `pdo_sqlite` -- does this image still support this full list, or would it need additional extensions to do so? (not necessarily advocating for additional extensions to be installed unless you think it's prudent -- if it doesn't support the others though, which database types this _does_ support probably ought to be noted in this document somewhere)

adminer/content.md
```diff
@@ -0,0 +1,27 @@
...
+
+### FastCGI
+
+If you are already running a FastCGI capable web server you might prefer running adminer via FastCGI:
+
+	$ docker run --link some_database:db -p 9000:9000 adminer:fastcgi
+
+Then point your web server to port 9000 of the container.
+
+Note: This exposes the FastCGI socket to the Internet. Make sure to add proper firewall rules to prevent a direct access.
```
tianon: Probably worth recommending Docker networks or something here too, ie, "Make sure to add proper firewall rules or use a private Docker network instead to prevent direct public access."

adminer/content.md
TimWolla: It would need additional extensions. At the very least MS SQL needs an extension. Thanks for the heads up, I'll check.

TimWolla: https://github.com/docker-library/docs/pull/722/files#diff-d05d6a3bde988e5f8307b2149876b1d3R29

adminer/content.md
TimWolla: Done :+1:

TimWolla:
No idea why the build failed after force pushing, but I clearly did not modify a `README.md`: https://travis-ci.org/docker-library/docs/builds/195995505

yosifkit:
No problem.  Sometimes the test gets confused by changes to master when it tests the merge.

PR #734, opened 2016-10-26
Add a section to Elasticsearch about host setup (especially the necessary sysctl)

tianon:
Closes https://github.com/docker-library/elasticsearch/issues/98
See also https://github.com/docker-library/elasticsearch/pull/127


tianon:
Semi-related to https://github.com/docker-library/official-images/pull/2288 (slightly higher importance given that 5.0 is now GA).

PR #737, opened 2016-10-28
centos add missing \ in content

kokspflanze:
https://github.com/docker-library/docs/blob/master/centos/content.md#dockerfile-for-systemd-base-image there is a `\` missing


yosifkit:
That looks like it was supposed to be a single line, but I'll let the maintainer of centos, @jperrin, decide on the fix.

tianon:
ping @jperrin -- is this change OK, or would you rather see this as a single line?

jperrin:
it was originally supposed to be a single line, @yosifkit is correct. I'm fine with it being amended with the missing \ however.

:thumbsup:

PR #746, opened 2016-11-11
on people's demand

openenergyprojects:


tianon:
Can you please elaborate a bit on the driving forces behind this PR?

Moghedrin:
Stale, no response as to why this change should be implemented. Closing for now.

openenergyprojects:
the previous examples do not have persistence.
I've noticed in few places on the net people asking what are the right folders to mount outside.
I provide this example from a production env, tested for quite some time. Kindly help others by adding this example to the main branch

PR #749, opened 2016-11-22
Update openSUSE description

flavio:
Mention latest available release.
Fix links to dockerfiles.


yosifkit:
This readme file is generated from the `content.md`, so changes need to be made there.  As for the "Supported tags and respective `Dockerfile` links" section, it is generated from the contents of the `library/` file in [`docker-library/official-images`](https://github.com/docker-library/official-images).

flavio:
Awesome, I didn't know about that. Then there's nothing to be done here :)

PR #751, opened 2016-11-23
Update doc about an error of CLI command

zepouet:
Just correct a minor error about CLI.

Thanks

### DIFF
--- a/influxdb/content.md
+++ b/influxdb/content.md
@@ -114,7 +114,7 @@ $ docker run --name=influxdb -d -p 8086:8086 influxdb
 Run the influx client in another container:
 
 ```console
-$ docker run --rm --net=container:influxdb -it influxdb influx -host influxdb
+$ docker run --rm --link=influxdb -it influxdb influx -host influxdb
 ```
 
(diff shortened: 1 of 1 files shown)


yosifkit:
This needs to go in `content.md`; `README.md` is generated.

tianon:
It's worth noting that `--link`, while discouraged (given that it doesn't work cross-host) is not officially deprecated yet.

I personally like `--link` over `--net container:xxx` because it shows a more general pattern that's easier to adjust for other needs (such as creating a second interactive container in an existing cross-host `docker network` which is on a separate physical host from the initial container, thus making `--net container:xxx` impossible to use), and makes it clear that "localhost" connections aren't treated as special in some way (like MySQL can be configured to do, for example).

jsternberg:
Ok. Whatever you guys think would be the clearer documentation on the matter. The reason why I used `--net=container:xxx` was because I thought that was the preferred method and I also show examples of setting up a network later on. Adding an example of using `--link` sounds fine to me in that case.

tianon:
Thanks @jsternberg (and @zepouet)!

PR #720, opened 2016-10-12
Update odoo usage including docker-compose

tiangolo:
Update the Odoo instructions including how to use with Docker Compose.

---

The current version depends on the legacy linking system, and as a legacy feature it doesn't work with the newer Docker networks and Docker Compose.

I included a couple `docker-compose.yml` example "contents" and explanations of the needed variables (that were provided before by the Docker legacy linking system).

I also updated the Postgres version in the documentation to use the current latest.

---

I also submitted a PR to the [Odoo Docker Image](https://github.com/odoo/docker/pull/74) to set sensible defaults, enable using the new tools (Docker networks, Docker Compose, etc) and don't depend on the legacy linking system while still being backwards compatible. 

If that PR gets accepted I will create a new PR here updating the docs again with the new options and env vars.

But as of now there are no instructions on how to use Odoo with the new Docker tools. So this PR should address that.

PR #752, opened 2016-11-28
Odoo: document usage of docker-compose

sle-odoo:


sle-odoo:
@tiangolo Hello,
Do you agree with my changes? If so i'll squash the commits and merge this to docker-docs.
Thank you in advance

tiangolo:
Excellent @sle-odoo ! 

I haven't been able to check the issue with the newest versions of Postgres you commented in https://github.com/docker-library/docs/pull/720#issuecomment-262517704 but it's clearly the safest bet to just use `9.4` as you are doing.

Thanks for taking the time to review and update it! It looks good to me.

tianon:
Nice, looks sane to me!

sle-odoo:
thanks @tiangolo and @tianon

PR #754, opened 2016-11-29
Correct image name in commands

katzien:
This PR corrects a small typo in the commands that run the config file check and bind the config file. They should use the name of the image ("my-proxy") instead of "haproxy:1.5".

This issue seems to have tripped up a few folks (myself included) and the correction for the config file check command was already pointed out in the comments, but never made it to the docs themselves :)


katzien:
Crap. I just realised this needs to go in content.md instead of README.md. I'll fix this and resubmit the PRs, sorry about that.

tianon:
@katzien feel free to simply rebase/force-push this PR instead of opening a new one -- GitHub will automatically update the PR when you do so :wink:

haproxy/README.md
```diff
@@ -56,10 +56,12 @@ $ docker run -it --rm --name haproxy-syntax-check haproxy:1.5 haproxy -c -f /usr
 $ docker run -d --name my-running-haproxy my-haproxy
 ```
 
+You may need to publish the ports your haproxy is listening on to the host by specifying the `-p` option, for example `-p 80:80` to publish port 80. Make sure the port you're using is free.
```
tianon: It'd probably be good if we used something like `-p 8080:80` instead of just `80:80` so it's clear that the ports do not have to be the same (it's a small change that IMO makes a huge difference in understanding).  ie:

> You may need to publish the ports your haproxy is listening on to the host by specifying the `-p` option, for example `-p 8080:80` to publish port 8080 from the container host to port 80 in the container. Make sure the port you're using is free.

haproxy/README.md
```diff
@@ -56,10 +56,12 @@ $ docker run -it --rm --name haproxy-syntax-check haproxy:1.5 haproxy -c -f /usr
...
+
 ## Directly via bind mount
 
 ```console
-$ docker run -d --name my-running-haproxy -v /path/to/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro haproxy:1.5
+$ docker run -d --name my-running-haproxy -v /path/to/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro my-haproxy
```
yosifkit: This one doesn't need to be `my-haproxy` since it is using a bind mount rather than doing the `docker build`.

haproxy/README.md
katzien: Hey @tianon ! Thanks for the suggestion, that's a good one! I think this PR has gone a bit crazy because this note was actually added in a separate PR - https://github.com/docker-library/docs/pull/755 and I'm not sure why/how sneaked in here - sorry about that! So I've updated it according to what you suggested in the other PR :)

katzien:
@tianon closing this PR as it was from my master branch, and that wasn't the best idea in hindsight...

PR #755, opened 2016-11-29
Add a note about publishing haproxy's input port to the host

katzien:
This has been mentioned in the comments twice, so I thought I'll add it for people to save them time wondering why their port configuration doesn't seem to be working :)


katzien:
Crap. I just realised this needs to go in content.md instead of README.md. I'll fix this and resubmit the PRs, sorry about that.

tianon:
Thanks @katzien! :+1:

PR #758, opened 2016-11-29
Update influxdb documentation to remove web interface as a default setting

jsternberg:
The web interface is deprecated and is now disabled by default. It will
be removed in the future so stop encouraging its use.


jsternberg:
As a note, the web interface is still defaulted to on in only the docker image of `influxdb:1.1.0`. I'm going to be disabling it by default for `influxdb:1.1.1` and this change is for that. It doesn't really matter when it's merged though.

tianon:
:heart:!

PR #765, opened 2016-12-09
Update about config volume

robfrank:
Adds note about providing an empty config volume, due to some bad questions on SO and GH


robfrank:
sorry, wrong file

PR #767, opened 2016-12-10
Update PHP memcached example to properly compile again

### LINKED ISSUE #764: PHP PECL memcached example from docs does not work
https://github.com/docker-library/docs/tree/master/php#php-core-extensions

```
FROM php:7.0-fpm
RUN apt-get update && apt-get install -y libmemcached-dev \
    && pecl install memcached \
```

Output: https://gist.github.com/mariusgrigaitis/5bc04c0bab82aed2b64502edc0438570

tianon:
Fixes #764


tianon:
The released versions of the `memcached` extension don't support PHP 7 yet, and we ought to be pinning to a specific version. :+1:

(Also, new versions seem to need ZLIB too :sweat_smile:)

PR #774, opened 2016-12-16
Adds Google DNS recursor to DNS example.

slackpad:
This should close https://github.com/docker-library/docs/pull/611.


tianon:
Nice, LGTM :+1:

PR #778, opened 2016-12-19
Add upstream quote about elastic 5.x vm.max_map_count check

yosifkit:
Related to https://github.com/docker-library/elasticsearch/issues/98.

### DIFF
--- a/elasticsearch/content.md
+++ b/elasticsearch/content.md
@@ -21,6 +21,10 @@ See the following sections of the upstream documentation for more information:
 -	[Setup Elasticsearch » Important System Configuration » Virtual memory](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/vm-max-map-count.html)
 -	[Setup Elasticsearch » Bootstrap Checks » Maximum map count check](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/_maximum_map_count_check.html)
 
+This [comment in elastic/elasticsearch#4978](https://github.com/elastic/elasticsearch/issues/4978#issuecomment-258676104) shows why this change was added in upstream. Production mode is when listening on non-localhost.
+
+> Elasticsearch will not start in production mode if `vm.max_map_count` is not high enough. [...] If the value on your system is NOT high enough, then your cluster is going to crash and burn at some stage and you will lose data.
+
 ## Running Containers
 
 You can run the default `elasticsearch` command simply:


yosifkit:
Reading further in https://github.com/elastic/elasticsearch/issues/4978, I found this:
> You can set `http.host` independently of `transport.host` (`network.host` sets both); the bootstrap checks are only enforced if a node is able to form a cluster with other instances not on localhost
> \- https://github.com/elastic/elasticsearch/issues/4978#issuecomment-258842584

So maybe, instead we change our default provided config to not set `network.host` and instead just `http.host` and print a warning in the logs and here in the docs to let users know how to set it up for clustering.

yosifkit:
So, it will work with a smaller `vm.max_map_count` on the host;  the only drawback is that it won't be able to cluster with other nodes, since that part is now only on localhost of the container.

```console
$ sudo sysctl -w vm.max_map_count=262144
$ docker run -it --rm elacticsearch:5
...
[2016-12-20T00:47:34,773][INFO ][o.e.b.BootstrapCheck     ] [vvWlFNU] bound or publishing to a non-loopback or non-link-local address, enforcing bootstrap checks
... works (no warning or error)
$ sudo sysctl -w vm.max_map_count=262000
$ docker run -it --rm elacticsearch:5
...
[2016-12-20T00:32:39,125][INFO ][o.e.b.BootstrapCheck     ] [YoNYX3Z] bound or publishing to a non-loopback or non-link-local address, enforcing bootstrap checks
ERROR: bootstrap checks failed
max virtual memory areas vm.max_map_count [262000] is too low, increase to at least [262144]
... fails
$ # build new elasticsearch:5 with 
diff --git a/5/config/elasticsearch.yml b/5/config/elasticsearch.yml
-network.host: 0.0.0.0
+http.host: 0.0.0.0
$ # vm.max_map_count still set to 262000
$ docker run -it --rm new-elacticsearch:5
...
[2016-12-20T00:36:44,447][INFO ][o.e.t.TransportService   ] [UdTHcL7] publish_address {127.0.0.1:9300}, bound_addresses {[::1]:9300}, {127.0.0.1:9300}
[2016-12-20T00:36:44,452][WARN ][o.e.b.BootstrapCheck     ] [UdTHcL7] max virtual memory areas vm.max_map_count [262000] is too low, increase to at least [262144]
... SUCCESS! (just a warning)
```

tianon:
+1 sounds great!

That's exactly what I wanted when I set this config in the first place;
mind commenting on that other issue and opening a PR? 👍

PR #783, opened 2016-12-22
Added haxe images documentation.

andyli:
See https://github.com/docker-library/official-images/pull/2478


haxe/content.md
```diff
@@ -0,0 +1,63 @@
...
+
+# How to use this image
+
+The most straightforward way to use this image is to use a Haxe container as both the build and runtime environment. In your `Dockerfile`, writing something along the lines of the following will compile and run your project:
+
+```dockerfile
+FROM haxe:3.4
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
```
tianon: The `RUN mkdir -p ...` here is no longer necessary (much older versions of Docker used to have issues with that sort of thing, but they've been fixed). :+1:

PR #785, opened 2016-12-27
Update haproxy docs to reference version 1.7

cblecker:
haproxy 1.7 is now stable, so updating the docs to reference it


tianon:
:+1:

PR #796, opened 2017-01-10
Update Registry content for v2

tianon:
This removes the Registry v1 content completely -- no v1 tags are supported anymore, and we only need to document what actually is still supported.

Closes https://github.com/docker-library/official-images/issues/2524

cc @stevvooe @dmcgowan

I left in the bits about "Docker Trusted Registry" at the end, but I was kind of wondering if those ought to be updated to talk about the fact that "Docker Trusted Registry" is now part of "Docker Datacenter" (and isn't sold standalone anymore AFAIK; the link we've got redirects to the DDC info) -- thoughts?

### DIFF
--- a/registry/README-short.txt
+++ b/registry/README-short.txt
@@ -1,1 +1,1 @@
-Containerized docker registry
+The Docker Registry 2.0 implementation for storing and distributing Docker images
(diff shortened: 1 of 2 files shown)


tianon:
I also figured this is already documented well elsewhere, so took the approach of including one of the most basic examples from the docs, and then linking outwards rather than duplicating too much of that info here.

stevvooe:
LGTM

@tianon Thank you for doing this!!

PR #801, opened 2017-01-19
[nats-streaming] Release v0.3.8

kozlovic:
Essentially a bug fix release.
Details can be found [here](https://github.com/nats-io/nats-streaming-server/releases/tag/v0.3.8)

### DIFF
--- a/nats-streaming/content.md
+++ b/nats-streaming/content.md
@@ -17,38 +17,38 @@ $ docker run -d nats-streaming
 
 Output that you would get if you had started with `-ti` instead of `d` (for daemon):
 
-[1] 2016/12/21 17:37:56.743377 [INF] Starting nats-streaming-server[test-cluster] version 0.3.6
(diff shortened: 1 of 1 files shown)


kozlovic:
@tianon Wondering if I am doing something wrong because it is the second time in a row that the PR is merged but `Run update.sh` is not started after the PR is merged, which leaves the docker up with new tag but old doc (see reference to v0.3.6 in [here](https://hub.docker.com/_/nats-streaming/)).

tianon:
@kozlovic arg, looks like our updater is having an issue: https://doi-janky.infosiftr.net/job/docs/job/update/

Thanks for pointing it out -- I'll go dig in and see if I can get it fixed now. :heart:

tianon:
Looks like I got it working again.  Did some updates yesterday on the Jenkins server which caused the issue. :disappointed:

kozlovic:
I saw it is updated, thanks!

PR #802, opened 2017-01-20
Add some notes about the caveats for "arbitrary user" support in postgres

tianon:
This is some initial documentation for the caveats noted in https://github.com/docker-library/postgres/pull/253.


tianon:
This will need some revisions (in a good way) if https://github.com/docker-library/postgres/pull/448 gets merged. :tada:

PR #804, opened 2017-01-23
Add documentation for the xwiki official image

vmassol:


vmassol:
@yosifkit Thanks for the review and the commits. I've just updated the doc too to align with the latest version from https://github.com/xwiki-contrib/docker-xwiki/blob/master/README.md

vmassol:
Apparently I made a mistake since the build is failing now :(

vmassol:
@yosifkit sorry I broke the build. I tried to fix it but I don't see the mistake I made. I guess I'll need your help once more... sorry

yosifkit:
No problem. The markdownfmt is very particular, but ensures specific formatting. :smiley_cat:  It looks like in this case it was needing an empty line between a paragraph and the list after it, needing each paragraph to take only one line, and needing to remove a trailing space on a line.

vmassol:
ok! Thanks for fixing it :)

xwiki/content.md
```diff
@@ -57,7 +55,7 @@ services:
       - MYSQL_USER=xwiki
       - MYSQL_PASSWORD=xwiki
   db:
-    image: "mysql:5"
+    image: "mysql:5.7"
```
vmassol: So `latest` needs to have all dependencies with a fixed versoin and it's only the main image software that doesn't have its version fixed?

xwiki/content.md
```diff
@@ -17,8 +17,6 @@ The goal is to provide a production-ready XWiki system running in Docker. This i
 
 # How to use this image
 
-[![](https://images.microbadger.com/badges/version/xwiki/xwiki-mysql-tomcat.svg)](https://microbadger.com/images/xwiki/xwiki-mysql-tomcat) [![](https://images.microbadger.com/badges/image/xwiki/xwiki-mysql-tomcat.svg)](https://microbadger.com/images/xwiki/xwiki-mysql-tomcat)
-
```
vmassol: ok I'll need to remember to remove those when in the future I'll sync the README from https://github.com/xwiki-contrib/docker-xwiki with the official image README! :)

PR #805, opened 2017-01-23
Added Convertigo images documentation.

nicolas-albert:
See https://github.com/docker-library/official-images/pull/2564.


nicolas-albert:
Hi, may can I help you for the review:

- associated with or contacted upstream?

I'm the upstream, see https://github.com/convertigo &
https://us.linkedin.com/in/nicolasalbert

- does it fit into one of the common categories?

"service"

- is it reasonably popular, or does it solve a particular use case well?

Yes it is reasonably popular, Convertigo MBaaS is public on source forge
since years, more than 10K developers are using Convertigo technologies
today. https://sourceforge.net/projects/convertigo/

- does a documentation PR exist? (should be reviewed and merged at roughly
the same time so that we don't have an empty image page on the Hub for very
long)

https://github.com/docker-library/docs/pull/805

- existing official images have been considered as a base? (ie, if foobar
needs Node.js, has FROM node:... instead of grabbing node via other means
been considered?)

This is well documented in the Dockerfiles. We us FROM 'openjdk' instead of
FROM 'tomcat' because tomcat 7.0.73 has an icomptibility with current
Convertigo Version, so we use tomcat 7.0.72 instead (inspired by the tomcat
Dockerfile). Futur releases (7.5.0) will be adapted for the latest tomcat
version and we will be able to use FROM =E2=80=98tomcat=E2=80=99

The web-connector/Dockerfile need a 32 bit JVM, so we get it from the
debian i386 repository (inspired by the openjdk Dockerfile).

- if FROM scratch, tarballs only exist in a single commit within the
associated history?

No.

- passes current tests? any simple new tests that might be appropriate to
add? (https://github.com/docker-library/official-images/tree/master/test)

No tests yet but we can create one and add it to the PR if needed.

PR #806, opened 2017-01-24
added Gradle

keeganwitt:


keeganwitt:
I don't know which has the more up-to-date code, but the requirements of the Travis job don't seem to match
```console
go get -u github.com/shurcooL/markdownfmt
markdownfmt -w <some-file>.md
```
Ideally you could run that command offline, and it'd satisfy the Travis job.

gradle/content.md
```diff
@@ -0,0 +1,22 @@
...
+
+`docker run --rm -v "$PWD":/project -w /project --name gradle gradle:alpine gradle <gradle-task>`
+
+## Reusing the Gradle cache
+
+The local Gradle cache can be reused across containers by creating a volume and mounting it in */root/.gradle*.
+
+```console
+docker volume create --name gradle-cache
+docker run -it -v gradle-cache:/root/.gradle gradle:latest gradle build
```
tianon: This is technically different now given your latest update to the image, right? (adding a user and a volume on that user's `.gradle` directory)

keeganwitt: Oops...Good catch.

PR #813, opened 2017-01-26
Fixed typo in README.md

bbrewington:
Fixed typo in docs/r-base/README.md


tianon:
`README.md` is generated from the contents of the other markdown files, so I've just fixed this typo here (and over in `crate` where it also appeared) directly in aab0824 :+1: :heart:

PR #814, opened 2017-01-27
Add --init flag to spiped examples

TimWolla:
Add Docker's `--init` flag to example invocations of spiped.

This removes the warning introduced by the patch at: https://github.com/TimWolla/docker-spiped/blob/master/1.5/Dockerfile#L14-L16

My intent is to remove the patch in the future and coordinate with upstream whether it should remain in the upcoming spiped 1.6 version or whether it should be completely removed.

### DIFF
--- a/spiped/content.md
+++ b/spiped/content.md
@@ -20,19 +20,19 @@ usage: spiped {-e | -d} -s <source socket> -t <target socket> -k <key file>
 For example running spiped to take encrypted connections on port 8025 and forward them to port 25 on localhost would look like this:
 
 ```console
-$ docker run -d -v /path/to/keyfile:/spiped/key:ro -p 8025:8025 spiped -d -s '[0.0.0.0]:8025' -t '[127.0.0.1]:25'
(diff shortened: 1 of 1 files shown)


tianon:
I'm a bit on the fence with this one.  On the one hand, `--init` is a really cool addition, but on the other it's only Docker 1.13+ and so will take some time to become ubiquitous.  Do you think upstream would be OK with keeping their pid1-signal-handler behavior for a bit longer?

For `spiped`, I'm probably less conflicted than I'd be for something less new -- it's still new enough (and somewhat obscure enough) that I'm guessing the userbase of this image is still smallish. :sweat_smile:

TimWolla:
> Do you think upstream would be OK with keeping their pid1-signal-handler behavior for a bit longer?

I would think so. The work around behaviour is explicitely activated if spiped runs as PID 1 and disabled otherwise. So „conforming“ platforms are not affected. Otherwise Colin would not have merged my PR I suppose.

> but on the other it's only Docker 1.13+ and so will take some time to become ubiquitous.

FWIW Docker Hub already shows this:

> Support for older versions (down to 1.6) is provided on a best-effort basis.

----

In the end I don't really care and hope for a Dockerfile instruction. It's up to you to merge this (or not). I run my spiped containers using --init already, so this does not affect me anyway :innocent:

tianon:
:+1:

PR #816, opened 2017-01-30
Updates Bonita BPM doc

JeremJR:
updates links and migration part
related to https://github.com/docker-library/official-images/pull/2598


tianon:
This looks like it's good to go as soon as @yosifkit is done giving docker-library/official-images#2598 one more set of eyes. :+1:

PR #817, opened 2017-01-31
Update NextCloud docs with non php-fpm info

tiangolo:
The NextCloud official docs state that by default it runs with PHP-fpm, but the default (`latest`) tag uses Apache.

Also, the docs state that it would run on port `9000`, while it will actually be running on port `80` (as it is run with Apache).

As it already has an internal Apache instance and that Apache is listening on port `80`, I added a paragraph specifying how to run it directly, without a reverse proxy, by binding to the port directly.

### DIFF
--- a/nextcloud/content.md
+++ b/nextcloud/content.md
@@ -8,17 +8,23 @@ A safe home for all your data. Access & share your files, calendars, contacts, m
 
 # How to use this image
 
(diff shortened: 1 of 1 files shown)


yosifkit:
Looks like that was changed with https://github.com/docker-library/official-images/pull/2560.  I'll let @pierreozoux decide how he wants to change the docs to accommodate this change.

tilosp:
The reworked readme (#929) has this already fixed.

PR #821, opened 2017-02-04
Add docs for ZNC image

DarthGandalf:
For https://github.com/docker-library/official-images/pull/2617

### DIFF
--- a/znc/content.md
+++ b/znc/content.md
@@ -29,3 +29,5 @@ often blocked by web browsers, and therefore is not recommended.
 
 If you use any external module, put the .cpp, .py or .pm file to
 `/znc-data/modules` (you may need to create that directory).
+
+IPv6 is currently disabled in this image.
--- a/znc/content.md
+++ b/znc/content.md
@@ -27,5 +27,5 @@ $ docker run -p 6697:6697 -v znc-cfg:/znc-data znc
 The port should match the port you used during `--makeconf`. Note that 6667 is
 often blocked by web browsers, and therefore is not recommended.
 
-If you use any external module, put the .cpp, .py or .pl file to
+If you use any external module, put the .cpp, .py or .pm file to
 `/znc-data/modules` (you may need to create that directory).
--- a/znc/content.md
+++ b/znc/content.md
@@ -26,3 +26,6 @@ $ docker run -p 6697:6697 -v znc-cfg:/znc-data znc
 
 The port should match the port you used during `--makeconf`. Note that 6667 is
 often blocked by web browsers, and therefore is not recommended.
+
+If you use any external module, put the .cpp, .py or .pl file to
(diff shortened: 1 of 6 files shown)


tianon:
Looks like this could use some `markdownfmt` to appease Travis -- the easiest way to apply it is via our `markdownfmt.sh` script in the root of this repo, but I'm also happy to apply the reformatting for you, if you'd like. :+1:

znc/content.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+To run ZNC:
+
+```console
+$ docker run -p 6697:6697 -v znc-cfg:/znc-data znc
+```
+
+The port should match the port you used during `--makeconf`. Note that 6667 is
+often blocked by web browsers, and therefore is not recommended.
```
tianon: Not sure what you mean here by "blocked by web browsers" -- maybe ISPs or firewalls?

znc/content.md
```diff
@@ -0,0 +1,33 @@
...
+
+If you use any external module, put the .cpp, .py or .pm file to
+`/znc-data/modules` (you may need to create that directory).
+
+IPv6 is currently disabled in this image.
```
tianon: Do you think it'd be worthwhile to expound a little bit about why here, similar to the notes in the `Dockerfile` comments?

znc/content.md
DarthGandalf: http://superuser.com/questions/188006/how-to-fix-err-unsafe-port-error-on-chrome-when-browsing-to-unsafe-ports
https://jazzy.id.au/2012/08/23/why_does_chrome_consider_some_ports_unsafe.html

tianon: Oh wow, that's adorable -- I had no idea.  Apologies! :+1:

DarthGandalf: Yeah, we had users complaining all the time that they can't connect, until ZNC itself started warning users during `--makeconf` that 6667 is not the best choice if they want to use a browser.

DarthGandalf:
Done.

tianon:
Tested locally since Travis is having webhook delays due to the GitHub/S3 cascade:
```console
$ curl -fsSL 'https://raw.githubusercontent.com/DarthGandalf/docker-docs/67f571b160c7080f02ec5489a16841dabc37e2f4/znc/content.md' | ./markdownfmt.sh -d
$ curl -fsSL 'https://raw.githubusercontent.com/DarthGandalf/docker-docs/67f571b160c7080f02ec5489a16841dabc37e2f4/znc/license.md' | ./markdownfmt.sh -d
$ curl -fsSL 'https://raw.githubusercontent.com/DarthGandalf/docker-docs/67f571b160c7080f02ec5489a16841dabc37e2f4/znc/variant-slim.md' | ./markdownfmt.sh -d
$ 
```

PR #831, opened 2017-02-10
[XWiki] Sync doc with https://github.com/xwiki-contrib/docker-xwiki

vmassol:


vmassol:
I think the issue is that the editor I use replaces tabs by spaces ;)

vmassol:
hmm I can't figure out what's making markdownfmt fail. Any idea? Thanks

tianon:
Looks like the line `Start by running a MySQL container ...` ends with a single space before the newline. :+1:

vmassol:
@tianon Thanks. How do you find what's wrong? When I look at the Travis logs I have a hard time understanding where the problem is.

tianon:
@vmassol the output is in the form of a "diff" from the current to what `markdownfmt` expects (which you can also generate locally via `./markdownfmt.sh -d xwiki/content.md`), so I found the space by selecting text and seeing that the line it's removing has an extra space at the end (since I couldn't see anything else that might be it)

If you're feeling lucky, you can let `markdownfmt` just apply the suggested changes directly by itself via `./markdownfmt.sh -w xwiki/content.md`, but I'd recommend reviewing the changes by hand when you do that (since it does foul things up a little sometimes).

PR #835, opened 2017-02-15
Remove warning for old docker setup

andrewelkins:
Using docker for Mac and Docker for Windows no longer uses virtualbox


yosifkit:
Many users are stuck with `docker-toolbox` and the `boot2docker` VirtualBox VM due not being able to meet the requirements of "Docker for Mac/Windows" (unable to upgrade OS, or required to use VirtualBox for other projects and thus unable to use Hyper-V/xhyve).  So, I'd rather keep the warning and instead reword it to not apply to "Docker for Mac/Windows" if we can get confirmation of users using bind mounts for MongoDB to their host OS on both platforms.  I think that there are still caveats when doing that since the shared file system was causing very high CPU usage for some users.

Also, any changes should be in `content.md`, since `README.md` is generated.

tianon:
This PR appears to be stale now (and I'm thus closing it).  If @yosifkit's comment is addressed, a new PR would be excellent.  Thanks! :+1: :heart:

PR #837, opened 2017-02-17
Add details about trust authentication.

ebrentnelson:
The PostgreSQL image sets up trust authentication and may be confusing for folks who are unfamiliar with it. Add a note to explain this.


tianon:
Sorry for the super-long delay here -- I rebased this and made a few minor tweaks (mostly just to make Travis happy with the formatting).  Thanks for the addition! :+1:

PR #838, opened 2017-02-18
Update README.md to show automatic daemon exit

pastcompute:
The `docker-entrypoint.sh` script always starts the postgresql daemon regardless. I wanted to script  the process of creating a local user as described, so rather than having to hit CTRL+C you can force an exit by passing the `--help` argument so this it self exist on completion.
I remember in passing seeing people asking this question in somewhere else so hopefuly this README update will help in the future.

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -99,6 +99,7 @@ The two easiest ways to get around this:
 	The files belonging to this database system will be owned by user "postgres".
 	...
 	( once it's finished initializing successfully and is waiting for connections, stop it )
+	( if you need the process to exit automatically, append `--help` to the command to avoid the daemon restarting )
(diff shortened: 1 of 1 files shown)


tianon:
Sorry for the long review delay. :disappointed:

I'm not against this, but explicitly documenting it _does_ add a bit of "permanence" (ie, "supported") to it that it doesn't really have right now (as a cute hack that happens to work due to how the entrypoint currently functions).

@yosifkit thoughts?

yosifkit:
Yeah, I don't think this is really a supported way; since it is possible that we improve this image to work similarly to the mysql image and pass all the user arguments along to the localhost-only postgres server.  And I could see us adding something similar to the [wantHelp](https://github.com/docker-library/mysql/blob/93e98b9454f0505194747ab37024a4c06aec6750/8.0/docker-entrypoint.sh#L11-L19) checking we have in mysql that skips all initialization if the user just wants the output of `help` or `version`.

PR #849, opened 2017-03-07
[XWiki] Add support for postgres

vmassol:


xwiki/content.md
```diff
@@ -106,17 +106,188 @@ services:
       - MYSQL_DATABASE=xwiki
     networks:
       - bridge
+# How to use this image
```
tianon: Uh, I think we've entered a time loop here -- this content looks very similar to the content above, but is now duplicated instead of replaced.  I don't think that was intentional. :smile:

vmassol:
Thanks @tianon for noticing the bad copy paste... apparently hitting del in github UI didn't remove the selected text for some reason... Should be ok now.

PR #866, opened 2017-03-29
centos remove MAINTAINER

corburn:
https://docs.docker.com/engine/deprecated/#maintainer-in-dockerfile


tianon:
Nice, LGTM! :heart:

FYI @jperrin

PR #874, opened 2017-04-09
Updated MySQL docs re: _FILE env vars

TFenby:
n/t

### DIFF
--- a/mysql/content.md
+++ b/mysql/content.md
@@ -120,6 +120,14 @@ This is an optional variable. Set to `yes` to generate a random initial password
 
 Sets root (*not* the user specified in `MYSQL_USER`!) user as expired once init is complete, forcing a password change on first login. *NOTE*: This feature is supported on MySQL 5.6+ only. Using this option on MySQL 5.5 will throw an appropriate error during initialization.
 
+## Docker Secrets
+
+As an alternative to passing sensitive information via environment variables, \_FILE may be appended to the previously listed environment variables, causing the initialization script to load the values for those variables from files present in the container. In particular, this can be used to load passwords from Docker secrets stored in /run/secrets/<secret_name> files. For example:
+
+```console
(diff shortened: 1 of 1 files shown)


mysql/content.md
```diff
@@ -120,6 +120,14 @@ This is an optional variable. Set to `yes` to generate a random initial password
 
 Sets root (*not* the user specified in `MYSQL_USER`!) user as expired once init is complete, forcing a password change on first login. *NOTE*: This feature is supported on MySQL 5.6+ only. Using this option on MySQL 5.5 will throw an appropriate error during initialization.
 
+## Docker Secrets
+
+As an alternative to passing sensitive information via environment variables, \_FILE may be appended to the previously listed environment variables, causing the initialization script to load the values for those variables from files present in the container. In particular, this can be used to load passwords from Docker secrets stored in /run/secrets/<secret_name> files. For example:
```
yosifkit: We'll probably want to point out which variables `_FILE` does or does not work on.  Currently it applies to `MYSQL_ROOT_PASSWORD`, `MYSQL_ROOT_HOST`, `MYSQL_DATABASE`, `MYSQL_USER`, and `MYSQL_PASSWORD`, but not to `MYSQL_ALLOW_EMPTY_PASSWORD`, `MYSQL_RANDOM_ROOT_PASSWORD`, `MYSQL_INITDB_SKIP_TZINFO`, and `MYSQL_ONETIME_PASSWORD`.

So far the dividing line seems to be variables that have a value, and are not just an on/off switch.  I'm not sure that this will always be the case; perhaps we just need to add a sentence to each one that supports `_FILE` to specifically call it out.

TFenby: Good point. I'll add a note about that.

TFenby:
poking @tianon

tianon:
Thanks!

PR #881, opened 2017-04-13
Publish port 80

prasadkatti:
Publish container port 80 to the host

### DIFF
--- a/httpd/content.md
+++ b/httpd/content.md
@@ -21,7 +21,7 @@ Then, run the commands to build and run the Docker image:
 
 ```console
 $ docker build -t my-apache2 .
-$ docker run -dit --name my-running-app my-apache2
+$ docker run -dit -p 80:80 --name my-running-app my-apache2
 ```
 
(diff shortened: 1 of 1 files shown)


yosifkit:
For the images we maintain we don't usually put a `-p` since it is not required to access the container and when used on a Linux host exposes the the container's port to everyone on the network.

prasadkatti:
Ohh okay. I was following the instructions from 'How to use this image' section, and trying to access the website from the host. After adding -p option, it started working; so I made a PR. But I am only starting with docker, so I guess I will read up a bit more and understand what other ways are there to connect to the website running in the container.

Thanks for your time. I will go ahead and close this.

yosifkit:
If you are running on a Linux host you can hit the IP address of the container directly.  To get the IP you can use inspect: `docker inspect --format '{{ .NetworkSettings.IPAddress }}' container-name`; then you can visit `http://ip-address` in your browser.

prasadkatti:
Great! Thank you.

PR #884, opened 2017-04-18
comment out note about reusing cache for now

keeganwitt:
to avoid confusion


tianon:
Sorry for not seeing this sooner :disappointed:

Docker Hub doesn't support HTML comments. :disappointed: (they render as-is)

keeganwitt:
Bummer.  I amended the commit to simply remove it for now.

PR #885, opened 2017-04-20
Update README.md

urfinjuezz:
fix error in docker-compose example

### DIFF
--- a/wordpress/README.md
+++ b/wordpress/README.md
@@ -99,7 +99,7 @@ services:
       - web
 networks:
    web:
-     driver= bridge
+     driver: bridge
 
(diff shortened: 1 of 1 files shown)


tianon:
Can you elaborate on why you're making this change?  I can't reproduce any kind of failure with the existing example file. :confused:

urfinjuezz:
https://docs.docker.com/compose/environment-variables/#setting-environment-variables-in-containers

With colon i have error "could not determine a constructor for the tag"

yosifkit:
That sounds like a whitespace issue.  For environment variables it can be either a list of strings that are each in the form of `key=value` (or the special case of `key`) or a map of keys to values:

```yaml
    environment:
      WORDPRESS_DB_PASSWORD: example
    environment:
      - WORDPRESS_DB_PASSWORD=example
```
￼

PR #886, opened 2017-04-21
Explain who maintains this image

mlhess:
We either need to explain that this is maintained by the Docker community or pull the image from docker.  We are getting push back on our side about this.

### DIFF
--- a/drupal/content.md
+++ b/drupal/content.md
@@ -8,7 +8,7 @@ Drupal is a free and open-source content-management framework written in PHP and
 
 # Who maintains this
 
-This image is not maintained by the Drupal community or the Drupal security team.  This image is maintained by the docker community. 
+This image is not maintained by the Drupal community or the Drupal Security Team. This image is maintained by the docker community.
 
 # How to use this image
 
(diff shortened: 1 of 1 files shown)


CashWilliams:
+1

tianon:
Sorry for any trouble we've caused, especially related to the unfortunate name of the "official images" program :disappointed:

I've been working on refactoring all the descriptions to make it clear who maintains each image (Docker Community vs upstream), so I'll merge this in the interim and hopefully make all the images more clear about this in the near future. :+1:

I'd also like to be clear that we'd be happy to turn over maintenance of the image if you'd like, or deprecate it entirely if you'd prefer (although we'd like to avoid that if possible :innocent:).

Thanks for the PR :heart:

sylus:
We can explain who maintains this image overall but there already are quite a few active Drupal community members here already submitting P.R.s. So there is Drupal community involvement. 

Additionally we get the benefit of the experienced Docker team, a few times now they were one of the first one with recent MySQL, Postgres, or PHP security issues. From a Drupal security point of review, what is needed? I think the Docker folks are usually within 24-48 having a docker image updated when there is a Drupal release. Heck they even went back and support 8.2.8 when the 8.2.x line was apparently done.

I just want to make sure this isn't the first step in Drupal wanting to maintain its own Docker images which I will heavily argue against.

PR #898, opened 2017-05-01
Fix TimWolla's last name

tianon:
Sorry again, @TimWolla :disappointed:

I've updated both `adminer` and `spiped` -- wanna take a look and make sure I got it right this time? :pray:


TimWolla:
LGTM

PR #900, opened 2017-05-06
Add notes about CPU and RAM limits

akashche:
Relevant links:

https://developers.redhat.com/blog/2017/04/04/openjdk-and-containers/
https://bugs.openjdk.java.net/browse/JDK-6515172
https://bugs.openjdk.java.net/browse/JDK-8170888
https://bugs.openjdk.java.net/browse/JDK-8157478
https://github.com/moby/moby/issues/27892
https://github.com/moby/moby/issues/30594
https://superuser.com/a/908870
https://technet.microsoft.com/en-us/library/cc770297%28v=ws.11%29.aspx

### DIFF
--- a/openjdk/content.md
+++ b/openjdk/content.md
@@ -41,7 +41,7 @@ This will add your current directory as a volume to the container, set the worki
 
 ## Make JVM respect CPU and RAM limits
 
-On starup JVM tries to detect the number of available CPU cores and the amount of RAM to adjust its internal parameters (like the number of garbage collector threads to spawn) accordingly. When container is run with limited CPU/RAM, standard system API, used by JVM for probing, will return host-wide values. This can cause excessive CPU usage and memory allocation errors with older versions of JVM.
+On startup JVM tries to detect the number of available CPU cores and the amount of RAM to adjust its internal parameters (like the number of garbage collector threads to spawn) accordingly. When container is run with limited CPU/RAM, standard system API, used by JVM for probing, will return host-wide values. This can cause excessive CPU usage and memory allocation errors with older versions of JVM.
 
 Inside Linux containers, recent versions of OpenJDK 8 can correctly detect container-limited number of CPU cores by default. To enable the detection of container-limited amount of RAM the following options can be used:
 
--- a/openjdk/content.md
+++ b/openjdk/content.md
@@ -38,3 +38,23 @@ $ docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%REPO%%:7 javac Ma
 ```
 
 This will add your current directory as a volume to the container, set the working directory to the volume, and run the command `javac Main.java` which will tell Java to compile the code in `Main.java` and output the Java class file to `Main.class`.
+
+## Make JVM respect CPU and RAM limits
+
+On starup JVM tries to detect the number of available CPU cores and the amount of RAM to adjust its internal parameters (like the number of garbage collector threads to spawn) accordingly. When container is run with limited CPU/RAM, standard system API, used by JVM for probing, will return host-wide values. This can cause excessive CPU usage and memory allocation errors with older versions of JVM.
+
+Inside Linux containers, recent versions of OpenJDK 8 can correctly detect container-limited number of CPU cores by default. To enable the detection of container-limited amount of RAM the following options can be used:
+
+```console
+$ java -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap ...
+```
+
+Inside Windows Server (non-Hyper-V) containers, limit for number of available CPU cores does not work (is ignored by Host Compute Service). To set such limit manually, JVM can be started the following way:
+
+```console
+$ start /b /wait /affinity 0x3 path/to/java.exe ...
+```
+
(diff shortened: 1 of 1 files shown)


openjdk/content.md
```diff
@@ -38,3 +38,23 @@ $ docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%REPO%%:7 javac Ma
 ```
 
 This will add your current directory as a volume to the container, set the working directory to the volume, and run the command `javac Main.java` which will tell Java to compile the code in `Main.java` and output the Java class file to `Main.class`.
+
+## Make JVM respect CPU and RAM limits
+
+On starup JVM tries to detect the number of available CPU cores and the amount of RAM to adjust its internal parameters (like the number of garbage collector threads to spawn) accordingly. When container is run with limited CPU/RAM, standard system API, used by JVM for probing, will return host-wide values. This can cause excessive CPU usage and memory allocation errors with older versions of JVM.
```
TimWolla: `s/starup/startup`.

akashche:
@TimWolla , thanks for the correction! Fixed.

openjdk/content.md
```diff
@@ -38,3 +38,23 @@ $ docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%REPO%%:7 javac Ma
 ```
 
 This will add your current directory as a volume to the container, set the working directory to the volume, and run the command `javac Main.java` which will tell Java to compile the code in `Main.java` and output the Java class file to `Main.class`.
+
+## Make JVM respect CPU and RAM limits
+
+On startup JVM tries to detect the number of available CPU cores and the amount of RAM to adjust its internal parameters (like the number of garbage collector threads to spawn) accordingly. When container is run with limited CPU/RAM, standard system API, used by JVM for probing, will return host-wide values. This can cause excessive CPU usage and memory allocation errors with older versions of JVM.
+
+Inside Linux containers, recent versions of OpenJDK 8 can correctly detect container-limited number of CPU cores by default. To enable the detection of container-limited amount of RAM the following options can be used:
```
tianon: Does the version we have actually include this new functionality? (I thought we were still a release or two behind due to Debian's backport process? 8u121 in `jessie-backports` vs 8u131 in `stretch`)

openjdk/content.md
```diff
@@ -38,3 +38,23 @@ $ docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%REPO%%:7 javac Ma
...
+$ java -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap ...
+```
+
+Inside Windows Server (non-Hyper-V) containers, limit for number of available CPU cores does not work (is ignored by Host Compute Service). To set such limit manually, JVM can be started the following way:
+
+```console
+$ start /b /wait /affinity 0x3 path/to/java.exe ...
+```
+
+In this example CPU affinity hex mask `0x3` will limit JVM to 2 CPU cores.
```
tianon: Does this limit to 2 CPU cores generally, or does this limit to 2 _specific_ cores? (ie, `docker run --cpuset-cpus` vs `docker run --cpus`)

openjdk/content.md
```diff
@@ -38,3 +38,23 @@ $ docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%REPO%%:7 javac Ma
...
+
+RAM limit is supported by Windows Server containers, but currently JVM cannot detect it. To prevent excessive memory allocations, `-XX:MaxRAM=...` option must be specified with the value that is not bigger than a containers RAM limit.
```
tianon: Ooh, nice -- I had no idea there was a JVM parameter for controlling `MaxRAM` rather than trying to set `-Xmx` and friends intelligently! (related to https://github.com/docker-library/openjdk/issues/57 and https://github.com/docker-library/openjdk/pull/71)

openjdk/content.md
carlossg: not yet, openjdk:8-jdk is at 1.8.0_121

openjdk/content.md
akashche: Mask `0x3 - 0011` limits CPU to 2 specific cores - core0 and core1, its a windows feature, not jdk-specific.

openjdk/content.md
akashche: Yep, `MaxRAM` may be more convenient. This one-liner can be used to check what `Xmx` (max heap) value will be used when `MaxRAM` is specified:

    jrunscript -J-XX:MaxRAM=256M -e "print(Packages.java.lang.Runtime.getRuntime().maxMemory()/(1<<20) + 'M')"

openjdk/content.md
rcoup: OpenJDK 9 too? ([>=b150 afaict](https://bugs.openjdk.java.net/browse/JDK-8170888))

akashche: Hi, it looks like all docker-library jdks 8 are now at least on 8u131 that has cgroups flag, is it possible to have this doc change merged?

tianon:
Thanks @akashche!

workmanw:
@akashche Thank you for this!!!

PR #905, opened 2017-05-11
Initial basic multiarch support

tianon:
This isn't _full_ multi-architecture support, but it does allow us to generate better docs for `s390x/python`, for example.

Here's the diff from `./update.sh python` to `./update.sh s390x/python`:
https://gist.github.com/anonymous/8d3fc86cbe0ee91d1e242a1e661a3e77

(in a Gist since it contains Markdown and so GitHub's rendering gets kind of frustrated if I put it inline)

No updates to `push.pl` or `push.sh` here since they're already in use for our existing hacky docs pushing


tianon:
See https://github.com/docker-library/docs/blob/b1d25105844d14c2dedc8986da514c20a6c06d1f/push.pl#L111-L117 for the precise `push.pl` code which implements the same logic I've added to `update.sh`. :sunglasses:

busybox/content.md
```diff
@@ -32,8 +32,8 @@ This `Dockerfile` will allow you to create a minimal image for your statically c
 
 This image contains BusyBox built against various "libc" variants (for a comparison of "libc" variants, [Eta Labs has a very nice chart](http://www.etalabs.net/compare_libcs.html) which lists many similarities and differences).
 
--	`busybox:glibc`: [glibc from Debian](https://packages.debian.org/jessie/libc6) (which is then included in the image)
--	`busybox:musl`: [musl from Alpine](https://pkgs.alpinelinux.org/package/main/x86_64/musl) (statically compiled)
--	`busybox:uclibc`: [uClibc](https://uclibc.org) via [Buildroot](https://buildroot.org) (statically compiled)
+-	`%%IMAGE%%:glibc`: [glibc from Debian](https://packages.debian.org/jessie/libc6) (which is then included in the image)
+-	`%%IMAGE%%:musl`: [musl from Alpine](https://pkgs.alpinelinux.org/package/main/x86_64/musl) (statically compiled)
```
yosifkit: I noticed the `x86_64` and then found that the link fails. :cry:  Maybe `https://pkgs.alpinelinux.org/packages?name=musl` instead?

tianon: Nice, updated!

PR #913, opened 2017-05-19
Convert "wordpress" to use "docker stack deploy" as an initial "stack" foray

### LINKED ISSUE #912: Adding "Try in PWD" buttons
I was thinking about adding "Try in PWD" buttons (https://medium.com/@marcosnils/showcasing-your-applications-with-docker-in-a-single-click-9e9fa2908c51) to some of the official images so it's easier to showcase them. 

Here's an example button for the redis image:

[Try in PWD](http://play-with-docker.com?stack=/redis/latest)

Is this something that I need to talk with the image owners?. How would you recommend to proceed?

tianon:
Closes #912

Here's an example of what this generates: (and that "Try in PWD" link should actually work!)

Example `stack.yml` for `wordpress`:

```yaml
version: '3.1'

services:

  wordpress:
    image: wordpress
    ports:
      - 8080:80
    environment:
      WORDPRESS_DB_PASSWORD: example

  mysql:
    image: mysql:5.7
    environment:
      MYSQL_ROOT_PASSWORD: example
```

[Try in PWD](http://play-with-docker.com?stack=https://raw.githubusercontent.com/docker-library/docs/0cb43ce8ad1da073bbc30c6245cdffb433ba51ba/wordpress/stack.yml)

Run `docker stack deploy -c stack.yml wordpress` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080` or `http://host-ip:8080` (as appropriate).

### DIFF
--- a/.template-helpers/stack.md
+++ b/.template-helpers/stack.md
@@ -1,0 +1,7 @@
+... via [`docker stack deploy`](https://docs.docker.com/engine/reference/commandline/stack_deploy/) or [`docker-compose`](https://github.com/docker/compose)
+
+Example `stack.yml` for `%%REPO%%`:
+
+%%STACK-YML%%
+
+[![Try in PWD](https://github.com/play-with-docker/stacks/raw/cff22438cb4195ace27f9b15784bbb497047afa7/assets/images/button.png)](http://play-with-docker.com?stack=%%STACK-URL%%)
--- a/update.sh
+++ b/update.sh
@@ -59,6 +59,17 @@ for image in "${images[@]}"; do
 			fi
 		fi
 
+		stack=
+		stackYml=
+		stackUrl=
+		if [ -f "$repo/stack.yml" ]; then
+			stack="$(cat "$repo/stack.md" 2>/dev/null || cat "$helperDir/stack.md")"
+			stackYml=$'```yaml\n'"$(cat "$repo/stack.yml")"$'\n```'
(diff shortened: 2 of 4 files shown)


marcosnils:
LGTM!.

marcosnils:
Thx guys, I'll send a PR to add the currents stacks from the PWD stacks repo here. How does that sound?

tianon:
As long as they don't refer to any non-official images (https://github.com/docker-library/official-images/blob/b16d43ab480ef45ca309ce3dec192e12e94344ac/README.md#repeatability), that sounds like an OK place to start!

We'll need to update `REPO/content.md` for each too, so starting with the ones that already have a `docker-compose.yml` might be an easier bit of low-hanging fruit. :smile:

marcosnils:
@tianon got it. Sad that non-official images can't be used, as for tools like mysql / redis / mongo / etc I found super useful to add a "GUI" container so people can play with it easier. But I understand what you mean and it totally makes sense. 

On a different subject, do you also maintain the docker store readme files?. Is there a way we could reach there as well?.

yosifkit:
We don't directly control them but they should eventually get the changes we push to the Docker Hub.

tianon:
For `mysql` (and derivatives), we've got [`adminer`](https://hub.docker.com/_/adminer/), which works decently well in my experience (and is super lightweight).

marcosnils:
@tianon thx for the `adminer` tip. I'll try to change our stacks to use instead.

PR #917, opened 2017-05-29
Alter RabbitMQ image description

michaelklishin:
Team RabbitMQ feels this description is more to the point
(and does not use blanket marketing statements).


tianon:
Very nice :+1: :heart: 

LGTM!

PR #918, opened 2017-05-30
Add documentation to make s3server an official repository

LaureVergeron:


yosifkit:
This appears to be an unfinished new image proposal. Please see https://github.com/docker-library/official-images if you'd like to add an official image.

And sorry for the massive delay. :bowing_man:

PR #919, opened 2017-06-01
Add infinit.

Dimrok:
Add infinit documentation to the official images documentation (see https://github.com/docker-library/official-images/pull/3005)


infinit/content.md
```diff
@@ -0,0 +1,67 @@
...
+
+#### Accessing storage logics
+
+For now, the only logic available is a distributed file system.
+
+##### File system
+
+On UNIX-based systems, Infinit uses `FUSE` to provide mount points. Hence, for UNIX-based containers, you need `FUSE` installed on your host (see Infinit's [get-started page](https://infinit.sh/get-started#installation)) and run Docker in privileged mode (`--privileged`).
+
+	docker run --privileged -e INFINIT_USER=$USER -v ~/.local/share/infinit:/root/.local/share/infinit --rm infinit # volume mount --name <VOLUME> --mountpoint <DIRECTORY>
```
tianon: For using FUSE inside a container, I think `--privileged` is a bit overkill (since it essentially allows _all_ privileges of the host, instead of the strictly-necessary subset).  Can we trim this down to just the subset necessary?  I'd imagine something like `--cap-add SYS_ADMIN --device /dev/fuse` is probably pretty close to the actual set required.

infinit/content.md
```diff
@@ -0,0 +1,67 @@
...
+
+*NOTE: On Windows systems, Infinit uses `Dokan`, but Windows-native containers aren't officially supported yet by Infinit.*
+
+More file system interfaces will be added in the future to acces file storage logics.
+
+##### Others (object, block, etc.)
+
+At current version (0,7.3), the single logic exposed is the distributed file system but more logics will be added in the upcoming releases, such as an object storage logic, a block device logic and various interfaces for them.
```
tianon: Minor typo: `0,7.3` vs `0.7.3`

infinit/github-repo
```diff
@@ -0,0 +1 @@
+http://github.com/infinit/infinit-docker
```
tianon: Can you swap this to use `https` please? (it'll redirect, but better to be explicit IMO)

infinit/license.md
```diff
@@ -0,0 +1 @@
+View [license information](https://www.infinit.sh/licenses/infinit.txt) for the software contained in this image.
```
tianon: This URL appears to be a 404. :confused:

image

infinit/content.md
```diff
@@ -0,0 +1,67 @@
+Infinit’s storage platform transforms commodity servers into multiple flexible, scalable, secure and fault-tolerant storage logics tailored to your applications.
```
tianon: You probably want to include an `h1` header above this, something like `# Infinit` or `# What is Infinit?` (similar to other images) or else this block will combine with the block in the template above it.  If you run `./update.sh infinit` it should generate `infinit/README.md` which you can then use to get an idea for how your generated content is going to flow (just don't commit that generated file -- we've got a bot that'll update it automatically after merge).

Also, that `logo.png` that's included here will only show up if `%%LOGO%%` is used somewhere here in the body.  In other descriptions, we usually put it right after this initial "what is XYZ?" paragraph, but some maintainers prefer to list it either earlier or later -- it's really up to you. :+1:

infinit/license.md
Dimrok: Waiting for the website to be deployed.

Dimrok: Fixed.

infinit/content.md
```diff
@@ -0,0 +1,45 @@
...
+
+Infinit stores its configuration under `/root/.local/share/infinit`. If you want to share this folder with your host, you can mount your host's configuration folder as a volume in your container.
+
+	docker run -v ~/.local/share/infinit:/root/.local/share/infinit --rm infinit
+
+*N.B. This might affect permissions of the mounted folder.*
+
+### Using your user
+
+By default, Infinit uses the current user, `root` when running on a container. If you want to use your local user name, you can set the INFINIT_USER environment variable.
```
yosifkit: This line is a bit confusing.  Does infinit use "the current user" or is it "`root`" "when running on a container"?

yosifkit: What does the `INFINIT_USER` environment variable do?  If I pass in my linux username, it is not going to mean anything inside the container since it does not have my `/etc/passwd` to get a user id.

infinit/content.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Accessing storage logics
+
+For now, the only logic available is a distributed file system.
+
+#### File system
+
+On UNIX-based systems, Infinit uses `FUSE` to provide mount points. Hence, for UNIX-based containers, you need `FUSE` installed on your host (see Infinit's [get-started page](https://infinit.sh/get-started#installation)) and run Docker with specific options (`--cap-add SYS_ADMIN --device /dev/fuse`).
+
+	docker run --cap-add SYS_ADMIN --device /dev/fuse -e INFINIT_USER=$USER -v ~/.local/share/infinit:/root/.local/share/infinit --rm infinit volume mount --name <VOLUME> --mountpoint <DIRECTORY>
```
yosifkit: It looks like this `docker run` would use a share that was already created (`--name`) and mount it at `--mountpoint`; would this be available only for that container or is the mount point relative to the host mount namespace?

How would someone relate this to using `infinit` as a volume plugin: https://infinit.sh/documentation/docker/volume-plugin?  I anticipate that this is what most people will want to use the image for, so it would be nice if there was a clear step-by-step process to go from "nothing" to "my swarm has access to share volumes via infinit".

yosifkit: Maybe it would need to mention where the silos are stored (`/var/storage/infinit/`) so that the inifinit plugin container will use the same local directory (`-v /path/to/storage/:/var/storage/infinit/`)?

Dimrok: > It looks like this docker run would use a share that was already created (--name) and mount it at --mountpoint; would this be available only for that container or is the mount point relative to the host mount namespace?

The example uses a configuration folder (`<blabl>/share/infinit`) shared between the host and the container but the mountpoint will be available only for the container, but everybody mounting the same volume anywhere else will have access and be able to share data. 

> [...] "my swarm has access to share volumes via infinit".

About swarm, because some options (`privileged`, `add-cap`, etc.) available for `docker run` are not available for `docker service`, this is not possible to have fuse mount points in a swarm. 

However, I guess I can write something about the `infinit deamon` in a container used by other containers.

yosifkit: I guess what I meant was making it a volume plugin on Docker and did not realize that plugins didn't work on swarm... until this next release!

So, it looks like volume plugins will work on swarm:

- PR: https://github.com/moby/moby/pull/33575
- merge commit https://github.com/moby/moby/commit/05c7c311390911daebcf5d9519dee813fc02a887
- commit in docker-ce: https://github.com/docker/docker-ce/commit/0887a18a3a2ea38cc4bf89e8f8e333be7288fe86

So, 17.07.0-ce-rc1 and above will work for running this as a volume plugin on a swarm. And I think an example or link to getting it to work as a plugin on a single docker would help users immensely.

Thanks!

Dimrok:
Great!

I'll give it a try and write a comprehensive plugin integration example.

tianon:
Any update on the comments here? :pray: :heart:

yosifkit:
Closing since image PR is closed: https://github.com/docker-library/official-images/pull/3005

PR #920, opened 2017-06-01
Adding docs for official SWI-Prolog image.

ninjarobot:
Publishing documentation ahead of forthcoming image.


tianon:
Looks like you've got two spaces after that period and the CI wants just one.

swipl/content.md
```diff
@@ -0,0 +1,29 @@
...
+docker run -it %%IMAGE%% 
+```
+
+## Build an image with your application
+
+It is also useful to build images on top of this base image which rely on SWI-Prolog and provide their own default command:
+
+```dockerfile
+FROM %%IMAGE%%
+ADD souce.tar.gz /app
```
tianon: This is a bit of a traditionally-odd way to deploy code in a `Dockerfile` -- is it common for Prolog sources to be developed-on/deployed via `.tar.gz` files?  Wouldn't it be more common to put a `Dockerfile` at the root of a Prolog application development tree, and thus use something like `COPY . /app` ?

ninjarobot: It's not particularly common to have them in .tar.gz files.  It's been a while, but I believe there were other similar docs somewhere.  I can adjust, as `COPY . /app` would work just as well.

PR #930, opened 2017-06-08
Updates Consul logo to latest branding.

slackpad:


slackpad:
Will resubmit with an .svg since that is supported.

PR #941, opened 2017-06-18
Migrate current compose examples to PWD capable stacks

marcosnils:
Added stack files for the current official images:

- joomla
- drupal
- geonetwork
- owncloud


tianon:
cc image maintainers @TimWolla @doublebyte1 @mbabker @brianteeman -- extra eyes on your relevant parts (or the whole thing!) of this would be A+ appreciated :+1: :heart:

tianon:
Just to thread the needle, this relates to the discussion in https://github.com/docker-library/docs/issues/912 and the implementations in https://github.com/docker-library/docs/pull/913 and https://github.com/docker-library/docs/pull/933.

mbabker:
Looks fine to me.

adminer/content.md
```diff
@@ -30,6 +30,10 @@ Then point your web server to port 9000 of the container.
 
 Note: This exposes the FastCGI socket to the Internet. Make sure to add proper firewall rules or use a private Docker network instead to prevent a direct access.
 
+## %%STACK%%
+
+Run `docker stack deploy -c stack.yml %%REPO%%` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
```
TimWolla: I'm not really sure whether this really provides a value-add for Adminer: Usually you would use an existing database, instead of creating a new one to use.

The online demo using this `stack.yml` works fine, from what I've seen though.

brianteeman:
Thanks. Looks good

adminer/content.md
marcosnils: > I'm not really sure whether this really provides a value-add for Adminer: Usually you would use an existing database, instead of creating a new one to use.

@TimWolla thx for the feedback. What about people that people that just want to try it the first time to see that it'll suite their needs?. Another use-case is whenever the image gets updated you can just launch the new version with a single click to try new features.

TimWolla: @marcosnils Is this paragraph required to the online demo?

In any case, I don't *object* this change, but this paragraph refers to a downloaded swarm.yml which IMO is unnecessary clutter for adminer.

adminer/content.md
```diff
@@ -30,6 +30,10 @@ Then point your web server to port 9000 of the container.
 
 Note: This exposes the FastCGI socket to the Internet. Make sure to add proper firewall rules or use a private Docker network instead to prevent a direct access.
 
+## %%STACK%%
```
TimWolla: This headline is either at the wrong position or of the wrong size.

marcosnils: Fixed!.

doublebyte1:
@tianon @marcosnils This looks fine to me, Thanks!

PR #842, opened 2017-02-22
Deprecating elasticsearch, kibana, and logstash

yosifkit:
This is the result of the discussion in https://github.com/docker-library/elasticsearch/pull/160.

cc @jasontedor, I would appreciate feedback to ensure the wording is acceptable.  I linked to elastic.co docs at version `5.2` instead of `current` to ensure that any future docs changes won't break the links.

This will appear at the top of the Docker Hub page, similar to [rails](https://hub.docker.com/_/rails/).

PR #945, opened 2017-06-20
Update deprecation notice

dliappis:
Version 5.5 of the Elastic stack, including Elasticsearch, Kibana and Logstash has been delayed.

Update documentation clarifying that images will keep receiving updates until version 5.6.0.


dliappis:
cc @yosifkit who, so kindly, worked on the related #842

tianon:
I've been looking at some of the Hub comments for these images since we added the deprecation notice, and wondering if we should maybe update this message to clarify that the reason the move is happening because this new image _is_ maintained by  Elastic, where the old image is not.  What're your thoughts?  Think that's worth clarifying further?

Otherwise this LGTM.

Once 5.6 is released, 5.5 is EOL, right? (wanting to make sure my previous understanding is correct and that I'm parsing this update properly :sweat_smile:)

tianon:
Probably also worth CCing @jasontedor :smile:

elasticdog:
A slightly reworded version based on @tianon's suggestion:

> This image has been deprecated in favor of the [official `elasticsearch` image](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html) provided and maintained by elastic.co. The upstream images are available to pull via `docker.elastic.co/elasticsearch/elasticsearch:[version]` like `5.4.1`. The images found here will receive no further updates once the `5.6.0` release is available upstream. Please adjust your usage accordingly.

dliappis:
I am happy with @elasticdog's [rewording above](https://github.com/docker-library/docs/pull/945#issuecomment-309808195); @tianon what do you think? If it's LGTM grade :) I can update my PR.

tianon:
Sounds fine to me!  Don't have strong opinions here, just figured I'd point out what users are saying in case it helps get something that makes it easier for them to understand why this is happening. :+1: :heart:

jasontedor:
I'm good with it too. Thanks for the iteration @tianon!

dliappis:
I pushed @elasticdog's [reworded deprecated notice](https://github.com/docker-library/docs/pull/945#issuecomment-309808195).

dliappis:
... and many thanks @tianon for reviewing and all the :heart:  and understanding!

PR #947, opened 2017-06-22
update to erlang 20

getong:


c0b:
looks good to reflect to https://hub.docker.com/_/erlang/

PR #948, opened 2017-06-22
deprecate official docker image

ndeloof:
see https://github.com/jenkinsci/docker/issues/515

### DIFF
--- a/jenkins/deprecated.md
+++ b/jenkins/deprecated.md
@@ -1,0 +1,2 @@
+This image has been deprecated in favor of the [`jenkins/jenkins:lts`](https://hub.docker.com/r/jenkins/jenkins) image provided and maintained by [Jenkins Community](https://jenkins.io/) as part of project's release process. 
+The images found here will receive no further updates. Please adjust your usage accordingly.
--- a/jenkins/deprecated.md
+++ b/jenkins/deprecated.md
@@ -1,2 +1,4 @@
 This image has been deprecated in favor of the [`jenkins/jenkins:lts`](https://hub.docker.com/r/jenkins/jenkins) image provided and maintained by [Jenkins Community](https://jenkins.io/) as part of project's release process. 
-The images found here will receive no further updates. Please adjust your usage accordingly.
+The images found here will receive no further updates after LTS 2.60.x. Please adjust your usage accordingly.
+
+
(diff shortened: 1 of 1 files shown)


tianon:
Totally understand where you're coming from; sad to see this one go! :heart:

My only ask would be that we have a window of deprecation -- something like a month or so where the image is deprecated but still receiving updates (to give users a chance to migrate off and/or even notice the deprecation notice).  Sound reasonable?

I'm happy to adjust/fixup the formatting myself to satisfy Travis, too. :+1:

ndeloof:
sounds perfectly reasonable, we can get the current LTS bits deployed here and wait for next LTS base to be selected for new image location.

tianon:
Does this sound accurate?

> This image has been deprecated in favor of the [`jenkins/jenkins:lts`](https://hub.docker.com/r/jenkins/jenkins) image provided and maintained by [Jenkins Community](https://jenkins.io/) as part of project's release process. The images found here will receive no further updates following the release of Jenkins LTS 2.74 (the next Jenkins LTS release). Please adjust your usage accordingly.

(Is there a different wording that would perhaps be more accurate?  Did I do my LTS version number math correctly or is it more fluid than that? :innocent:  We can re-word to make it based on "the next LTS following 2.60" or something, if that makes sense, or simply make it a time-based deprecation. :+1:)

Also, it doesn't look like anything is pushed to `jenkins/jenkins:lts` yet -- will you be updating in both places, or only start pushing over there once this image isn't receiving updates? (because the description as we're currently wording it implies users ought to start using that other version right away, and we're only maintaining this version to help patch them over and give them time to notice the switch :smile:)

ndeloof:
sounds good.
new LTS in 2.60.1. We could keep the `jenkins` docker image updated for future 2.60.x LTS releases, and fully switch to the `jenkins/jenkins` repo for the next one

We will soon push releases to `jenkins/jenkins`, we have discussions with jenkins-infra team to get this setup. We can keep this PR pending in the meantime.

ndeloof:
@tianon during jenkins governance meeting yesterday we agreed to deprecate "jenkins" official image and move to jenkins/jenkins. We will propose LTS updates as long as the 2.60.x branch exists (2.60.2 is coming soon).

ndeloof:
2.60.2 actually is already pushed to https://hub.docker.com/r/jenkins/jenkins.
also see https://github.com/docker-library/official-images/pull/3218

tianon:
Nice, thanks :heart:

PR #955, opened 2017-07-03
Add a quote blurb from the upstream docs to the scratch description

tianon:
There are _several_ comments on the Hub noting this section of the docs, so I figured we should add it to the description directly instead.


yosifkit:
Only has README.md, ignoring failure of "readme has changed".

PR #956, opened 2017-07-03
Update release referenced in the prose as latest

hobti01:
"the time of this writing" should always be current.


hobti01:
Wrong place for doc change

PR #960, opened 2017-07-07
Update the doc for Silverpeas

mmoqui:


.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
yosifkit: I don't think this file is necessary.

mmoqui: This file is just for informing to git to ignore some IDE specific files (like here the .idea directory created by IntelliJ IDEA)

tianon: Please use local ignores instead (`.git/info/exclude`; see https://help.github.com/articles/ignoring-files/#explicit-repository-excludes).

mmoqui: done

PR #962, opened 2017-07-12
Update docs to filter arch-specific content a bit better (variants, supported tags)

tianon:
This uses the combination of us pushing to a namespace + `BASHBREW_ARCH` being set to determine whether we're pushing to an arch-specific page and should thus appropriately filter all content to the current architecture.

Special thanks to https://github.com/docker-library/official-images/pull/3189 and https://github.com/docker-library/go-dockerlibrary/pull/15 for making this easier to implement.


tianon:
This will allow us to push docs to `amd64/xxx` and have it filter out Windows-specific content, for example.

tianon:
(and will stop `winamd64/xxx` from mentioning Alpine)

PR #968, opened 2017-07-28
Add example on how to make an onbuild image for Perl

zakame:
Instead of providing an `onbuild` variant, show to the user how it is
created instead, by example.

See Perl/docker-perl#13 and
docker-library/official-images#2076

### DIFF
--- a/perl/content.md
+++ b/perl/content.md
@@ -34,7 +34,7 @@ $ docker run -it --rm --name my-running-script -v "$PWD":/usr/src/myapp -w /usr/
 
 ## Creating a reusable `perl:onbuild` image for Perl projects
 
-Suppose you have a project that uses [Carton](https://metacpan.org/pod/Carton) to manage Perl dependencies. You can write a `Dockerfile` that makes use of the [ONBUILD](https://docs.docker.com/engine/reference/builder/#onbuild) instruction like this:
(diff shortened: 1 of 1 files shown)


yosifkit:
I think it would be better to not mention the `ONBUILD` command. It requires the user to manage an extra image build and hides the useful bits (copy/run etc) in a Dockerfile that is separate from the project Dockerfile.  What benefit would this be to a user of the perl image?

zakame:
Thanks for the feedback @yosifkit!

For the example Dockerfile I documented, the main benefit is that the user can have a common `perl:onbuild` image for building base images for  multiple (and possibly unrelated) Carton-based projects (admittedly, I might suggest renaming the image to `perl:carton` instead;) furthermore, I want to provide the user the option of making this `onbuild` image on their own (and thus opt in to the responsibility of using `ONBUILD`, as such,) per your [original comment](https://github.com/docker-library/official-images/issues/2076#issue-172237857).

I think I can still improve on the documentation such that it raises the points you mentioned here, so users would know what to expect (and be guided when raising issues to https://github.com/Perl/docker-perl later.)

yosifkit:
Great description of the situation for users!  Thanks!

zakame:
Thanks @yosifkit! :tada:

PR #969, opened 2017-08-01
add the image to arm64v8

longquan7:
I create the node image in arm64 system


longquan7:
travis-ci use check is in the amd64 is not arm64

longquan7:
update file

longquan7:
update file

longquan7:
update

yosifkit:
Other architecture docs are generated from the regular docs. Considering that this is just a new architecture for a current image it will be best to go through the maintainers of the current Node.js image, the [Node.js Docker Working group](https://github.com/nodejs/docker-node/tree/778cfef0cb007d1173a5baa504c2b0c2c44d924c#governance-and-current-members).  Please use the issue you opened on their repo (https://github.com/nodejs/docker-node/issues/480).  Most likely they will just need to update their scripts to include arm64v8 in their list of supported architectures for the library file in official-images when they want to add it, but they have their own process on what arches they support for the node image.

longquan7:
@yosifkit Ok,thanks

PR #972, opened 2017-08-07
Add MediaWiki

davidbarratt:
Most of this was copied from `drupal` and `wordpress`


magol:
I understand you do not want to add support for allextensions, but it also makes it difficult to use. You have to create your own dockerfile to have some practical benefit from this, and you need to figure out what each individual extension needs to work.

If I wish free, I would like Docker to be able to solve this. I think that Docker can have have extension packages. Then you could run for example `docker run --name myWiki --extension wikimedia/VirtualEditor --extension wikimedia/Cite -d wikimedia/mediawiki`

Extensions are located on hub.docker.com, but are not built. Instead, they contains only source files and must be built locally.

In the example above, images are built in the following order

wikimedia/mediawiki
    wikimedia/VirtualEditor
      wikimedia/Cite

I'm really no docks expert, and I do not know at all if I'm thinking right.

davidbarratt:
This image _should_ support all of the extensions that are included with mediawiki, but there isn't currently a way to enable them from docker itself until [T172555](https://phabricator.wikimedia.org/T172555) is resolved.

As a work-around, you can mount the extensions directory to your host machine (i.e. `--volume=./extensions:/var/www/html/extensions`) and put whatever extensions you want in there.

mediawiki/stack.yml
```diff
@@ -0,0 +1,21 @@
...
+    links:
+      - database
+    volumes:
+      - /var/www/html/images
+  database:
+    image: mariadb
+    environment:
+      # @see https://phabricator.wikimedia.org/source/mediawiki/browse/master/includes/DefaultSettings.php
+      MYSQL_DATABASE: 'my_wiki'
+      MYSQL_USER: 'wikiuser'
```
yosifkit: If the users are meant to use the `wikiuser`, then the password env is required.  `MYSQL_USER` will not be created, since there is no `MYSQL_PASSWORD`.

mediawiki/stack.yml
```diff
@@ -0,0 +1,21 @@
...
+version: '3'
+services:
+  mediawiki:
+    image: mediawiki
+    ports:
+      - 8080:80
+    links:
+      - database
+    volumes:
+      - /var/www/html/images
```
yosifkit: Can the web setup just auto create the `LocalSettings.php` file in the correct place when the directory is writable?  It feels a bit clunky to download the file and then have to get it back into the container (which could be on a swarm).

If we are stuck with downloading the file, I'd recommend having a couple lines here similar to this, but it will only work well on a single host.
```yaml
# after initial setup, download LocalSettings.php to the same directory as this yaml and uncomment the following line and use compose to restart the mediawiki service
#      - ./LocalSettings.php:/var/www/html/LocalSettings.php
```

It might also make sense to

davidbarratt: @yosifkit Sadly, the web install does not write the `LocalSettings.php` file. I created [T172555](https://phabricator.wikimedia.org/T172555) to help deal with that either by allowing for a default `LocalSettings.php` file that we would put in the container (that would read the environment variables) or by allowing the `LocalSettings.php` to not exist at all and _only_ using environment variables (unless overridden).

The `cli` installer _will_ create the`LocalSettings.php` file, but you would have to be in the container (and have a database service up and running).

I will add your suggestion for now and hopefully in the future we can remove it! :)

davidbarratt:
@yosifkit Should be good to go now!

mediawiki/stack.yml
```diff
@@ -0,0 +1,26 @@
...
+      # this yaml and uncomment the following line and use compose to restart
+      # the mediawiki service
+      # - ./LocalSettings.php:/var/www/html/LocalSettings.php
+  database:
+    image: mariadb
+    environment:
+      # @see https://phabricator.wikimedia.org/source/mediawiki/browse/master/includes/DefaultSettings.php
+      MYSQL_DATABASE: 'my_wiki'
+      MYSQL_USER: 'wikiuser'
+      MYSQL_PASSWORD: ''
```
tianon: Empty string won't quite cut it. :wink:

We usually just use `example` as the value here:

https://github.com/docker-library/docs/blob/9828a79d269dfadd0da0371822e742a088f40d51/wordpress/stack.yml#L12-L15

mediawiki/content.md
```diff
@@ -0,0 +1,62 @@
...
+
+## MySQL
+
+```console
+$ docker run --name some-%%REPO%% --link some-mysql:mysql -d %%REPO%%
+```
+
+-	Database type: `MySQL, MariaDB, or equivalent`
+-	Database name/username/password: `<details for accessing your MySQL instance>` (`MYSQL_USER`, `MYSQL_PASSWORD`, `MYSQL_DATABASE`; see environment variables in the description for [`mariadb`](https://registry.hub.docker.com/_/mariadb/))
+-	ADVANCED OPTIONS; Database host: `mysql` (for using the `/etc/hosts` entry added by `--link` to access the linked container's MySQL instance)
```
tianon: According to the `--link` above, this should be `some-mysql` instead of just `mysql`.

mediawiki/content.md
```diff
@@ -0,0 +1,62 @@
...
+
+## %%STACK%%
+
+Run `docker stack deploy -c stack.yml %%REPO%%` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
+
+## Adding additional libraries / extensions
+
+This image does not provide any additional PHP extensions or other libraries, even if they are required by popular plugins. There are an infinite number of possible plugins, and they potentially require any extension PHP supports. Including every PHP extension that exists would dramatically increase the image size.
+
+If you need additional PHP extensions, you'll need to create your own image `FROM` this one. The [documentation of the `php` image](https://github.com/docker-library/docs/blob/master/php/README.md#how-to-install-more-php-extensions) explains how to compile additional extensions.
```
tianon: This `master` link should probably be to a specific commit instead, ala https://github.com/docker-library/docs/blob/31280550a3c7104fef824450753844d2f3d917be/php/README.md#how-to-install-more-php-extensions

magol:
On [hub.docker.com/r/library/mediawiki](https://hub.docker.com/r/library/mediawiki), it is a link to [github.com/wikimedia/mediawiki-docker/issues](https://github.com/wikimedia/mediawiki-docker/issues)
But the issues is disabled for the repository, the issues is now logged in [phabricator.wikimedia.org](https://phabricator.wikimedia.org/T92826).

I think that the possible to add issues should be added to the repository, or the link should be changed.

davidbarratt:
@magol I've requested a new project be created https://phabricator.wikimedia.org/T173959

PR #977, opened 2017-08-10
Add documentation for the rust image

sfackler:
Content was modeled after gcc and golang's.

cc https://github.com/docker-library/official-images/pull/3273


rust/content.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Compile your app inside the Docker container
+
+There may be occasions where it is not appropriate to run your app inside a container. To compiler, but not run your app inside the Docker instance, you can write something like:
+
+```console
+$ docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%IMAGE%%:1.19.0 cargo build --release
+```
+
+This will add your current directory, as a volume, to the container, set the working directory to the volume, and run the command `cargo build --release`. This tells Cargo, Rust's build system, to compile the crate in `myapp` and output the executable to `target/release/myapp`.
```
yosifkit: My only comment is a warning about this example; since it runs as root the folders and compiled files will be owned by root.  You may want to add a `--user "$(id -u)":"$(id -g)"` in the docker run or a sentence in this paragraph about `--user`.

sfackler: Ah, good point. Fixed!

PR #981, opened 2017-08-15
composer: fix example, phrasing

alcohol:


composer/content.md
```diff
@@ -61,7 +61,7 @@ docker run --rm --interactive --tty \
 
 ### PHP Extensions
 
-We strive to deliver an image that is as lean as possible, aimed at running Composer only.
+We aim to deliver an image that is as lean as possible, build for running Composer only.
```
yosifkit: Probably swap `s/build/built/`?

alcohol: You never saw that. 🤔

PR #985, opened 2017-08-21
Add note about WiredTiger being the default storage engine.

knpwrs:
Fixes https://github.com/docker-library/mongo/issues/197.


tianon:
Thanks!

PR #986, opened 2017-08-22
update note about exec influxdb

rdxmb:


rdxmb:
just edited the old note about the exec-bug in docker

tianon:
Seems OK -- cc @jsternberg

jsternberg:
Is the bug resolved now?

Also, can we do `-it` instead of `-ti`? It doesn't matter, but I like how it looks better and it's how I always write it so it would be more consistent with the rest of the documentation.

jsternberg:
Also, if the bug is resolved, can we make a note of which version it was resolved in and a note of how to get around the bug in older versions? I just don't want to get issues on our issue tracker about this bug if someone is using an old version of Docker.

rdxmb:
I have started with Docker version 17.06.0-ce, build 02c1d87 and never had the problem with using exec.
https://github.com/moby/moby/issues/8755 is closed as resolved - with milestone 17.06.0
However, I do not find the fix in the release notes https://docs.docker.com/release-notes/docker-ce/
Maybe someone else wants to try if this is a problem in older versions?

'-ti' is the same as '-it' , as both are a short version of '-i -t' or '-t -i'
so feel free about that

rdxmb:
just took some time to edit the PR with your wishes. Unfortunately the travis check failed at this time. Don't know whats the problem here. I am not that well versed in Markdown.
I'm sorry that I cannot spend any more time on this. Just wanted to keep the docs uptodate. I did not foresee that this will take so much time with installing dependencies on my machine ...
Thank you for your understanding.

yosifkit:
Sorry for the slow reply. :bowing_man:

Since the related Docker bug was fixed in `17.06.0-ce` and Docker supported versions are currently `17.09.x` and `17.12.x`, I think this could be safely changed without needing some extra description for old versions.

cc @jsternberg, What do you think? A rebase and good to go? Make your own PR?

image
 - image from https://blog.docker.com/2017/03/docker-enterprise-edition/

jsternberg:
👍 

@rdxmb can you rebase this so it can be merged?

rdxmb:
@jsternberg yes, if you tell me how to do that. I have never done that before ;)

jsternberg:
I just opened my own PR that removes the note in #1116. It also keeps the method for showing how to use a separate container just in case somebody wants that.

rdxmb:
looks good to me.

tianon:
Replaced by #1116 :+1:

PR #998, opened 2017-09-08
add docker-compose example for bonita

JeremJR:
to finish answering to https://github.com/Bonitasoft-Community/docker_bonita/issues/5

### DIFF
--- a/bonita/content.md
+++ b/bonita/content.md
@@ -75,6 +75,10 @@ $ docker run --name=bonita -e "TENANT_LOGIN=tech_user" -e "TENANT_PASSWORD=secre
 
 Now you can access the Bonita BPM Portal on localhost:8080/bonita and login using: tech_user / secret
 
+## %%STACK%%
+
+Run `docker stack deploy -c stack.yml %%REPO%%` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
+
 ## Where to store data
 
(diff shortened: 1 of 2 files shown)


yosifkit:
Sorry for the slow reply.  This stack file is a bit complicated by making scripts from an `sh -c` in yaml that are then run.  Here is my slightly simplified version that should make it a little easier to understand and has a limit to the retries so that the bonita container can fail if the postgres doesn't come up.

```yaml
version: '3'

services:
  db:
    image: postgres:9.3
    environment:
      POSTGRES_PASSWORD: example
    restart: always
    command:
      - -c
      - max_prepared_transactions=100
  bonita:
    image: bonita
    ports:
      - 8080:8080
    environment:
      - POSTGRES_ENV_POSTGRES_PASSWORD=example
      - DB_VENDOR=postgres
      - DB_HOST=db
      - TENANT_LOGIN=tech_user
      - TENANT_PASSWORD=secret
      - PLATFORM_LOGIN=pfadmin
      - PLATFORM_PASSWORD=pfsecret
    restart: always
    depends_on:
      - db
    entrypoint:
      - bash
      - -c
      - |
        set -e
        echo 'Waiting for Postgres to be available'
        export PGPASSWORD="$$POSTGRES_ENV_POSTGRES_PASSWORD"
        maxTries=10
        while [ "$$maxTries" -gt 0 ] && ! psql -h "$$DB_HOST" -U 'postgres' -c '\l'; do
          sleep 1
        done
        echo
        if [ "$$maxTries" -lte 0 ]; then
          echo >&2 'error: unable to contact postgres after 10 tries'
          exit 1
        fi
        exec /opt/files/startup.sh
```

JeremJR:
Sorry for the slow reply too ^^'. Thanks a lot for your answer, I've applied your advice.

PR #1001, opened 2017-09-11
added note about mounting volumes with uid other than 1000

keeganwitt:


keeganwitt:
Can this be merged soon?

PR #1007, opened 2017-09-15
Update README.md for Kong

jonocodes:
Remove -it from migration run since it was tripping up a script I was writing. This is not needed because it is not being used interactively here.


tianon:
LGTM

fyi @thefosk

PR #1008, opened 2017-09-17
Add port option in the example without a Dockerfile

FlorianOnmyown:
Add '-p 80:80' to the example without a Dockerfile.
This expose the port 80 so that users can conect to the server through a web client.

### DIFF
--- a/httpd/content.md
+++ b/httpd/content.md
@@ -29,8 +29,9 @@ $ docker run -dit --name my-running-app my-apache2
 If you don't want to include a `Dockerfile` in your project, it is sufficient to do the following:
 
 ```console
-$ docker run -dit --name my-apache-app -p 80:80 -v "$PWD":/usr/local/apache2/htdocs/ httpd:2.4
+$ docker run -dit --name my-apache-app -p 8080:80 -v "$PWD":/usr/local/apache2/htdocs/ httpd:2.4
 ```
+Please be aware that exposing a port may require root priviliges on your host, depanding your configuration.
 
 ### Configuration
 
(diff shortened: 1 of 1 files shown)


themccallister:
@FlorianOnmyown it might be worth while to mention that this may require root permissions to bind to the port, just food for thought.

httpd/content.md
```diff
@@ -29,7 +29,7 @@ $ docker run -dit --name my-running-app my-apache2
 If you don't want to include a `Dockerfile` in your project, it is sufficient to do the following:
 
 ```console
-$ docker run -dit --name my-apache-app -v "$PWD":/usr/local/apache2/htdocs/ httpd:2.4
+$ docker run -dit --name my-apache-app -p 80:80 -v "$PWD":/usr/local/apache2/htdocs/ httpd:2.4
```
tianon: I'm +1, but we usually try to use a different host port (like `8080`) to help reinforce that there's no inherent dependency between the two port values (so `-p 8080:80`, for example).

FlorianOnmyown:
@themccallister @tianon I have implemented the suggested changes, can you have a look?

httpd/content.md
```diff
@@ -29,9 +29,11 @@ $ docker run -dit --name my-running-app my-apache2
 If you don't want to include a `Dockerfile` in your project, it is sufficient to do the following:
 
 ```console
-$ docker run -dit --name my-apache-app -v "$PWD":/usr/local/apache2/htdocs/ httpd:2.4
+$ docker run -dit --name my-apache-app -p 8080:80 -v "$PWD":/usr/local/apache2/htdocs/ httpd:2.4
 ```
 
+Please be aware that exposing a port may require root priviliges on your host, depanding your configuration.
```
tianon: Docker takes care of this, so this does not apply.

FlorianOnmyown:
@tianon The comment has been removed.

PR #1011, opened 2017-09-18
Add some notes in rabbitmq about "RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS"

tianon:
See https://github.com/docker-library/rabbitmq/pull/156#issuecomment-305346747 and the following discussion (and https://github.com/docker-library/rabbitmq/pull/159).


tianon:
This should be less of an issue with 3.7+ (being discussed in https://github.com/docker-library/rabbitmq/pull/197), since it has a newer (simpler) configuration file format. :metal:

PR #1017, opened 2017-09-26
amend Composer documentation

alcohol:
- [x] mention ssh-agent auth socket issue on osx
- [x] document multi-stage option

### DIFF
--- a/composer/content.md
+++ b/composer/content.md
@@ -81,7 +81,7 @@ Sometimes dependencies or Composer [scripts](https://getcomposer.org/doc/article
 
 -	Create your own image, and copy Composer from the official image into it:
 
-	```
+	```dockerfile
 	COPY --from=composer:1.5 /usr/bin/composer /usr/bin/composer
 	```
 
--- a/composer/content.md
+++ b/composer/content.md
@@ -36,6 +36,8 @@ docker run --rm --interactive --tty \
 
 When you need to access private repositories, you will either need to share your configured credentials, or mount your `ssh-agent` socket inside the running container:
 
+**Note:** This currently does not work on OSX, see [docker/for-mac#410](https://github.com/docker/for-mac/issues/410).
+
 ```sh
 docker run --rm --interactive --tty \
     --volume $PWD:/app \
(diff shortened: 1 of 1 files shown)


alcohol:
I do not understand why the build failed :-\

composer/content.md
```diff
@@ -75,6 +77,16 @@ Sometimes dependencies or Composer [scripts](https://getcomposer.org/doc/article
 
 -	Create your own image (possibly by extending `FROM composer`).
 
+**Note:** Docker introduced [multi-stage](https://docs.docker.com/engine/userguide/eng-image/multistage-build/) builds in 17.05:
+
+-	Create your own image, and copy Composer from the official image into it:
+
+	```
```
tianon: Add `dockerfile` as the "format" for this code block and Travis should be happy :+1:

composer/content.md
```diff
@@ -75,6 +77,16 @@ Sometimes dependencies or Composer [scripts](https://getcomposer.org/doc/article
...
+	COPY --from=composer:1.5 /usr/bin/composer /usr/bin/composer
```
tianon: `--from` references a stage AFAIK, so this needs to be either a number, or a stage name, right?

tianon: Indeed, see https://docs.docker.com/engine/reference/builder/#copy:
> Optionally COPY accepts a flag --from=<name|index> that can be used to set the source location to a previous build stage (created with FROM .. AS <name>) that will be used instead of a build context sent by the user. The flag also accepts a numeric index assigned for all previous build stages started with FROM instruction. In case a build stage with a specified name can’t be found an image with the same name is attempted to be used instead.

tianon: Ok, my reading comprehension skills are low -- it literally does copy from another image, which is terrifying for official images (for similar reasons to what's discussed in https://github.com/docker-library/official-images/issues/3383#issuecomment-329259161 but with the side benefit of being even harder to scrape properly), but in the context of the docs that's totally fine. :innocent: :+1:

tianon:
Content looks good, just needs `dockerfile` added to that one code block and we're good to go. :+1:

composer/content.md
alcohol: Admittedly, that is also how I understood it the first time I read it. That last sentence is a bit tricky.

composer/content.md
alcohol: Done 👍

PR #1021, opened 2017-09-28
nginx: ensure best practice when using envsubst

lenartj:
Background: `envsubst` replaces all references that look like references
to variables including those that do not have a corresponding environment
variable - it just replaces those with an empty string. Since it's very
common for nginx.conf to have references like $uri or $host, which are
not definately not to be replaced by environment variables, there is a
need to limit the scope of `envsubsts` by specifying each variable name.

https://stackoverflow.com/questions/46473266/nginx-on-kubernetes-docker-doing-infinite-redirect-when-generating-conf

Also, using `exec nginx` is better, since we can avoid an extra shell
being run and can also make sure that nginx receives the signals
correctly.

### DIFF
--- a/nginx/content.md
+++ b/nginx/content.md
@@ -82,7 +82,7 @@ web:
   environment:
    - NGINX_HOST=foobar.com
    - NGINX_PORT=80
-  command: /bin/bash -c "envsubst < /etc/nginx/conf.d/mysite.template > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'"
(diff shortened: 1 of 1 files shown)


yosifkit:
Build failure is unrelated, LGTM.

cc maintainer @thresheek

lenartj:
I have moved the change into `content.md`

tianon:
Sorry for the long tail :bow:

This was fixed in the image via https://github.com/nginxinc/docker-nginx/commit/cd98e75ce640e77633cd79035e41c9a122897fdf (automatic `envsubst`) and in the documentation in https://github.com/docker-library/docs/pull/1471. :+1:

PR #1034, opened 2017-10-07
[Nextcloud] Add Play with Docker button

tilosp:
nextcloud/docker#166

### DIFF
--- a/nextcloud/content.md
+++ b/nextcloud/content.md
@@ -14,6 +14,8 @@ The `apache` tag contains a full Nextcloud installation including an apache web 
 
 The second option is a `fpm` container. It is based on the [php-fpm](https://hub.docker.com/_/php/) image and runs a fastCGI-Process that serves your Nextcloud page. To use this image it must be combined with any webserver that can proxy the http requests to the FastCGI-port of the container.
 
+[![Try in PWD](https://github.com/play-with-docker/stacks/raw/cff22438cb4195ace27f9b15784bbb497047afa7/assets/images/button.png)](http://play-with-docker.com?stack=https://raw.githubusercontent.com/nextcloud/docker/dba66a2d18b1020f6de6e22defd1b5e6df87c894/stack.yml)
+
 ## Using the apache image
 
 The apache image contains a webserver and exposes port 80. To start the container type:


tilosp:
@tianon please review

tianon:
That `cron` service is really hacky -- how are users normally expected to implement `cron` for Nextcloud, especially in a Docker context?

tilosp:
@tianon At the moment there are two ways. One is the hacky docker-compose one and the other one is this [Dockerfile](https://github.com/nextcloud/docker/blob/master/.examples/dockerfiles/cron/apache/Dockerfile).
We could make the docker-compose way less hacky by adding the script to the container. And then just change the entrypoint in the `docker-compose.yml` for the cron container.

tianon:
How about this?
```dockerfile
FROM nextcloud:apache

# busybox's crond is more flexible than what's provided by the "cron" package
# (especially around logging)
RUN apt-get update && apt-get install -y busybox-static && rm -rf /var/lib/apt/lists/*

RUN echo '*/15 * * * * www-data php -f /var/www/html/cron.php' > /etc/crontab

WORKDIR /var/spool/cron/crontabs
CMD ["busybox", "crond", "-f", "-l", "0", "-L", "/dev/stdout"]
```

Combined with:
```yaml
version: '3.2'

services:
  db:
    image: postgres
    restart: always
    volumes:
      - db:/var/lib/postgresql/data
    environment:
      - POSTGRES_DB=nextcloud
      - POSTGRES_USER=nextcloud
      - POSTGRES_PASSWORD=nextcloud

  app:
    image: nextcloud
    restart: always
    ports:
      - 8080:80
    volumes:
      - nextcloud:/var/www/html
    environment:
      - POSTGRES_HOST=db
      - POSTGRES_DB=nextcloud
      - POSTGRES_USER=nextcloud
      - POSTGRES_PASSWORD=nextcloud
    depends_on:
      - db

  cron:
    image: nextcloud:cron
    restart: always
    init: true
    volumes:
      - nextcloud:/var/www/html
    depends_on:
      - db

volumes:
  db:
  nextcloud:
```

You could even simply include `busybox-static` with the pre-configured `crontab` in the base `nextcloud` image, and add a `nextcloud-crond` script which is basically just:

```bash
#!/usr/bin/env bash
set -Eeuo pipefail
mkdir -p /var/spool/cron/crontabs
exec busybox crond -f -l 0 -L /dev/stdout "$@"
```

So you'd just use:
```yaml
    image: nextcloud
    command: nextcloud-crond
    init: true
```
(and `init: true` is only necessary if you want to be able to Ctrl+C to kill it interactively -- it seems to respond fine to the SIGTERM that `docker stop` sends without `--init`)

tilosp:
@pierreozoux What do you think?

pierreozoux:
I like the second option with the `nextcloud-crond` script and the different command in the compose, I find it nicer.
(yes, it was hacky, but still works :) also had a lot of trouble to find a nice crond script that was minimal for docker, but now I'll know! Thanks!)

tilosp:
Started to implement the second option in https://github.com/nextcloud/docker/pull/220

PR #1038, opened 2017-10-11
fix running consul for development

gotcha:
At least under ``Docker version 17.07.0-ce-rc4``,  running consul for development does not work
without being explicit about ``CONSUL_BIND_INTERFACE``


tianon:
cc @slackpad

slackpad:
Hi @gotcha this LGTM. There's also a built-in way to get an interface IP in Consul, `-bind='{{ GetInterfaceIP "eth0" }}'` should be equivalent.

PR #1039, opened 2017-10-13
update jetty documentation for latest images

gregw:
update jetty documentation for latest images. 

@md5 I ended up doing a bit of refactoring here, following the tomcat doco as an example.

### DIFF
--- a/jetty/README.md
+++ b/jetty/README.md
@@ -53,27 +53,27 @@ WARNING:
 
 # What is Jetty?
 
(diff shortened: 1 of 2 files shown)


tianon:
Whew, this one has a long tail, sorry! :bow:
(I think we were probably waiting for @md5, and also hadn't gotten around to pointing out the failing CI :see_no_evil: :disappointed:)

Looks like it's probably outdated - did you want to take a stab at updating it, or do you think it's probably best to just close instead?

tianon:
I'm going to close this for now -- please feel free to rebase/reopen if it is still relevant! (Sorry :bow:)

Issue #1046, opened 2017-10-25
Missing license.md files


tianon:
Several images here are missing `license.md`, and we'd like to add that to Travis for verifying that everything has it.  The full list of images currently missing `license.md` is as follows:

- [x] alpine: #1089
- [x] amazonlinux: #1061
- [x] cassandra: #1048
- [x] centos: #1090
- [x] cirros: #1091
- [x] clearlinux: #1047
- [x] crux: #1048
- [x] debian: #1048
- [x] fedora: #1049
- [x] ghost: #1048
- [x] hello-seattle: #1048
- [x] hello-world: #1048
- [x] hola-mundo: #1048
- [x] jenkins: #1048
- [x] mageia: #1054
- [x] mariadb: #1048
- [x] mongo-express: #1048
- [x] mysql: #1048
- [x] neo4j: #1086
- [x] neurodebian: #1048
- [x] nginx: #1057
- [x] opensuse: #1051
- [x] orientdb: #1085
- [x] percona: #1048
- [x] postgres: #1048
- [x] registry: #1092
- [x] scratch: literally the empty image; no reason for `license.md`
- [x] ubuntu: #1048
- [x] wordpress: #1048

tianon:
@ncopa do you know what we ought to list for the license of Alpine Linux itself?

@yarikoptic should this blurb be any different for `neurodebian` than it might be for Debian itself?

cc @ilianaw @praveen-pk @Eric-Warehime (`amazonlinux`)
cc @jperrin (`centos`)
cc @ewindisch (`cirros`)
cc @bryteise (`clearlinux`)
cc @maxamillion @puiterwijk (`fedora`)
cc @juanluisbaptiste (`mageia`)
cc @knickers (`mongo-express`)
cc @benbc @spacecowboy @srbaker @praveenag (`neo4j`)
cc @thresheek (`nginx`)
cc @flavio @cyphar @jordimassaguerpla @davidcassany (`opensuse`)
cc @robfrank (`orientdb`)
cc @stevvooe @dmp42 @RichardScothern @aaronlehmann @dmcgowan (`registry`)

If any of you don't have the time for a PR for this, feel free to simply provide the relevant one-liner I ought to put into `license.md` and I'll be happy to make the update on your behalf. :+1:

Most of the existing files use something similar to the following, but any free-form description of the image license information will do:

```markdown
View [license information](URL-TO-UPSTREAM-LICENSE-INFO) for the software contained in this image.
```

[...]

PR #1051, opened 2017-10-26
opensuse: add license information

cyphar:
Ref: #1046
Signed-off-by: Aleksa Sarai <asarai@suse.de>

/cc @flavio -- Can you double-check that this is correct?


flavio:
LGTM

tianon:
LGTM :heart:

PR #1059, opened 2017-11-06
Add documentation for a Scientfic Linux official container

sl-team:
This PR adds documentation for Scientific Linux


sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+
+A tag will be provided for each of the [maintained major releases](http://www.scientificlinux.org/downloads/sl-versions/) of Scientific Linux (`6`, `7`).
+
+The `latest` tag will track the highest version numbered release of Scientific Linux.
+
+Updated images will be released roughly once per month.
+
+## Build Logs
+
+The SL docker containers are built on our internal build system and packaged at [Docker Cloud](https://cloud.docker.com/app/scientificlinux/repository/docker/scientificlinux/sl/builds/).
```
tianon: This link appears to be throwing a 500 error. :confused:

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+By default the SL docker images do not include these files.  If you require them, please remove `tsflags=nodocs` from `/etc/yum.conf` and run `yum reinstall mypackage` to recieve the documentation.
+
+---
+# Quick Reference
+
+## Example Usage
+You can try out the containers via:
+
+```
+$ docker pull scientificlinux/sl
```
tianon: This should be `%%IMAGE%%` instead of `scientificlinux/sl`.

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+# RUN yum -y install httpd; yum clean all; systemctl enable httpd.service
+# EXPOSE 80
+### To enable apache within the container uncomment the previous two lines
+VOLUME [ "/sys/fs/cgroup" ]
+CMD ["/usr/sbin/init"]
+```
+
+You can build and run this example (with apache) via:
+```
+$ docker build --rm -t local/mycontainer
```
tianon: This is missing the "directory" argument (probably want just `.`?)

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+The SL7 docker container ships with systemd mostly functional.  You can build a SL7 systemd enabled container with the following Dockerfile
+
+In order to run a container with systemd, you will need to mount the cgroups volumes from the host.
+
+```
+# Example SL7 systemd Dockerfile
+FROM %%IMAGE%%:7
+ENV container docker
+### To enable apache within the container uncomment the next two lines
+# RUN yum -y install httpd; yum clean all; systemctl enable httpd.service
```
tianon: Shouldn't this be using `&&` instead of `;` so that it short-circuits on failure? (or `set -e;` / `set -ex;` at the beginning)

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+You can build and run this example (with apache) via:
+```
+$ docker build --rm -t local/mycontainer
+$ docker run -ti -v /sys/fs/cgroup:/sys/fs/cgroup:ro -p 80:80 local/mycontainer
+```
+Which will run systemd within the container in a limited context.
+
+It is recommended that you install any relevant [OCI hooks](https://www.opencontainers.org/) for your container host - such as `oci-register-machine` or `oci-systemd-hook`.
+
+Some container hosts must add `-v /tmp/$(mktemp -d):/run` to the `docker run` command.
```
tianon: Doesn't `mktemp -d` return a full path?  Why include a `/tmp/` prefix here?

tianon: Also, this could be replaced with `--tmpfs /run` instead, which might be simpler overall.

tianon:
(I've also taken the liberty of rebasing this so that Travis will run and provide additional checks.)

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+
+By default the SL docker images do not include these files.  If you require them, please remove `tsflags=nodocs` from `/etc/yum.conf` and run `yum reinstall mypackage` to recieve the documentation.
+
+---
+# Quick Reference
+
+## Example Usage
+You can try out the containers via:
+
+```
```
yosifkit: You'll want "```console" on this line so that the code block gets colors

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+$ docker run -it %%IMAGE%%:6 cat /etc/redhat-release
+$ docker run -it %%IMAGE%%:7 cat /etc/redhat-release
+```
+
+## Enabling systemd in SL7
+The SL7 docker container ships with systemd mostly functional.  You can build a SL7 systemd enabled container with the following Dockerfile
+
+In order to run a container with systemd, you will need to mount the cgroups volumes from the host.
+
+```
```
yosifkit: You'll want "```Dockerfile" on this line so that the code block gets colors

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+### To enable apache within the container uncomment the next two lines
+# RUN yum -y install httpd; yum clean all; systemctl enable httpd.service
+# EXPOSE 80
+### To enable apache within the container uncomment the previous two lines
+VOLUME [ "/sys/fs/cgroup" ]
+CMD ["/usr/sbin/init"]
+```
+
+You can build and run this example (with apache) via:
+```
```
yosifkit: "```console" here as well :smile:

sl/content.md
tianon: This should be a typed block, ala:
~~~markdown
```console
$ ...
```
~~~

sl/content.md
```diff
@@ -0,0 +1,89 @@
+# Scientific Linux
+
+%%LOGO%%
+
+This image containers for [Scientific Linux](http://www.scientificlinux.org/).
```
yosifkit: This sentence seems a little wonky.  Maybe "These images contain Scientific Linux"?

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+
+The source for each release is in [its own branch](http://github.com/scientificlinux/sl-docker/branches) and contains the kickstart used to build the image.
+
+These images are maintained by the Scientific Linux Team in [github](https://github.com/scientificlinux/sl-docker).
```
yosifkit: Since you have a shortened version of this in `maintainer.md`, this seems repetitive.  `maintainer.md` will end up in the Quick Reference section (ex. [arango](https://github.com/docker-library/docs/tree/3383325346b0ac532752d2f3725d1c654c3a3227/arangodb#quick-reference)) under the "Maintained By" bullet point.

sl/content.md
yosifkit: Also, maybe swap "docker containers" for "root file system for the docker images"?  Since the resulting images are actually built via the official-images build servers.

sl/content.md
yosifkit: Why not just leave it uncommented and have an initial comment like `# Example enabling httpd in systemd`?

sl/content.md
```diff
@@ -0,0 +1,89 @@
...
+$ docker run -ti -v /sys/fs/cgroup:/sys/fs/cgroup:ro -p 80:80 local/mycontainer
+```
+Which will run systemd within the container in a limited context.
+
+It is recommended that you install any relevant [OCI hooks](https://www.opencontainers.org/) for your container host - such as `oci-register-machine` or `oci-systemd-hook`.
+
+Some container hosts must add `-v /tmp/$(mktemp -d):/run` to the `docker run` command.
+
+---
+# About Scientific Linux
```
yosifkit: I would put this "About SL" section first, but it is up to you. 

In any case, the resulting docs will have two generated sections added to the beginning:

- Supported tags and respective Dockerfile links (ex [arango](https://github.com/docker-library/docs/tree/3383325346b0ac532752d2f3725d1c654c3a3227/arangodb#supported-tags-and-respective-dockerfile-links))
- Quick Reference (ex [arango](https://github.com/docker-library/docs/tree/3383325346b0ac532752d2f3725d1c654c3a3227/arangodb#quick-reference))

And then the "License" section added to the end (ex. [arango](https://github.com/docker-library/docs/tree/3383325346b0ac532752d2f3725d1c654c3a3227/arangodb#license)).

sl-team:
I believe I've made the requested changes and cleaned up some of the redundant bits.

Issue #786, opened 2016-12-28
update mongodb docs, --link is deprecated


AlJohri:

tianon:
As far as I'm aware, `--link` is _discouraged_, but not officially deprecated yet (and is the simplest form of connecting two containers to one another).

dhoinka:
According to the docs it is.

> Warning: The --link flag is a deprecated legacy feature of Docker.
[docker docs](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/)

tianon:
See https://github.com/docker-library/docs/pull/1063 for another discussion of this same topic.

Essentially, `--link` is still actively supported, and does not print a deprecation warning on use.  There are no near-term plans that I'm aware of to remove the feature it provides.

We use `--link` in the documentation as a very short from of `docker network` which doesn't require explicit network creation or cleanup, and the astute Docker reader ought to be able to make the conversion to networks.  I'd be open to adding a short blurb to each (first-per-image-description) usage of `--link` to link to some good `docker network` documentation, but we'll need to come up with good wording for it before we copy-paste it to every description (which in itself feels like we're trending down the wrong path).

It definitely feels completely backwards to pollute every image description with a long-form example that sets up custom networks just for the purposes of showing how to connect a database to an application container.

tianon:
See also https://github.com/docker-library/faq#--link-is-deprecated. 👍

PR #1063, opened 2017-11-09
Delete --link flag in the mysql doc.

seiyab:
Document of mysql image uses deprecated --link flag.
https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/

Avoided --link flag using docker network.

I worry about my English because I'm not an English speaker.

### DIFF
--- a/aerospike/README.md
+++ b/aerospike/README.md
@@ -17,9 +17,9 @@ WARNING:
 # Supported tags and respective `Dockerfile` links
 
 -	[`3.12.1.3` (*Dockerfile*)](https://github.com/aerospike/aerospike-server.docker/blob/eb4758c8bcb30158cef593383ac0aab119b066c1/Dockerfile)
--	[`3.13.0.7` (*Dockerfile*)](https://github.com/aerospike/aerospike-server.docker/blob/0b1246af2674c7668b332964b27dfef6773935fe/Dockerfile)
--	[`3.14.1.4` (*Dockerfile*)](https://github.com/aerospike/aerospike-server.docker/blob/a4a15147e1e3b782fe365cd8d2d3e686426f0a03/Dockerfile)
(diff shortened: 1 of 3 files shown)


yosifkit:
As much as I'd like to switch away from the _"may eventually be removed"_ `link` flag, it makes one time samples effortless to read and run. It also doesn't pollute the user's machine with random unused networks after they try out a demo. :disappointed:

seiyab:
I see. I agree with you. 😞 

Now, I omitted docker network. How about this one ?
It doesn't create new network. However, the command may be complex.

tianon:
See https://github.com/docker-library/docs/issues/786 for another discussion of this same topic.

I'm still a fan of `--link`, and my rationale for being so is in https://github.com/docker-library/docs/issues/786#issuecomment-350491848.

ivor52:
from what i understand, then, there will be no update of the documentation?  i think the overall point is being missed.  as long as --link is supported, there is nothing wrong with using it since it is simpler to make examples.
however, this does not address the issue: once the --link option is removed (since it will be be if it's been deprecated), the documentation will be wrong and won't even work anymore.
i'm not sure i understand why a single example of using the new docker convention (specifying a network) cannot be included for completeness, not to mention that once --link does go away, the documentation will be already updated for the new scenario.

tianon:
@ivor52 `--link` is deprecated, yes, but in documentation only (I'd call it a "soft deprecation")

For Docker's normal deprecation policy, the deprecated functionality needs to throw a warning for some number of releases before it's removed.  With `--link`, its usage is essentially only discouraged, not actually deprecated, and there's no current plan to remove that functionality which I'm aware of.

Currently, the full docs for all official images have no less than 104(!) usages of `--link`.  To suggest that _every_ one of those be updated to a long-form example that includes a full `docker network create ...` etc is pretty aggressive.  If/when `--link` is _actually_ deprecated, I'd likely be more in favor of adjusting the documentation to be more vague in those 104 places and to instead punt to some of Docker's upstream documentation for "how to configure containers to communicate with one another" (and instead simply describe what the built-in expectations or control knobs of the image are for using those Docker upstream features).

PR #1069, opened 2017-11-21
Add url env var in ghost

pierreozoux:


pierreozoux:
Fix https://github.com/TryGhost/Ghost/issues/9265

ghost/stack.yml
```diff
@@ -17,6 +17,7 @@ services:
       database__connection__user: root
       database__connection__password: example
       database__connection__database: ghost
+      url: example.org
```
tianon: I believe this format is wrong -- this should be a full base URL, including protocol and port (as in the examples linked at the top of this section of `stack.yml`)

pierreozoux:
@tianon sorry, fixed!

tianon:
I tested this via https://labs.play-with-docker.com/?stack=https://github.com/pierreozoux/docs/raw/9fe78a889cb0a29795f59514bbaf4d2fefd469f3/ghost/stack.yml (which is similar to the link that'll be generated and included below this in the documentation) and it has the quirk that it redirects to `https`, which is one reason we didn't include `url` in this example before.  Perhaps we should make this a comment instead, to simply show the format?

tianon:
Superseded by https://github.com/docker-library/docs/pull/1306 :+1:

PR #1070, opened 2017-11-21
Add instruction for ghost image

pierreozoux:


ghost/content.md
```diff
@@ -14,6 +14,8 @@ This will start a Ghost instance listening on the default Ghost port of 2368.
 $ docker run -d --name some-ghost %%IMAGE%%
 ```
 
+Once it is up, you can navigate to `/admin` to create your user account.
```
tianon: It looks like `/admin` redirects to `/ghost` -- also, I think this might contextually more useful following the line below about accessing via `localhost`/`host-ip` in a browser (since this particular section doesn't talk about how to access the running instance at all).

tianon:
Closing old PR.

PR #1072, opened 2017-11-22
fix the syntax for geonetwork docker environment

metazool:
in the environment stanza of stack.yml:

* added a '-'
* replaced : with =

### DIFF
--- a/geonetwork/stack.yml
+++ b/geonetwork/stack.yml
@@ -12,9 +12,9 @@ services:
       image: geonetwork
       restart: always
       ports:
-          - 8080:8080
+         - 8080:8080
       environment:
-          DATA_DIR: /var/lib/geonetwork_data
+         - DATA_DIR=/var/lib/geonetwork_data
       volumes:
          - geonetwork:/var/lib/geonetwork_data
 


yosifkit:
`environment` can be either a list (`- key=val`) or a hash (`key: val`), but I think we should stick with hash since that is what all but one `stack.yml` have in this repo.

geonetwork/stack.yml
```diff
@@ -12,9 +12,9 @@ services:
       image: geonetwork
       restart: always
       ports:
-          - 8080:8080
+         - 8080:8080
       environment:
-          DATA_DIR: /var/lib/geonetwork_data
+         - DATA_DIR=/var/lib/geonetwork_data
```
doublebyte1: @metazool  Could you change this to use keep the ":", instead of the "="? I agree with the other changes

doublebyte1:
Thanks for the contribution @metazool . I agree with @yosifkit that it looks more uniform if we stick to the same separator in every entry. All your other changes look fine!

tianon:
We've since applied `ymlfmt` to get consistent whitespacing for all YAML files within this repository (https://github.com/docker-library/docs/pull/1077). :+1:

Thanks!

PR #1078, opened 2017-11-29
Add YOURLS container documentation

LeoColomb:
Proposing [YOURLS](https://yourls.org) (https://github.com/YOURLS/YOURLS) for creating a new official image.


yourls/content.md
```diff
@@ -0,0 +1,56 @@
...
+
+> [github.com/YOURLS/YOURLS](https://github.com/YOURLS/YOURLS)
+
+%%LOGO%%
+
+# How to use this image
+
+```console
+$ docker run --name some-%%REPO%% --link some-mysql:mysql -d %%IMAGE%% \
+    -e YOURLS_SITE="https://example.com"
```
tianon: This `-e` needs to be before `%%IMAGE%%` to work, doesn't it?

yourls/content.md
```diff
@@ -0,0 +1,56 @@
...
+$ docker run --name some-%%REPO%% --link some-mysql:mysql -p 8080:80 -d %%IMAGE%%
+```
+
+Then, access it via `http://localhost:8080` or `http://host-ip:8080` in a browser.
+
+If you'd like to use an external database instead of a linked `mysql` container, specify the hostname and port with `YOURLS_DB_HOST` along with the password in `YOURLS_DB_PASS` and the username in `YOURLS_DB_USER` (if it is something other than `root`):
+
+```console
+$ docker run --name some-%%REPO%%s -e YOURLS_DB_HOST=10.1.2.3:3306 \
+    -e YOURLS_DB_USER=... -e YOURLS_DB_PASS=... -d %%REPO%%
```
tianon: `%%REPO%%` here at the end should be `%%IMAGE%%` instead (so the multi-arch descriptions have the right `ARCH/yourls` image name here)

tianon:
(If you don't get to those before @yosifkit reviews https://github.com/docker-library/official-images/pull/3752, we'll likely just take care of them directly to get this merged today.)

yourls/content.md
LeoColomb: You seem right. I took WordPress template as example: https://github.com/docker-library/docs/blob/8117c12eda4829c1c57eb07c02a62028d1e2bfe8/wordpress/content.md

yourls/content.md
LeoColomb: Oooh, right!

LeoColomb:
@tianon Thanks for the review, I updated to reflect your comments.

PR #1079, opened 2017-11-30
Update Example of Connecting to Postgres Container

doublebyte1:
Instead of the legacy [container links](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/), let the geonetwork container discover the postgres container by container name, using auser-defined network (recommended approach).
The publishing of the postgres container port on the localhost was removed, as geonetwork does not connect through it.


geonetwork/variant-postgres.md
```diff
@@ -8,20 +8,25 @@ In order to setup the connection from geonetwork, you **must** inject the follow
...
+$ docker network create --driver bridge mynet
+```
+
+Then if you want to run the official image for postgres, you could launch it like this:
 
 ```console
-$ docker run --name some-postgres -p 5432:5432 -d postgres
+$ docker run --name postgres --network=mynet -d postgres
 ```
+In order to enable geonetwork to discover it, **make sure you name it `postgres`**.
```
tianon: Wouldn't it be more flexible for users to instead recommend that they use `--network-alias postgres` so that it's more clear they can name their container whatever is useful for them?

tianon: Also, isn't there an environment variable they can use on the `geonetwork` container to adjust which hostname it attempts to connect to?

doublebyte1: @tianon you are right: I simplified the explanation a bit, by using the $POSTGRES_DB_HOST variable in both use cases (IP addresses and container names) 👍 . Please have a look at my latest [PR] (https://github.com/docker-library/docs/pull/1087). Thanks!

doublebyte1:
I am closing this PR, in favour of [this](https://github.com/docker-library/docs/pull/1087) one.

PR #1080, opened 2017-11-30
Introducing i386 to README.md

ericcurtin:


yosifkit:
This will happen automatically now that https://github.com/docker-library/official-images/pull/3759 is merged. The next run of https://doi-janky.infosiftr.net/job/docs/job/library/ should pick it up.

PR #1081, opened 2017-12-03
Add HTTP/2 info to nginx

jareddlc:
add a section about HTTP/2 to nginx docs

### DIFF
--- a/nginx/content.md
+++ b/nginx/content.md
@@ -117,3 +117,7 @@ With Amplify it is possible to collect and aggregate metrics across containers, 
 In order to use Amplify, a small Python-based agent software (Amplify Agent) should be [installed](https://github.com/nginxinc/docker-nginx-amplify) inside the container.
 
 For more information about Amplify, please check the official documentation [here](https://github.com/nginxinc/nginx-amplify-doc).
+
+## HTTP/2
+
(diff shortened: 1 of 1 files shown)


yosifkit:
Seem fine to me, but I'll leave it up to the maintainer, @thresheek.

thresheek:
I am against that change, since current nginx:latest (since 1.13.0 and 1.12.0) ship with http2 enabled due to usage of debian stretch base image.  This makes this change incorrect.

tianon:
Thanks for reviewing @thresheek :heart:

jareddlc:
@thresheek here is what i've found:

```
$ docker run -it --rm --name some-nginx nginx:latest bash
root@9db74cfc1b6b:/# nginx -V
nginx version: nginx/1.13.7
built by gcc 6.3.0 20170516 (Debian 6.3.0-18)
built with OpenSSL 1.1.0f  25 May 2017
```

and per the nginx docs: (http://nginx.org/en/docs/http/ngx_http_v2_module.html)

```
Note that accepting HTTP/2 connections over TLS requires the “Application-Layer Protocol Negotiation” (ALPN) TLS extension support, which is available only since OpenSSL version 1.0.2. Using the “Next Protocol Negotiation” (NPN) TLS extension for this purpose (available since OpenSSL version 1.0.1) is not guaranteed to work.
```

jareddlc:
@thresheek thanks for catching my mistake

PR #1083, opened 2017-12-04
Documentation for an Open Liberty official image

NottyCode:


open-liberty/content.md
```diff
@@ -0,0 +1,148 @@
...
+	kernel-java8-ibmsfj
+	microProfile1
+	microProfile1-java8-ibm
+	microProfile1-java8-ibmsfj
+	webProfile7
+	webProfile7-java8-ibm
+	webProfile7-java8-ibmsfj
+	javaee7
+	javaee7-java8-ibm
+	javaee7-java8-ibmsfj
```
tianon: None of this section applies to the current image, right? (since it's only got `kernel` tags?)

open-liberty/content.md
```diff
@@ -0,0 +1,148 @@
...
+
+The images are designed to support a number of different usage patterns. The following examples assume that [DefaultServletEngine.zip](https://github.com/WASdev/sample.servlet/releases/download/V1/DefaultServletEngine.zip) has been extracted to `/tmp`.
+
+1.	Each image contains a default server configuration that specifies the corresponding features and exposes ports 9080 and 9443 for HTTP and HTTPS respectively. A .WAR file can therefore be mounted in the `dropins` directory of this server and run. The following example starts a container in the background running a .WAR file from the host file system with the HTTP and HTTPS ports mapped to 80 and 443 respectively.
+
+	```console
+	$ docker run -d -p 80:9080 -p 443:9443 \
+	    -v /tmp/DefaultServletEngine/dropins/Sample1.war:/config/dropins/Sample1.war \
+	    %%IMAGE%%:webProfile7
+	```
```
tianon: Again, this tag isn't in the current image proposal -- is this just a bad copy-paste?

open-liberty/content.md
```diff
@@ -0,0 +1,148 @@
...
+	$ docker run -d -p 80:9080 \
+	  --volumes-from app %%IMAGE%%:webProfile7
+	```
+
+# Providing your own keystore/truststore
+
+When an `open-liberty` image starts, it can generate a Liberty server XML snippet in `/config/configDropins/defaults/keystore.xml` that specifies a `keyStore` stanza with a generated password. This causes Open Liberty to generate a default keystore and truststore with a self-signed certificate when it starts. The `javaee7` image does this automatically, but other images can request this by setting:
+
+```console
+ENV KEYSTORE_REQUIRED "true"
```
tianon: This particular variable name seems to be unique to open-liberty -- any particular reason for it?  In websphere-liberty, this is just done unilaterally, which should be fine, right?

Code from https://github.com/WASdev/ci.docker/blob/9dc5c2e942d20ddc2f913ae0a0dbcc9d174afc00/ga/developer/kernel/docker-server#L22-L33:

```bash
keystorePath="/config/configDropins/defaults/keystore.xml"

if [ ! -e $keystorePath ]
then
  # Generate the keystore.xml
  PASSWORD=$(openssl rand -base64 32)
  XML="<server description=\"Default Server\"><keyStore id=\"defaultKeyStore\" password=\"$PASSWORD\" /></server>"

  # Create the keystore.xml file and place in configDropins
  mkdir -p $(dirname $keystorePath)
  echo $XML > $keystorePath
fi
```

open-liberty/content.md
```diff
@@ -0,0 +1,148 @@
...
+
+```console
+docker run -d -p 80:9080 -p 443:9443 \
+    -v /tmp/open-liberty/classCache:/opt/ol/wlp//output/.classCache app
+```
+
+Or, create a named data volume container that exposes a volume at the location of the shared file:
+
+```console
+docker run -e LICENSE=accept -v /opt/ol/wlp//output/.classCache \
```
tianon: Does Open Liberty still require license acceptance?  (maybe just another bad copy-paste?)

open-liberty/content.md
NottyCode: I'm planning on committing those tags once the kernel ones get accepted. I wasn't sure I could create tags for Dockerfiles that depend on other Dockerfiles in the initial commit for the official images. However I will prepare a commit that removes this until I provide a commit with those tags.

open-liberty/content.md
NottyCode: see above

open-liberty/content.md
NottyCode: You are right, total copy and paste error.

open-liberty/content.md
NottyCode: This is unique to Open Liberty image. There are a few issues with what the websphere-liberty does. The main one is that most people who use the websphere-liberty image create their own Dockerfile with a FROM clause. In our experience people who care about security tend to have their own certificates from some kind of CA. When they provide those in their Dockerfile they will have their own certificates. That means that the step the websphere-liberty driver has to be 'disabled' which requires them to put their SSL configuration in /config/configDropins/defaults/keystore.xml which isn't exactly intuitive. We have tried very hard not to force conventions on our users when naming their configuration files.

Another advantage this has is the creation of this password causes the app server to generate an SSL certificate during startup. We do this because it makes it easier to do SSL out of the box than requiring someone to create a self signed cert and knowing what config to do for a JVM, however it also affects server startup time, so docker run takes longer. If you don't require SSL then that is an overhead you don't want. The intent of making open-liberty differ from websphere-liberty was to allow an environment variable to dictate whether this happens or not. This also feels to me to be more in keeping with the 3rd factor for 12 factor apps.

PR #1087, opened 2017-12-11
Updated documentation of how-to connect to a postgres database

doublebyte1:
- explain how-to connect to a postgres db, by IP/DNS or container name, using the POSTGRES_DB_HOST environment variable.
  - remove reference to legacy docker --link

### DIFF
--- a/geonetwork/variant-postgres.md
+++ b/geonetwork/variant-postgres.md
@@ -25,14 +25,14 @@ If are want to **run postgres on a container**, you can use the container name a
 $ docker network create --driver bridge mynet
 ```
 
-Then if you want to run the official image of postgres, using `$POSTGRES_DB_HOST` as container name, you could launch it like this:
+Then if you want to run the official image of postgres, using `some-postgres` as container name, you could launch it like this:
 
 ```console
-$ docker run --name $POSTGRES_DB_HOST --network=mynet -d postgres
+$ docker run --name some-postgres --network=mynet -d postgres
 ```
 
 And then you could launch geonetwork, making sure you join the same network, and setting the required environment variables, including the `POSTGRES_DB_HOST`:
 
 ```console
-$ docker run --name geonetwork -d -p 8080:8080 --network=mynet -e POSTGRES_DB_HOST=$POSTGRES_DB_HOST -e POSTGRES_DB_PORT=5432 -e POSTGRES_DB_USERNAME=postgres -e POSTGRES_DB_PASSWORD=mysecretpassword geonetwork:postgres
+$ docker run --name geonetwork -d -p 8080:8080 --network=mynet -e POSTGRES_DB_HOST=some-postgres -e POSTGRES_DB_PORT=5432 -e POSTGRES_DB_USERNAME=postgres -e POSTGRES_DB_PASSWORD=mysecretpassword geonetwork:postgres
 ```
--- a/geonetwork/variant-postgres.md
+++ b/geonetwork/variant-postgres.md
@@ -10,15 +10,16 @@ If your postgres instance is listening on a non-standard port, you must also set
 
 ### Connecting to a postgres database
 
-If you want to connect to a postgres server, you need to pass an extra environment variable, containing the IP address for this server (which could be localhost, if you are running it locally). - `POSTGRES_DB_HOST`: IP address of your database server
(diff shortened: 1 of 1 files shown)


geonetwork/variant-postgres.md
```diff
@@ -8,28 +8,30 @@ In order to setup the connection from geonetwork, you **must** inject the follow
 
 If your postgres instance is listening on a non-standard port, you must also set that variable: - `POSTGRES_DB_PORT`: postgres port on your database server (defaults to `5432`)
 
-### Linking to a postgres container
+### Connecting to a postgres database
 
-Linking to a postgres container, is pretty straightforward: - `--link <some-postgres-container>:postgres`
+If you want to connect to a postgres server, you need to pass an extra environment variable, containing the IP address for this server (which could be localhost, if you are running it locally). - `POSTGRES_DB_HOST`: IP address of your database server
```
yosifkit: I believe `localhost` would be a bad option to use since that would mean `localhost` relative to the geonetwork container and thus not connect to any postgres instance.

geonetwork/variant-postgres.md
```diff
@@ -8,28 +8,30 @@ In order to setup the connection from geonetwork, you **must** inject the follow
...
-And then you could launch geonetwork, linking to this container, and setting the required environment variables:
+If you want to use the container name as `POSTGRES_DB_HOST`, just make sure that containers can discover each other, by **running them in the same user-defined network**. For instance, you can create a bridge network:
 
 ```console
-$ docker run --name geonetwork -d -p 8080:8080 --link some-postgres:postgres -e POSTGRES_DB_USERNAME=postgres -e POSTGRES_DB_PASSWORD=mysecretpassword geonetwork:postgres
+$ docker network create --driver bridge mynet
 ```
 
-### Connecting to a postgres instance
+Then if you want to run the official image of postgres, using `$POSTGRES_DB_HOST` as container name, you could launch it like this:
```
yosifkit: Rather than using a variable as the postgres container name, I think just picking a name would make it easier since a user will probably copy/paste these lines and not set the variable in the first place.  We usually pick something like `some-postgres` or similar variation.

geonetwork/variant-postgres.md
yosifkit: Perhaps you meant the something running directly on the docker host? That would need to be explicitly pointed out to use the IP address of the docker host (or maybe the OSX or Windows host IP address outside the VM).  It gets complicated fairly quickly and is easier to focus on other containers rather than getting into the networking of everything outside the containers.

doublebyte1: @yosifkit I agree with you that using "localhost" for networking is "asking for trouble"... my idea was only to tell users that they can connect to an existing database server somewhere (which does not necessarily needs to run on docker), not to explain them how to run a postgres instance locally and then connect to it using its ip address. I [updated the documentation](https://github.com/GeoCat/docs/commit/8e9fed3c072b58a580c8a070fe975203e44f3a93) to reflect this. Now I present only two simple use cases: 1) connect to an external db using DNS 2) connect to a container running on the same docker host, using the container name.

geonetwork/variant-postgres.md
doublebyte1: @yosifkit Updated [here](https://github.com/GeoCat/docs/commit/618bbf942e420fc399bf871a7d7681c8416811c3).

geonetwork/variant-postgres.md
```diff
@@ -8,28 +8,31 @@ In order to setup the connection from geonetwork, you **must** inject the follow
 
 If your postgres instance is listening on a non-standard port, you must also set that variable: - `POSTGRES_DB_PORT`: postgres port on your database server (defaults to `5432`)
 
-### Linking to a postgres container
+### Connecting to a postgres database
 
-Linking to a postgres container, is pretty straightforward: - `--link <some-postgres-container>:postgres`
+If you want to connect to a postgres server, you need to pass an extra environment variable, `POSTGRES_DB_HOST`,
+containing the address of this server.
```
yosifkit: The rest of the content seems fine to me, just need to combine these two lines (13-14) to make markdownfmt and the TravisCI test happy.

doublebyte1: [Done](https://github.com/GeoCat/docs/commit/9013c0dddc6ba169c47b1da1104f644abe15e7f1)!

PR #1088, opened 2017-12-11
change docs to use --node not --name for starting consul

drhenner:
I needed to use --node to start consul.

### DIFF
--- a/consul/README.md
+++ b/consul/README.md
@@ -86,7 +86,7 @@ The entry point also includes a small utility to look up a client or bind addres
 ## Running Consul for Development
 
 ```console
-$ docker run -d --name=dev-consul -e CONSUL_BIND_INTERFACE=eth0 consul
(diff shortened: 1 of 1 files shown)


yosifkit:
1. the change would have to be in `content.md` and not `README.md` since it is generated.
1. `docker run` does not have a `node` option; is `node` supposed to be an option to `consul` itself?

   ```console
   $ docker run --help | grep -- "--n"
      --name string                    Assign a name to the container
      --network string                 Connect a container to a network (default
      --network-alias list             Add network-scoped alias for the container
      --no-healthcheck                 Disable any container-specified HEALTHCHECK
   ```

tianon:
Also, cc @slackpad for maintainer input :+1:

slackpad:
I think in this case it was intended to be `--name` in order to set the name for the `-t dev-consul` in the next step of the example. @drhenner I think in your case you likely passed that as an argument to Consul itself?

PR #1090, opened 2017-12-12
Add CentOS license link

tianon:
cc @jperrin -- does this seem like a sane place to link for this?

For context, it will also have https://github.com/docker-library/docs/blob/a857080d495c0a0a69299733a6343ca12b77ba45/.template-helpers/license-common.md appended to the end.

Refs #1046

### DIFF
--- a/centos/license.md
+++ b/centos/license.md
@@ -1,0 +1,1 @@
+View [license information](https://www.centos.org/legal/) for the software contained in this image.


jperrin:
yup. Seems fine. For some reason that original bug was buried in email so I didn't notice it. Sorry

tianon:
No worries, thanks for confirming! :+1: :heart:

PR #1015, opened 2017-09-21
Update CouchDB documentation

wohali:
This updates the CouchDB documentation in affiliation with docker-library/official-images#3474.

PR #1103, opened 2017-12-21
couchdb: Update volume paths

bryanburgers:
In the switch from CouchDB 1.7 to CouchDB 2, the location of the data
volume changed.

Update the documentation to reflect the new location.

---

*Note: I don't have `bashbrew` so I didn't run `update.sh`. That will still need to be done.*


bryanburgers:
Oops. #1015 fixes this.

PR #1106, opened 2017-12-29
Point to a couple Official Images "wait-for-db" examples

tianon:
See https://github.com/docker-library/mysql/issues/81 for more discussion around this topic.


tianon:
Do you think it's also worth linking to https://docs.docker.com/compose/startup-order/, which has a very similar example?

tianon:
IMO these are sufficient -- that page is pretty easy to find via a quick Google search.

PR #1112, opened 2018-01-04
Tells to use the `i386/convertigo` image for the Web Connector.

nicolas-albert:
In relation to https://github.com/docker-library/official-images/pull/3862


tianon:
Ah, this is a nice solution! :+1:

PR #1121, opened 2018-01-14
Update Joomla stack/compose example

BretFisher:
The Joomla README example of a stack.yml file could use updating to remove legacy/confusing options:

1. Links are no longer needed as of compose v2. Since this example uses v3.1, links should be removed.
2. restart: option is only used for docker-compose, while restart_policy: is only used for stack deploy. Since it's an option setting, I'd recommend removing it from the example to avoid confusion like this question on SO.

### DIFF
--- a/joomla/README.md
+++ b/joomla/README.md
@@ -97,9 +97,6 @@ version: '3.1'
 services:
   joomla:
     image: joomla
-    restart: always
-    links:
-      - joomladb:mysql
     ports:
       - 8080:80
     environment:
@@ -108,7 +105,6 @@ services:
 
   joomladb:
     image: mysql:5.6
-    restart: always
     environment:
       MYSQL_ROOT_PASSWORD: example
 ```


yosifkit:
Can you move this to [stack.yml](https://github.com/docker-library/docs/blob/7aa24f1b085a696bd7820286934f4a0edd817fd1/joomla/stack.yml)?  `README.md` is autogenerated.

As for the content of the change, I think keeping `restart` on jooma is nice since for some users the first start of MySQL takes longer than [php startup script](https://github.com/joomla/docker-joomla/blob/9cc09b86340bd4eabce331807e6f9ec5adbfcd1a/docker-entrypoint.sh#L58) is willing to [wait](https://github.com/joomla/docker-joomla/blob/9cc09b86340bd4eabce331807e6f9ec5adbfcd1a/makedb.php#L16-L34) (ie, about 30 seconds).

cc @tianon

tianon:
Almost all our "stack.yml" files are intentionally written to work roughly
the same via both "docker-compose" and "docker stack deploy" as much as
they can, which is why we include things like "restart: always" (which is
effectively the default in swarm mode).

BretFisher:
OK I'll move, sorry about that, and keep restart... so this will just be removing links :)

tianon:
:bow:

I'm going to close this now -- if you believe it's still relevant, feel free to open a new PR. :heart:

PR #1126, opened 2018-01-19
Add teamspeak docs

muenchow:


teamspeak/content.md
```diff
@@ -0,0 +1,131 @@
...
+TeamSpeak 3 offers the ideal voice communication for online gaming, education and training, internal business communication, and staying in touch with friends and family. Our primary focus is delivering a solution that is easy to use, with high security standards, excellent voice quality, and low system and bandwidth usage.
+
+> [teamspeak.com](https://teamspeak.com/)
+
+%%LOGO%%
+
+# How to use this image
+
+```console
+$ docker run -p 9987:9987/udp -p 10011:10011 -p 30033:30033 -e TS3SERVER_LICENSE=accept %%IMAGE%%
```
tianon: Do you think it'd be good to encourage users to do `TS3SERVER_LICENSE=view` before they accept it? (just to cover our bases?)

ie, something like:

> # How to use this image
> 
> To view the license:
> 
> ```console
> $ docker run --rm -e TS3SERVER_LICENSE=view %%IMAGE%%
> ```
> 
> To accept the license and run the server (mapping the appropriate ports to the host):
> 
> ```console
> $ docker run -p 9987:9987/udp -p 10011:10011 -p 30033:30033 -e TS3SERVER_LICENSE=accept %%IMAGE%%
> ```

teamspeak/content.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+Then you can connect to `localhost` in your TeamSpeak3 client. Please write down the server query password, and server admin privilege key that were generated. There are needed to administrate the TeamSpeak 3 server.
```
tianon: "There are needed to administrate ..." was probably intended to be "_These_ are needed to administrate ..."

teamspeak/content.md
```diff
@@ -0,0 +1,131 @@
...
+
+# Accept License
+
+In order to use the TeamSpeak server you must accept the license first. You can view the license using:
+
+```console
+$ docker run -e TS3SERVER_LICENSE=view %%IMAGE%%
+```
+
+The license can be accepted by setting `TS3SERVER_LICENSE` to `accept`.
```
tianon: Ah, notes about the license are down here -- IMO this should really be called out _before_ we tell folks how to run the image (and thus accept the license), but your call.

teamspeak/content.md
```diff
@@ -0,0 +1,131 @@
...
+The Docker documentation is a good starting point for understanding the different storage options and variations, and there are multiple blogs and forum postings that discuss and give advice in this area. We will simply show the basic procedure here for the latter option above:
+
+1.	Create a data directory on a suitable volume on your host system, e.g. `/my/own/datadir`.
+2.	Start your `%%REPO%%` container like this:
+
+```console
+$ docker run --name some-%%REPO%% -v /my/own/datadir:/var/ts3server/ -d %%IMAGE%%:tag
+```
+
+The `-v /my/own/datadir:/var/ts3server/` part of the command mounts the `/my/own/datadir` directory from the underlying host system as `/var/ts3server` inside the container, where Cassandra by default will write its data files.
```
tianon: Cassandra? :smirk:

muenchow:
@tianon thank you for the review. I very much agree with the suggested changes and made appropriate changes.

teamspeak/stack.yml
```diff
@@ -1,24 +1,24 @@
...
-            TS3SERVER_DB_WAITUNTILREADY: 30
-            TS3SERVER_LICENSE: accept
-    db:
-        image: mariadb
-        restart: always
-        environment:
-            MYSQL_ROOT_PASSWORD: example
-            MYSQL_DATABASE: teamspeak
+  teamspeak:
+    image: teamspeak-server
```
tianon: Whoops, missed one!  This should be just `teamspeak`, right? (I can fix this post-merge if we merge before you get to it)

teamspeak/stack.yml
```diff
@@ -1,24 +1,24 @@
...
+      TS3SERVER_DB_PLUGIN: ts3db_mariadb
+      TS3SERVER_DB_SQLCREATEPATH: create_mariadb
+      TS3SERVER_DB_HOST: db
+      TS3SERVER_DB_USER: root
+      TS3SERVER_DB_PASSWORD: example
+      TS3SERVER_DB_NAME: teamspeak
+      TS3SERVER_DB_WAITUNTILREADY: 30
+      TS3SERVER_LICENSE: accept
+  db:
+    image: mariadb
```
tianon: Did you want to pin this to a particular release channel (such as `mariadb:10.2`), or is `:latest` expected to always work?

tianon: (this can be a follow-up PR if it needs to change)

PR #1128, opened 2018-01-25
Remove all reference to deprecated "go-wrapper" script

tianon:
This is a companion to https://github.com/docker-library/golang/pull/205.

### DIFF
--- a/golang/content.md
+++ b/golang/content.md
@@ -18,10 +18,10 @@ FROM %%IMAGE%%:1.8
 WORKDIR /go/src/app
 COPY . .
 
-RUN go-wrapper download   # "go get -d -v ./..."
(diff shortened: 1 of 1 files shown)


jstangroome:
Updating my existing Dockerfiles to use `FROM golang:1.10` instead of `FROM golang:1.9` and to use `RUN go install -v ./...`  instead of `RUN go-wrapper install` as per this change to the docs, now breaks the `docker build` with:

```
can't load package: package app: code in directory /go/src/app expects import "github.com/jstangroome/what-how"
```

See gist reproducing issue with both previously working and now broken Dockerfiles:
https://gist.github.com/jstangroome/93742d47d8b4965877a924fc99f60c9e

PR #1129, opened 2018-01-25
Elaboration on POSTGRES_PASSWORD

qedrakmar:
Added a note to explain that POSTGRES_PASSWORD is not the same as PGPASSWORD.  Closes https://github.com/docker-library/postgres/issues/402

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -55,7 +55,9 @@ The PostgreSQL image uses several environment variables which are easy to miss. 
 
 This environment variable is recommended for you to use the PostgreSQL image. This environment variable sets the superuser password for PostgreSQL. The default superuser is defined by the `POSTGRES_USER` environment variable. In the above example, it is being set to "mysecretpassword".
 
-Note: The PostgreSQL image sets up `trust` authentication locally so you may notice a password is not required when connecting from `localhost` (inside the same container). However, a password will be required if connecting from a different host/container.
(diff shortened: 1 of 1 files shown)


postgres/content.md
```diff
@@ -55,7 +55,9 @@ The PostgreSQL image uses several environment variables which are easy to miss.
 
 This environment variable is recommended for you to use the PostgreSQL image. This environment variable sets the superuser password for PostgreSQL. The default superuser is defined by the `POSTGRES_USER` environment variable. In the above example, it is being set to "mysecretpassword".
 
-Note: The PostgreSQL image sets up `trust` authentication locally so you may notice a password is not required when connecting from `localhost` (inside the same container). However, a password will be required if connecting from a different host/container.
+Note 1: The PostgreSQL image sets up `trust` authentication locally so you may notice a password is not required when connecting from `localhost` (inside the same container). However, a password will be required if connecting from a different host/container.
+
+Note 2: This variable sets the superuser password in the PostgreSQL instance, and has no effect on the PGPASSWORD variable that may be used by the psql client. PGPASSWORD, if used, will be specified as a separate environment variable.
```
tianon: I like the addition, but I think the wording ought to make it a little more clear what makes these different (and thus why we even "Note" this in the first place).  How about something like this?

> Note 2: This variable controls the superuser password set during `initdb` at initial container startup, and has no effect on `psql` (which reads from `PGPASSWORD` instead; see https://www.postgresql.org/docs/10/static/libpq-envars.html).

PR #1130, opened 2018-01-26
MySQL and MariaDB: Update --link in examples

ajcrites:
Docker's recommendation is to avoid using `--link`. I have updated the examples
in the docs for these two containers to use user-defined networks instead. I
think that this is actually simpler than using `--link` as well.

Changes:

278d4838 (Andrew Crites, 29 seconds ago)
   mariadb+mysql: update examples to use networks over --link

22e85132 (Andrew Crites, 2 hours ago)
   mysql: update examples for user-defined networks

   The --link method suggested is legacy and is no longer recommended by
   Docker.

   I have updated the examples to use user-defined networks instead.

### DIFF
--- a/mariadb/content.md
+++ b/mariadb/content.md
@@ -24,18 +24,34 @@ $ docker run --name some-%%REPO%% -e MYSQL_ROOT_PASSWORD=my-secret-pw -d %%IMAGE
 
 Since MariaDB is intended as a drop-in replacement for MySQL, it can be used with many applications.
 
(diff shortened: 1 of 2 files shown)


tianon:
The discussions in https://github.com/docker-library/docs/issues/786#issuecomment-350491848 and https://github.com/docker-library/docs/pull/1063 are relevant here.

tianon:
See also https://github.com/docker-library/faq#--link-is-deprecated. :+1:

PR #1131, opened 2018-01-28
added note about Java 10 support

keeganwitt:
removed note about Java 9 support

### DIFF
--- a/gradle/content.md
+++ b/gradle/content.md
@@ -6,10 +6,13 @@
 
 # How to use this image
 
-Note that if you are mounting a volume and the uid running Docker is not `1000`, you should run as user `root` (`-u root`).
-
 ## Building a Gradle project
 
 Run this from the directory of the Gradle project you want to build.
 
-`docker run --rm -v "$PWD":/home/gradle/project -w /home/gradle/project %%IMAGE%% gradle <gradle-task>`
+`docker run --rm -u gradle -v "$PWD":/home/gradle/project -w /home/gradle/project %%IMAGE%% gradle <gradle-task>`
+
+Note the above command runs using uid/gid 1000 (user _gradle_) to avoid running as root.
+
+If you are mounting a volume and the uid/gid running Docker is not _1000_, you should run as user _root_ (`-u root`).
+_root_ is also the default, so you can also simply not specify a user.


gradle/content.md
```diff
@@ -14,4 +14,4 @@ Run this from the directory of the Gradle project you want to build.
 
 `docker run --rm -v "$PWD":/home/gradle/project -w /home/gradle/project %%IMAGE%% gradle <gradle-task>`
 
-**Note: Java 9 support is experimental**
+**Note: Java 10 support is experimental**
```
tianon: OpenJDK 9 is still based on Debian Unstable as well -- you sure you're comfortable updating this support level?  (We don't mind either way, just trying to look out for your sanity. :smile:)

keeganwitt: My intent was to give folks something to play with there, not really supported as-such. But I probably should have asked -- does it being an official image imply at least moderate support? I'd say Java 9 is in a somewhat supported state, but I'm not sure anybody has started looking at 10 yet.

tianon: It's somewhat supported, but the `openjdk` image is a bit of a special case because we're at the behest of the upstream distributions for providing packages (as you've experienced, especially around SSL breakage).

The fact that 9 and 10 are both on Sid currently really bothers me, and makes me worry for their stability:

```console
$ bashbrew from --uniq openjdk:8-jdk openjdk:9-jdk openjdk:10-jdk
openjdk:8u151-jdk-stretch: buildpack-deps:stretch-scm
openjdk:9.0.1-11-jdk-sid: buildpack-deps:sid-scm
openjdk:10-ea-32-jdk-experimental: buildpack-deps:sid-scm
```

keeganwitt: I meant do you think it was inappropriate for me to add a Java 10 image if it's not at all supported by the upstream team?  I wondered if official images were meant to imply support -- because that wasn't my intent here.

TheHydGuy: Hello @keeganwitt,

Do we have support for Java 10 and Jetty 9.4 yet? I'm wondering if I can use openjdk-10 for my container.

Thank you!

tianon: Honestly at this point, `openjdk:11-oracle` is the _only_ version I'd be willing to say is "well-supported" (see https://github.com/docker-library/openjdk/pull/235 for details about that) given that it's actually using binaries supported and created by the OpenJDK project directly, not beholden to the constant-issue-laden packages we get from the distributions.

PR #1144, opened 2018-02-06
Add express-gateway docs

XVincentX:
This PR will add the required documentation to make Express Gateway land in the official program images.


XVincentX:
@yosifkit @tianon 

Is there any way I can speed up this and make it happen?

express-gateway/README-short.txt
```diff
@@ -0,0 +1 @@
+The official release of Express-Gateway.
```
tianon: This is the text that will show up when folks do `docker search express-gateway` and on https://hub.docker.com/search?q=express-gateway&type=image -- this description is fine, but it doesn't really tell folks anything _about_ express gateway (which is OK by us, but you might want something better so folks know what it is and might try it out).

express-gateway/content.md
```diff
@@ -0,0 +1,61 @@
...
+
+_Note:_ You might want to expose other ports to the host in case you're serving your APIs through **HTTPS**.
+_Note:_ You need to mount a volume with configuration files and volumes in order to make Express-Gateway start correctly.
+
+You can now read the docs at [express-gateway.io/docs](http://express-gateway.io/docs) to learn more about Express-Gateway and configure it accordingly to your needs.
+
+### Install plugin
+
+You can install custom plugins to the current Express Gateway image just creating a new `Dockerfile`,
+use `expressgateway/express-gateway` as base image and then install the required plugins as global yarn
```
tianon: `expressgateway/express-gateway` here should instead be `%%IMAGE%%` so it gets replaced with the correct image name in each architecture-specific image description too.

express-gateway/content.md
```diff
@@ -0,0 +1,61 @@
...
+packages
+
+`RUN yarn add global express-gateway-plugin-name`
```
tianon: This should be a full code block, ala:

```markdown
```dockerfile
FROM %%IMAGE%%
RUN yarn add global express-gateway-plugin-name
...
```

express-gateway/content.md
```diff
@@ -0,0 +1,61 @@
...
+If so, skip directly to the point **2**; else, please keep going with this guide.
+
+### 1. Link Express-Gateway to a Redis container
+
+#### Start Redis
+
+Start a Redis container by executing:
+
+```shell
+docker run -d --name express-gateway-data-store \
```
tianon: This should have a `$` prefix like the following one (for consistency).

XVincentX:
@tianon I addressed all the comments — however the markdown format checking is failing but to be honest I do not really understand its reason. Can you give me a hint?

yosifkit:
Looks like a couple things:
 - list items need to be formatted with tabs 
    ```markdown
    -	item
    ```
 - italics needs to be star a `*` rather than underscore `_` (bold should be two star)
 - lines next to each other need to either have a empty line between them to be considered different paragraphs, or be one line to be a single paragraph
   - i.e., paragraphs need to be a single line and have a single empty line between paragraphs

XVincentX:
Oh I understood only now that the error message was instead a diff that was trying to help me!

I should have fixed all the issues. @yosifkit @tianon

XVincentX:
This is so great! Thanks a lot!

PR #1145, opened 2018-02-06
[Nextcloud] Add play with docker button

tilosp:
Same as #1034 with less hacky cron container


tilosp:
@tianon please review

yosifkit:
We are bad a documenting current features here in the docs, but could you do it more like #913, #933, and #941 so that the yaml will also be embedded in the description so users can see what is being used in the PWD stack deploy?

PR #1147, opened 2018-02-08
Add initial Postfix Admin docs

J0WI:


postfixadmin/content.md
```diff
@@ -0,0 +1,72 @@
...
+           -p 8080:80
+        %%IMAGE%%
+```
+
+Then, access it via `http://localhost:8080` or `http://host-ip:8080` in a browser.
+
+## Existing config.local.php
+
+```console
+$ docker run --name some-%%REPO%% -p 8080:80 %%image%%
```
yosifkit: Should this be showing an example of bind mounting a local existing `config.local.php`?

Also `s/%%image%%/%%IMAGE%%/`

yosifkit:
Also, does there need to be something mentioning how to point this to a postfix server? (since there are no env vars for that and users may get confused)

postfixadmin/content.md
```diff
@@ -0,0 +1,73 @@
...
+$ docker run --name some-%%REPO%% -v /local/path/to/config.local.php:/var/www/html/config.local.php -p 8080:80 %%IMAGE%%
```
tianon: I don't think this code block was intended to have two `docker run` lines in it. :smile:

postfixadmin/content.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Image Variants
```
tianon: This section is going to look really weird next to the section that `update.sh` is going to auto-generate with the same title.  I think you wanted to instead put these blurbs into `variant-apache.md`, etc files.

postfixadmin/stack.yml
```diff
@@ -0,0 +1,25 @@
+version: '3'
+
+services:
+  db:
+    image: mysql:5.7
+    restart: always
+    environment:
+      MYSQL_ROOT_PASSWORD: notSecureChangeMe
```
tianon: You might consider swapping to simply using `MYSQL_RANDOM_ROOT_PASSWORD: 1` since you create an explicit non-`root` user/database (which makes access to the `root` account mostly unnecessary).

postfixadmin/stack.yml
```diff
@@ -0,0 +1,25 @@
...
+      MYSQL_DATABASE: postfixadmin
+      MYSQL_USER: postfixadmin
+      MYSQL_PASSWORD: example
+
+  postfixadmin:
+    depends_on:
+      - db
```
tianon: FYI, `depends_on:` doesn't really do much these days (especially with services in a stack), and I'd recommend against relying on it (or using it).

PR #1152, opened 2018-02-16
Updating postgres notes/example

TigerC10:
The example init script in this documentation fails when the `$POSTGRES_USER` and `$POSTGRES_DB` values mismatch.  By default, `psql` attempts to set the database in use to the same name as the logging in `$POSTGRES_USER`, so without setting the `-d` flag users will encounter an error similar to: psql: FATAL:  database "root" does not exist when they don't match.  Related to docker-library/postgres#41

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -142,7 +142,7 @@ For example, to add an additional user and database, add the following to `/dock
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


TigerC10:
Looks like someone else added it separately. Closing.

PR #1164, opened 2018-03-01
Add docs for Sematext Docker Agent

megastef:
See also pending PR for the image here: https://github.com/docker-library/official-images/pull/4044


megastef:
Does the check fail because docker-library/official-images#4044 is not merged?

PR #1170, opened 2018-03-09
Correct file names in Jetty documentation

adrianmsmith:
The `start.d` files are called `*.ini` (not `*.mod`, as was stated in
the previous documentation)


adrianmsmith:
To confirm that `*.ini` is the correct extension:

```
$ sudo docker run -it jetty sh
$ ls /var/lib/jetty/start.d/
deploy.ini  ext.ini  http.ini  jsp.ini	
jstl.ini  resources.ini  server.ini  websocket.ini
```

tianon:
cc @gregw @md5

PR #1171, opened 2018-03-09
Correct wrong parameter name in Jetty docs

adrianmsmith:
To run Jetty with particular modules, use the `--module` option; the
previous doc said `--modules`. The Jetty documentation
https://www.eclipse.org/jetty/documentation/9.3.x/startup.html
references the `--module` option many times, but never `--modules`.


tianon:
Sorry for the delay!

cc @gregw @md5

Issue #183, opened 2015-03-13
Support named groups in "Supported tags"


md5:
I was thinking today that it would be interesting to support named groups in the "Supported tags" lists in the `README.md` files.

For example, given a library file like this:

```
# maintainer: InfoSiftr <github@infosiftr.com> (@infosiftr)

# group: PHP 5.6
5.6.6-cli: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6
5.6-cli: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6
5-cli: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6
cli: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6
5.6.6: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6
5.6: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6
5: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6
latest: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6

5.6.6-apache: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6/apache
5.6-apache: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6/apache
5-apache: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6/apache
apache: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6/apache

5.6.6-fpm: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6/fpm
5.6-fpm: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6/fpm
5-fpm: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6/fpm
fpm: git://github.com/docker-library/php@1e07e9134490f53d24fbd23ee3521067c8ab2d29 5.6/fpm

# group: PHP 5.5
[...]

PR #1181, opened 2018-03-25
Readability Improvements

chamunks:
Just broke up the kludge of tags with some contextual titles.


tianon:
This entire section is autogenerated -- see https://github.com/docker-library/docs/issues/183 for a relevant discussion around "tag grouping" (which is essentially what you're accomplishing here, and PHP is the very example being discussed there :+1:).

PR #1184, opened 2018-03-30
fix(volume): add missing port bindings

laurentleseigneur:
example with volume missed the `-p` arg in order to browse bonita portal


yosifkit:
`README.md` files are autogenerated, so it is unnecessary to modify them.  Please update `content.md` and docker-library-bot will take care of `README.md` after merge.

laurentleseigneur:
thanks, i will open a new PR for that

PR #1197, opened 2018-04-17
[nats][nats-streaming] Update logos

kozlovic:
Updating logos for NATS and NATS Streaming with the new color
scheme logo.


kozlovic:
@yosifkit Let me know if I should create 2 separate PRs or if it's fine to have 1 PR for the 2 libraries.

yosifkit:
The one PR is fine. LGTM

PR #1200, opened 2018-04-18
Fix #1199,PECL extensions version number is not required.

### LINKED ISSUE #1199: PECL extensions version number is not required.
```
$ docker build .
Sending build context to Docker daemon  6.144kB
Step 1/8 : FROM php:7.1-fpm
 ---> acc15556e165
Step 2/8 : RUN pecl install redis-3.1.0 	&& pecl install xdebug-2.5.0 	&& docker-php-ext-enable redis xdebug
 ---> Running in 4b78f6800eba
pecl/redis requires PHP (version >= 5.2.0, version <= 7.1.0), installed version is 7.1.16
No valid packages found
install failed
The command '/bin/sh -c pecl install redis-3.1.0 	&& pecl install xdebug-2.5.0 	&& docker-php-ext-enable redis xdebug' returned a non-zero code: 1
```

dravenk:
Please reviews.

### DIFF
--- a/php/README.md
+++ b/php/README.md
@@ -191,8 +191,8 @@ Some extensions are not provided with the PHP source, but are instead available 
 
 ```dockerfile
 FROM php:7.1-fpm
-RUN pecl install redis-3.1.0 \
-	&& pecl install xdebug-2.5.0 \
+RUN pecl install redis \
+	&& pecl install xdebug \
 	&& docker-php-ext-enable redis xdebug
 ```
 


yosifkit:
Thanks! Can you move this to `content.md`?  `README.md` is a generated file.

tianon:
I'm -1 on this change -- we use specific versions specifically _because_ of the error seen in #1199.

PCEL itself does not use the PHP version to determine which version of the requested extension to install, so it will happily attempt to install the latest version and then fail because the PHP version requirements aren't met.

tianon:
If this `Dockerfile` doesn't build as-is, we need to update the version numbers to ones which _do_ support PHP 7.1 properly.

PR #1203, opened 2018-04-20
Fix authentication plugin

GitHub30:
Authentication plugin 'caching_sha2_password' cannot be loaded

### DIFF
--- a/mysql/stack.yml
+++ b/mysql/stack.yml
@@ -5,6 +5,7 @@ services:
 
   db:
     image: mysql
+    command: mysqld --default-authentication-plugin=mysql_native_password
     restart: always
     environment:
       MYSQL_ROOT_PASSWORD: example


yosifkit:
This looks more like an issue with adminer (or even php) needing an update to support the [new default authentication method](https://mysqlserverteam.com/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/) since `8.0` is now `latest`.

yosifkit:
Here it is: https://bugs.php.net/bug.php?id=76243

yosifkit:
Ok, I think that the PHP issue may take too long to leave this broken. Given that this stack is for mysql, not php and that MySQL's recommended fix for PHP clients is to downgrade the authentication, I am :+1: on this.

>For example, in PHP, MySQL connectivity usually is based on `mysqlnd`, which currently does not know about `caching_sha2_password`. Until an updated version of `mysqlnd` is available, the way to enable PHP clients to connect to MySQL 8.0 is to reconfigure the server to revert to `mysql_native_password` as the default authentication plugin, as previously discussed.

Could we add a comment in the stack file pointing to the php bug so we know when to remove it? (also drop `mysqld` since it is not necessary)

```yaml
  db:
    image: mysql
    # adminer cannot use the new auth plugin, so revert to the old one
    # https://bugs.php.net/bug.php?id=76243
    command: --default-authentication-plugin=mysql_native_password
...
```

What do you think @tianon?

tianon:
Sounds fair. :+1:

GitHub30:
Thank you

PR #1204, opened 2018-04-20
Add a blurb about installing PostgreSQL extensions

tianon:
See https://github.com/docker-library/postgres/issues/340.


tianon:
(props to @md5 :heart:)

PR #1205, opened 2018-04-23
Update docs

kraag22:
There is issue with mountig directories on Mac. https://stackoverflow.com/questions/49923502/error-when-running-ghost-docker-image-second-time


tianon:
Sorry for the delay.  IMO, this is a normal (semi-expected) quirk of the shared filesystem used on Mac OS, and thus overkill for us to document (and the proposed change is a bit too thin anyhow -- no links to why this would be necessary).

I think this is already reasonably searchable via the above Stack Overflow.

PR #1206, opened 2018-04-24
Adds Matomo

pierreozoux:
and deprecate piwik


matomo/content.md
```diff
@@ -0,0 +1,60 @@
...
+- Password: MYSQL_ROOT_PASSWORD
+- Database Name: piwik (or you can choose)
+
+And leave the rest as default.
+
+Then you can continue the installation with the super user.
+
+## Docker-composer examples and log import instructions
+
+A minimal set-up using docker-compose is available in the [.examples folder](.examples/docker-compose.yml), a more complete [example can be found at IndieHosters/piwik](https://github.com/libresh/compose-matomo/blob/master/docker-compose.yml).
```
yosifkit: broken link: `.examples/docker-compose.yml`.  Perhaps you could use `%%GITREPO%%/blob/master/.examples/docker-compose.yml`.

It looks like that compose file could use an update to `matomo` on the image name as well as the IndieHosters link text

matomo/content.md
```diff
@@ -0,0 +1,60 @@
...
+- Database Server: `db`
+- Login: `root`
+- Password: MYSQL_ROOT_PASSWORD
+- Database Name: piwik (or you can choose)
+
+And leave the rest as default.
+
+Then you can continue the installation with the super user.
+
+## Docker-composer examples and log import instructions
```
yosifkit: `s/composer/compose/`

pierreozoux:
@yosifkit thanks for feedback, I addressed your comments!

pierreozoux:
Ok, fixed all the errors, there is one remaining, but I don't get it:
There is no README in the PR :/

Thanks for helping :)

PR #1208, opened 2018-05-02
Link Deprecation

Dargmuesli:
+ Note about --link parameter deprecation (section can be completely removed in the future)
~ Moved stack description above deprecated parts


Dargmuesli:
For consideration in parallel to #786.

PR #1211, opened 2018-05-03
Provides information around handling kernels with vsyscall disabled

sgowie:
Related to https://github.com/CentOS/sig-cloud-instance-images/issues/103

@jperrin Would you consider this a reasonable addition?

### DIFF
--- a/centos/content.md
+++ b/centos/content.md
@@ -85,7 +85,7 @@ This container is running with systemd in a limited context, with the cgroups fi
 
 ## A note about vsyscall
 
-CentOS 6 binaries and/or libraries are built to expect some system calls to be accessed via `vsyscall` mappings.  Some linux distributions have opted to disable `vsyscall` entirely (opting exclusively for more secure `vdso` mappings), causing segmentation faults.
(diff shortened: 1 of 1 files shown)


centos/content.md
```diff
@@ -82,3 +82,26 @@ $ docker run -ti -v /sys/fs/cgroup:/sys/fs/cgroup:ro -p 80:80 local/c7-systemd-h
 ```
 
 This container is running with systemd in a limited context, with the cgroups filesystem mounted. There have been reports that if you're using an Ubuntu host, you will need to add `-v /tmp/$(mktemp -d):/run` in addition to the cgroups mount.
+
+## A note about vsyscall
+
+Legacy CentOS binaries and/or libraries are built to expect some system calls to be accessed via `vsyscall` mappings.  Some linux distributions have opted to disable `vsyscall` entirely (opting exclusively for more secure `vdso` mappings), causing segmentation faults.
```
jperrin: Not a fan of the word "Legacy" here. We could say something to the effect of "CentOS 6 binaries..."

Other than that I'm good with the content.

sgowie: Apologies.  For some reason I was under the impression there were el5 generation images available.

tianon:
Thanks folks! :+1:

PR #1224, opened 2018-05-14
Add deprecated.md notice for opensuse

Vogtinator:


opensuse/deprecated.md
```diff
@@ -0,0 +1,9 @@
+This image has been deprecated in favor of the [`opensuse/leap`](https://hub.docker.com/r/opensuse/leap)
+and [`opensuse/tumbleweed`](https://hub.docker.com/r/opensuse/tumbleweed) images provided and maintained
+by the [openSUSE Project](https://www.opensuse.org/) release team.
+
+`opensuse:42.3` (= `opensuse:latest`) will receive security relevant fixes until the [EOL of Leap 42.3](https://en.opensuse.org/Lifetime#openSUSE_Leap).
+Newer openSUSE Leap releases such as 15.x are only available at [`opensuse/leap`](https://hub.docker.com/r/opensuse/leap).
```
yosifkit: Lines 5 and 6 (as well as 8 and 9) either need to be combined to a single line or need a empty line between them. Which is why the CI is failing.

Vogtinator: Indeed. I didn't get a notification that the CI failed at all, so thanks for the comment. I'll fix it.

PR #1243, opened 2018-06-01
Add hint to install PECL extensions in series

J0WI:


php/content.md
```diff
@@ -140,7 +140,7 @@ RUN apt-get update && apt-get install -y libmemcached-dev zlib1g-dev \
 	&& docker-php-ext-enable memcached
 ```
 
-It is *strongly* recommended that users use an explicit version number in their `pecl install` invocations to ensure proper PHP version compatibility (PECL does not check the PHP version compatiblity when choosing a version of the extension to install, but does when trying to install it).
+It is *strongly* recommended that users use an explicit version number in their `pecl install` invocations to ensure proper PHP version compatibility (PECL does not check the PHP version compatiblity when choosing a version of the extension to install, but does when trying to install it). Unlike PHP core extensions, PECL extensions should be installed in series to fail properly if something went wrong. Otherwise errors are just skipped by PECL.
```
tianon: This sounds great, but I think it deserves to be in its own paragraph (with an example to illustrate), especially since the paragraph after this one is further expounding on the previous portion of this paragraph, so this addition here makes it sound odd.

I think it'd be good to move this down below the "Beyond the compatibility issue, ..." paragraph, and add something simple like "for example, `pecl install memcached-2.2.0 && pecl install redis-2.2.8` instead of `pecl install memcached-2.2.0 redis-2.2.8`".

PR #1247, opened 2018-06-06
Update postgres initdb information about sourcing scripts vs executing then

yosifkit:
Docs for https://github.com/docker-library/postgres/pull/452


tianon:
Change merged in https://github.com/docker-library/official-images/pull/4443. :+1:

PR #1249, opened 2018-06-07
Update examples to use PHP 7.2 instead of 7.0

jasonmccallister:
Since PHP 7.0 is no longer supported (security only), updated the documentation examples to 7.2.

http://php.net/supported-versions.php

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -15,7 +15,7 @@ For PHP projects run through the command line interface (CLI), you can do the fo
 ### Create a `Dockerfile` in your PHP project
 
 ```dockerfile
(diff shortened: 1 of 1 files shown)


mccallisterio:
build failed: 

```
Error: at least one repo README.md has changed
These files are autogenerated, so it is unnecessary to modify them
Please update content.md and docker-library-bot will take care of README.md
See: https://github.com/docker-library/docs/#image-namereadmemd
```

Will make the changes in the correct spot when I can.

J0WI:
Try to change it [here](https://github.com/docker-library/docs/blob/master/php/content.md).

PR #1260, opened 2018-06-26
Add instructions to run nginx in read-only mode

alessfg:
There's some use cases where it may be necessary/beneficial to run NGINX in read-only mode. I've added some simple instructions to the docs to address these use cases.


thresheek:
I approve of that.

tianon:
Thanks @thresheek! :+1:

PR #1272, opened 2018-07-16
Update stack.yml to fix adminer connection

lalyos:
adminer has connectivity issues with mysql versions from 8.0.

### DIFF
--- a/mysql/stack.yml
+++ b/mysql/stack.yml
@@ -5,6 +5,7 @@ services:
 
   db:
     image: mysql
+    command: --default-authentication-plugin=mysql_native_password
     restart: always
     environment:
       MYSQL_ROOT_PASSWORD: example


lalyos:
see: https://mysqlserverteam.com/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/

lalyos:
or as an alternative solution, default authentication plugin can be set explicitly:

```
  db:
      image: mysql
 +    command: --default-authentication-plugin=mysql_native_password
      restart: always
      environment:
        MYSQL_ROOT_PASSWORD: example
```
[modified stack.yml](https://github.com/lalyos/docs/commit/d8a0f40a9a15b4ea251e4f4032dec4e87965d8c1)

[try with PWD](https://labs.play-with-docker.com/?stack=https://raw.githubusercontent.com/lalyos/docs/master/mysql/stack.yml)

yosifkit:
Let's got he `command` route since then it will be the newer version of mysql.

lalyos:
@yosifkit changed to the `command` version.

PR #1274, opened 2018-07-19
Add springBoot content to official liberty image readmes

dazavala:


tianon:
Just to thread the needle, this is additional content following https://github.com/docker-library/official-images/pull/4609 (which added these tags; fyi @arthurdm).

It looks like you've got some whitespace issues that our `markdownfmt` isn't happy with.

dazavala:
@tianon Thanks for your earlier comment. Took a while to determine the non-obvious offending whitespace, and I committed the open-liberty update in addition to websphere-liberty. Would you please review these changes?  -Regards.

open-liberty/content.md
```diff
@@ -94,6 +100,68 @@ The images are designed to support a number of different usage patterns. The fol
...
+	$ docker run -d -p 8080:9080 \
+	    -v /tmp/hellospringboot.jar:/config/dropins/spring/hellospringboot.jar \
+	    %%IMAGE%%:springBoot2
+	```
+
+	Similarly, you can create a Spring Boot application layer over this image by adding the application JAR to the `dropins/spring` directory. In this example we copied `hellospringboot.jar` from `/tmp` to the same directory containing the following Dockerfile.
+
+	```dockerfile
+	FROM %%IMAGE%%:springBoot2
+	ADD  hellospringboot.jar /config/dropins/spring/
```
yosifkit: This should be `COPY` instead of `ADD` (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy). `ADD` should only be recommended when extracting a local tar file.

 (should be updated in `websphere-liberty` as well)

dazavala:
Thank you, @yosifkit.  The changes are ready for final approval.  -Regards.

PR #1284, opened 2018-08-13
Add a note about WP-CLI's "USER"

tianon:
Pointing to https://github.com/docker-library/wordpress/issues/256, inspired by https://github.com/docker-library/wordpress/issues/322. :+1:


tofumatt:
You are awesome, thanks so much! 😄

PR #1286, opened 2018-08-23
Implement "archFilter" to fix arch-specific "Shared Tags" lists

tianon:
This implements https://github.com/docker-library/official-images/pull/4764 in order to resolve https://github.com/docker-library/mongo/issues/296.

I've verified that it has the expected behavior for arm64v8 mongo and postgres, and that doing a full `update.sh` across all images generates no delta. :+1:


tianon:
(Best way to review is ignoring whitespace: https://github.com/docker-library/docs/pull/1286/files?w=1)

PR #1288, opened 2018-08-27
Add MySQL SSL note to WordPress description

tianon:
Closes docker-library/wordpress#299


tianon:
Really good points 👍

tianon:
> Perhaps just point to the trac issue and the plugin; and tell users to extract the plugin manually onto their wordpress directory and edit the wp-config to have the appropriate defines?

Text updated.

PR #1289, opened 2018-08-27
Add an "arbitrary user" section to MySQL

tianon:
Closes docker-library/percona#64

This needs to be copied to Percona and MariaDB as well, but there are other changes they need too, so I figured we could start here and then do a separate "resync" PR.

### DIFF
--- a/mysql/content.md
+++ b/mysql/content.md
@@ -176,6 +176,17 @@ If the application you're trying to connect to MySQL does not handle MySQL downt
 
 If you start your `%%IMAGE%%` container instance with a data directory that already contains a database (specifically, a `mysql` subdirectory), the `$MYSQL_ROOT_PASSWORD` variable should be omitted from the run command line; it will in any case be ignored, and the pre-existing database will not be changed in any way.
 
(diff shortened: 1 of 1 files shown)


mysql/content.md
```diff
@@ -176,6 +176,17 @@ If the application you're trying to connect to MySQL does not handle MySQL downt
 
 If you start your `%%IMAGE%%` container instance with a data directory that already contains a database (specifically, a `mysql` subdirectory), the `$MYSQL_ROOT_PASSWORD` variable should be omitted from the run command line; it will in any case be ignored, and the pre-existing database will not be changed in any way.
 
+## Running as an arbitrary user
+
+If you know the permissions of your directory are already set appropriately (such as running against an existing database, as described above) or you have need of running `mysqld` with a specific UID/GID, it is possible to invoke this image with `--user` set to any value in order to achieve the desired access/configuration:
```
yosifkit: We should mention that it needs to be any user other than `root` since root will step down to the mysql user of the image.

tianon: Good point, updated!

PR #1293, opened 2018-08-31
update ghost readme with 1.x => 2.x migration

acburdine:
Several users have run into issues upgrading from Ghost v1 to Ghost v2. Added this section to the Dockerfile to hopefully alleviate such upgrade issues in the future.


acburdine:
cc @tianon

Issue #1295, opened 2018-09-01
3rd approach in postgres "arbitrary user" notes fails with wrong ownership error


darthbinamira:
Instructions from https://github.com/docker-library/docs/tree/master/postgres#arbitrary---user-notes fails with wrong ownership error.

See logs below:

```
$  docker volume create pgdata
pgdata
$  docker run -it --rm -v pgdata:/var/lib/postgresql/data postgres
The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.

The database cluster will be initialized with locale "en_US.utf8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

fixing permissions on existing directory /var/lib/postgresql/data ... ok
creating subdirectories ... ok
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting dynamic shared memory implementation ... posix
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
syncing data to disk ... ok

WARNING: enabling "trust" authentication for local connections
You can change this by editing pg_hba.conf or using the option -A, or
--auth-local and --auth-host, the next time you run initdb.

Success. You can now start the database server using:

    pg_ctl -D /var/lib/postgresql/data -l logfile start

****************************************************
WARNING: No password has been set for the database.
         This will allow anyone with access to the
         Postgres port to access your database. In
         Docker's default configuration, this is
         effectively any other container on the same
         system.

         Use "-e POSTGRES_PASSWORD=password" to set
         it in "docker run".
****************************************************
waiting for server to start....2018-09-01 08:00:20.536 UTC [38] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2018-09-01 08:00:20.544 UTC [39] LOG:  database system was shut down at 2018-09-01 08:00:20 UTC
2018-09-01 08:00:20.552 UTC [38] LOG:  database system is ready to accept connections
 done
server started
ALTER ROLE


/usr/local/bin/docker-entrypoint.sh: ignoring /docker-entrypoint-initdb.d/*

2018-09-01 08:00:20.671 UTC [38] LOG:  received fast shutdown request
waiting for server to shut down...2018-09-01 08:00:20.672 UTC [38] LOG:  aborting any active transactions
2018-09-01 08:00:20.675 UTC [38] LOG:  worker process: logical replication launcher (PID 45) exited with exit code 1
[...]

PR #1296, opened 2018-09-01
Update postgres "arbitrary user" notes

darthbinamira:
This updates the steps outlined in #1295

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -126,8 +126,8 @@ The three easiest ways to get around this:
 	The files belonging to this database system will be owned by user "postgres".
 	...
 	( once it's finished initializing successfully and is waiting for connections, stop it )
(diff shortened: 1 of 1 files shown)


tianon:
I'm on board with switching from the `bash` image to the `postgres` image (which should use the `%%IMAGE%%` placeholder like the other `docker run` lines here do), but I think it almost entirely defeats the purpose if one uses the `postgres` user -- that's not arbitrary at all, and in fact is the exact user the image itself will use if none is explicitly specified.

darthbinamira:
ahhh.. right, didn't think of the "arbitrary" part of the proposed change

darthbinamira:
After upgrading to the latest docker as of this writing (18.06.1), I could no longer reproduce the error. So closing this MR.

PR #1301, opened 2018-09-05
Add a short blurb about TLS + WordPress + NGINX

tianon:
See https://github.com/docker-library/wordpress/blob/cb50cbaabba8eea67ea6a1f7702a7a1e238f10b9/docker-entrypoint.sh#L147-L151 (where we auto-add the appropriate `HTTP_X_FORWARDED_PROTO` code to `wp-config.php`).

See also https://codex.wordpress.org/Function_Reference/is_ssl#Notes and http://codex.wordpress.org/Administration_Over_SSL#Using_a_Reverse_Proxy.

Closes https://github.com/docker-library/wordpress/issues/331

### DIFF
--- a/wordpress/content.md
+++ b/wordpress/content.md
@@ -40,6 +40,8 @@ $ docker run --name some-%%REPO%% -e WORDPRESS_DB_HOST=10.1.2.3:3306 \
     -e WORDPRESS_DB_USER=... -e WORDPRESS_DB_PASSWORD=... -d %%IMAGE%%
 ```
 
+When running WordPress with TLS behind a reverse proxy such as NGINX which is responsible for doing TLS termination, be sure to set `X-Forwarded-Proto` appropriately (see ["Using a Reverse Proxy" in "Administration Over SSL" in upstream's documentation](https://codex.wordpress.org/Administration_Over_SSL#Using_a_Reverse_Proxy)). No additional environment variables or configuration should be necessary (this image automatically adds the noted `HTTP_X_FORWARDED_PROTO` code to `wp-config.php` if *any* of the above-noted environment variables are specified).
+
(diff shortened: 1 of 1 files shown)


wordpress/content.md
```diff
@@ -40,6 +40,8 @@ $ docker run --name some-%%REPO%% -e WORDPRESS_DB_HOST=10.1.2.3:3306 \
     -e WORDPRESS_DB_USER=... -e WORDPRESS_DB_PASSWORD=... -d %%IMAGE%%
 ```
 
+When running WordPress with TLS behind a reverse proxy such as NGINX which is responsible for doing TLS termination, be sure to set `X-Forwarded-Proto` appropriately (see ["Using a Reverse Proxy" in "Administration Over SSL" in upstream's documentation](https://codex.wordpress.org/Administration_Over_SSL#Using_a_Reverse_Proxy)).
```
yosifkit: We should also note that they won't have to set any **extra** environment variables or edit `wp-config` for the WordPress container since we already inject the the `if proto` statement when they use any of the above environment variables.

tianon: Updated (and rebased).

wordpress/content.md
```diff
@@ -40,6 +40,8 @@ $ docker run --name some-%%REPO%% -e WORDPRESS_DB_HOST=10.1.2.3:3306 \
     -e WORDPRESS_DB_USER=... -e WORDPRESS_DB_PASSWORD=... -d %%IMAGE%%
 ```
 
+When running WordPress with TLS behind a reverse proxy such as NGINX which is responsible for doing TLS termination, be sure to set `X-Forwarded-Proto` appropriately (see ["Using a Reverse Proxy" in "Administration Over SSL" in upstream's documentation](https://codex.wordpress.org/Administration_Over_SSL#Using_a_Reverse_Proxy)). No additional environment variables or configuration should be necessary (this image automatically adds the noted `HTTP_X_FORWARDED_PROTO` code to `wp-config.php`).
```
yosifkit: Did you want to qualify that the `HTTP_X_FORWARDED_PROTO` code is only added to `wp-config` if you use at least one of the above environment variables?

tianon: Fair enough, updated.

PR #1308, opened 2018-09-14
Fix README

aaabramov:


tianon:
Not sure what this is "fixing" ?

Besides "example" being a fine example password, this change would need to be either in `content.md` or `stack.yml` instead (`README.md` is auto-generated).

PR #1312, opened 2018-09-21
Add a blurb about sourcing the Erlang cookie from a file

tianon:
Closes https://github.com/docker-library/rabbitmq/issues/279

See also https://github.com/docker-library/rabbitmq/pull/280.

### DIFF
--- a/rabbitmq/content.md
+++ b/rabbitmq/content.md
@@ -72,6 +72,14 @@ Listing users ...
 guest   [administrator]
 ```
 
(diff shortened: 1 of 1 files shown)


yosifkit:
We should mention the permissions explicitly:

https://github.com/docker-library/rabbitmq/issues/171
```
"Cookie file /var/lib/rabbitmq/.erlang.cookie must be accessible by owner only"
```

tianon:
Updated.

YuryHrytsuk:
@yosifkit I didn't manage to make erlang cookie via docker secrets work until I found out https://github.com/docker-library/rabbitmq/issues/279 issue. 

I could understand that there is a wrong owner but couldn't find out why until I get deep inside code

PR #1316, opened 2018-09-28
Bump IBM Java version to 1.8.0_sr5fp22 (8.0.5.22).

dinogun:


tianon:
(It might be worth considering rewording this such that it doesn't have to have an explicit version number embedded so that you don't have to make two PRs for every version bump. :sweat_smile:  Maybe a one-liner to help users extract the exact version number from the image themselves?)

PR #1317, opened 2018-09-28
Adjust Cassandra configuration section to emphasize using a custom configuration file

tianon:
This also attempts to de-emphasize the use of our fiddly environment variables.


tianon:
Crap, https://github.com/docker-library/cassandra/issues/58 makes this annoying (especially for a bind mount).  More effort required here.

tianon:
Probably have to update this with the `-Dcassandra.config=xxx` solution in https://github.com/docker-library/cassandra/issues/58#issuecomment-343300020.

PR #1322, opened 2018-10-01
Update sentry logo

EvanPurkhiser:
Updates the old sentry logo to the modern logo.


tianon:
It seems like there's something funky going on with this `.png` file.  GitHub doesn't seem to be able to render it, and when I check out the branch locally, `ristretto` is able to display the image, but it can't seem to generate a thumbnail for it.  Any ideas?

tianon:
(For example, https://github.com/docker-library/docs/blob/7d1c6fff37893bcefc186de7b978f5bdb2f801f6/sentry/logo.png shows a broken image.)

tianon:
Of course, the minute I comment that, it works.  Awesome.

EvanPurkhiser:
Yeah I was just looking at that. I resized it with `convert`, let me fiddle. Thanks for the quick review.

tianon:
LGTM

tianon:
I think GitHub was probably just having a bad time -- it appears to be fine now. :+1:

PR #1326, opened 2018-10-08
Fixed config sample for Nginx

Code1110:
The option `access_log on` is invalid and makes nginx create a logfile called `on`.

### DIFF
--- a/telegraf/content.md
+++ b/telegraf/content.md
@@ -113,8 +113,8 @@ Create an `nginx_status.conf` configuration file to expose metric data:
 server {
     listen 8090;
     location /nginx_status {
-        stub_status on;
-        access_log on;
+        stub_status;
+        access_log off;
     }
 }
 ```


tianon:
cc @jsternberg

jsternberg:
nginx is pronounced "engine-x" so "an `nginx_config.conf`" is correct.

If you're uncertain about it because that's a filename, you can just modify it so the problem disappears:

> Create a configuration file at `nginx_status.conf` to expose metric data

The actual content of the configuration file is fine. I think. I don't actually know about nginx configuration.

Code1110:
Ouch, I guess, there was a misunderstanding. My main concerns where about the config file itself:
* The parameter `stub_status` doesn't takes any arguments. See http://nginx.org/en/docs/http/ngx_http_stub_status_module.html.
* The parameter `access_log` can be a path or `off`. So, writing `on` can be misleading, also because normally you don't want to write an access log for the monitoring endpoint. See http://nginx.org/en/docs/http/ngx_http_log_module.html.

I tested this configuration and can confirm, that it works.

tianon:
@jsternberg this is updated, and I think the changes here match the NGINX documentation for these arguments appropriately -- are you good with this now?

PR #1333, opened 2018-10-11
Add note about ghost-cli not intended to work

loasishospitality:
Mention ghost-cli is not working out of the box, refer to @acburdine comment explaining why on https://github.com/docker-library/ghost/issues/156

### DIFF
--- a/ghost/content.md
+++ b/ghost/content.md
@@ -84,9 +84,9 @@ $ docker exec <container-id> node --version
 v6.11.2
 ```
 
-## Note about ghost-cli
+## Note about Ghost-CLI
 
-While the docker images do have ghost-cli available and do use some of its commands to set up the base Ghost image, many of the other ghost-cli commands won't work correctly, and really aren't designed/intended to. For more info see [Why sudo is not packed and ghost-cli wont work out of the box](https://github.com/docker-library/ghost/issues/156#issuecomment-428159861)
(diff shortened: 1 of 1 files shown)


ghost/content.md
```diff
@@ -84,6 +84,10 @@ $ docker exec <container-id> node --version
 v6.11.2
 ```
 
+## Note about ghost-cli
+
+While the docker images do have ghost-cli available and do use some of its commands to set up the base Ghost image, many of the other ghost-cli commands won't work correctly, and really aren't designed/intended to. For more info see [Why sudo is not packed and ghost-cli wont work out of the box](https://github.com/docker-library/ghost/issues/156#issuecomment-428159861)
```
tianon: I think we should capitalize proper nouns throughout this according to their upstream-preferred capitalization.  For example, `ghost-cli` should probably be `Ghost-CLI`, and `docker` should definitely be `Docker`.

Also, I'm not sure it makes sense to mention `sudo` here at all; here's what I'd suggest instead:

> ## Note about Ghost-CLI
> 
> While the Docker images do have Ghost-CLI available and do use some of its commands to set up the base Ghost image, many of the other Ghost-CLI commands won't work correctly, and really aren't designed/intended to. For more details see [docker-library/ghost#156 (comment)](https://github.com/docker-library/ghost/issues/156#issuecomment-428159861).

I'm also happy to amend/carry this for you from here if you'd prefer. :+1: (please don't open further PRs -- if you push to this same branch, it'll simply update this PR in-place)

tianon:
Thanks!

PR #1338, opened 2018-10-17
Add section about upgrading Ghost

wglambert:
Resolves https://github.com/docker-library/ghost/issues/158

### DIFF
--- a/ghost/content.md
+++ b/ghost/content.md
@@ -24,6 +24,12 @@ $ docker run -d --name some-ghost -p 3001:2368 %%IMAGE%%
 
 Then, access it via `http://localhost:3001` or `http://host-ip:3001` in a browser.
 
+### Upgrading Ghost
+
+You will want to ensure you are running the latest minor version (1.25.5 or 0.11.9) of Ghost before upgrading major versions. Otherwise, you may run into database errors.
+
+For upgrading your Ghost container you will want to mount your data to the appropriate path in the predecessor container (see below): import your content from the admin panel, stop the container, and then re-mount your content to the successor container you are upgrading into; you can then export your content from the admin panel.
+
 ## Stateful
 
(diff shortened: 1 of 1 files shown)


ghost/content.md
```diff
@@ -40,23 +46,6 @@ $ docker run -d --name some-ghost -p 3001:2368 -v /path/to/ghost/blog:/var/lib/g
 $ docker run -d --name some-ghost -p 3001:2368 -v /path/to/ghost/blog:/var/lib/ghost %%IMAGE%%:0.11-alpine
 ```
 
-### Breaking changes
```
yosifkit: I think maybe we should keep this section.  It makes the `/var/lib/ghost/content` vs `/var/lib/ghost` clear.

wglambert: I thought the section was redundant with `## STATEFUL` below it also explicitly specifying the proper path, so to make it clear I re-arranged and worded it:
> . . . mount your data to the appropriate path in the predecessor container (see below)

> ## Stateful
> 
> Mount your existing content. In this example we also use the Alpine base image.
> 
> ### Ghost 1.x.x
> 
> ```console
> $ docker run -d --name some-ghost -p 3001:2368 -v /path/to/ghost/blog:/var/lib/ghost/content ghost:1-alpine
> ```
> 
> ### Ghost 0.11.xx
> 
> ```console
> $ docker run -d --name some-ghost -p 3001:2368 -v /path/to/ghost/blog:/var/lib/ghost ghost:0.11-alpine
> ```

And then the next section likewise reiterates the proper paths again

> ### Docker Volume
> 
> Alternatively you can use a [data container](http://docs.docker.com/engine/tutorials/dockervolumes/) that has a volume that points to `/var/lib/ghost/content` (or /var/lib/ghost for 0.11.x) and then reference it:

PR #1339, opened 2018-10-18
Add empty variant-alpine.md

jethr0null:


tianon:
Just to expand -- this is because Elastic doesn't support Alpine, nor recommend it, and the default variant text describing Alpine makes it seem like there's more support there than there is.  Once 5.x is deprecated, all Alpine variants will be unsupported and receive no further updates.  (Hopefully @jethr0null will correct me if any of this is inaccurate. :+1:)

PR #1340, opened 2018-10-21
Odoo: add missing database parameter to `docker run` commands

odony:
As of Odoo 12, the database parameter needs to be explicitly provided, otherwise the database will not be initialized. Our example `docker` run commands need to be updated to reflect this, otherwise users risk meeting an error page if they connect to the Odoo web interface, as the implicit database "odoo" still needs to be initialized.

For older versions this extra parameter will be accepted as well without causing any issue.


odony:
After further internal discussion, it seems this error case would be better handled in Odoo itself with a workaround, rather than expecting users to pass an extra parameter. Let's not change the instructions yet and see how we can handle that more elegantly.
Sorry for the noise! 😓

PR #1352, opened 2018-11-02
Bump IBM Java version to 1.8.0_sr5fp25 (8.0.5.25).

dinogun:


dinogun:
@tianon Sorry missed your comment. Actually I use a script that generates both the PRs so there is not much effort from my side. However I will remove the version specific comment and add a comment on how to get the version instead.

PR #1353, opened 2018-11-05
Update WebSphere Liberty documentation

arthurdm:
Upon feedback from some users we decided to heavily improve the documentation related to non-root.

We're also updating the instructions to make clear that building an `application image` that extends Liberty's base image is the best practice.

### DIFF
--- a/websphere-liberty/content.md
+++ b/websphere-liberty/content.md
@@ -4,7 +4,11 @@ The images in this repository contain IBM WebSphere Application Server Liberty f
 
 # Image User
 
-This image runs by default with `USER 1001` (non-root), as part of group `0`. All of the folders accessed by WebSphere Liberty have been given the appropriate permissions, but if your extending Dockerfile needs permission to another location you can simply temporarily switch into root and provide the needed permissions, example:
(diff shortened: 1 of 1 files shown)


websphere-liberty/content.md
```diff
@@ -4,15 +4,42 @@ The images in this repository contain IBM WebSphere Application Server Liberty f
...
+If you're using Docker v17.09.0-ce and newer you can take advantage of the flag `--chown=<user>:<group>` during either `ADD` or `COPY`. For example: `COPY --chown=1001:0 jvm.options /config/jvm.options`
+
+### Updating permissions after copy
+
+If you need your Dockerfile to work with older versions of Docker CE and don't want to pre-process the permissions of the files you can temporarily switch into root to change the permissions of the needed files. For example:
+
+```dockerfile
+USER root
+RUN chown 1001:0 /config/jvm.options
+RUN chown 1001:0 /output/resources/security/ltpa.keys
```
yosifkit: Do you want to note that if theses files were copied to the image via `COPY` or `ADD`  in a previous line, then they will take up twice as much space (since the file will be present in two docker layers)?

PR #1356, opened 2018-11-09
Update Logstash docs for 6.x

jarpy:
Redo the Logstash docs to focus on the new 6.x images which are
repackaged versions of the images from the upstream vendor, Elastic.

Relates:
 - https://github.com/docker-library/official-images/pull/5039
 - https://github.com/docker-library/logstash/pull/91


jarpy:
cc @jethr0null

jarpy:
This patch also calls out the `latest` tag as being unavailable, so it would be great to delete it if this gets approved/merged. Thanks.

PR #1357, opened 2018-11-09
Update logos for Elasticsearch, Kibana

jarpy:
Elastic updated their branding a bit recently. This patch uses the new logos for Elasticsearch and Kibana.


jarpy:
cc @jethr0null

PR #1358, opened 2018-11-09
Update README.md

Ploppy3:
Missing argument to publish the port when running the container


tianon:
Sorry for the long delay :bow:

This was intentionally left as an exercise for the reader (and would be completely different if using `php:fpm`, for example). :heart:

PR #1362, opened 2018-11-18
Update Memory settings documentation for Java 11

alexnederlof:
Since Java 11, the `UseCGroupMemoryLimitForHeap` is removed, and enabled by default. See https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-3B1CE181-CD30-4178-9602-230B800D4FAE

I've updated the docs to reflect this.


tianon:
Thanks! :+1:

PR #1366, opened 2018-11-21
Adjust PHP documentation organization, especially to utilize "variants" template functionality

tianon:
This will hopefully help clarify some of the image variants and should remove things like the Apache documentation from arm32v6/php (where the image doesn't exist so the documentation doesn't apply).

Closes https://github.com/docker-library/php/issues/741
Refs https://github.com/docker-library/wordpress/issues/346#issuecomment-439700124

### DIFF
--- a/php/variant-apache.md
+++ b/php/variant-apache.md
@@ -1,0 +1,40 @@
+## `%%IMAGE%%:<version>-apache`
+
+This image contains Debian's Apache httpd in conjunction with PHP (as `mod_php`) and uses `mpm_prefork` by default.
+
+### Apache with a `Dockerfile`
+
+```dockerfile
+FROM %%IMAGE%%:7.2-apache
+COPY src/ /var/www/html/
+```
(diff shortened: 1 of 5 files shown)


php/variant-apache.md
```diff
@@ -0,0 +1,40 @@
+## `%%IMAGE%%:<version>-apache
```
yosifkit: missing closing backtick :open_mouth: :wink:

php/variant-fpm.md
```diff
@@ -0,0 +1,13 @@
+## `%%IMAGE%%:<version>-fpm
```
yosifkit: missing closing backtick :open_mouth: :wink:

php/variant-apache.md
tianon: I blame the fellow I stole most of this content from while it was still a WIP on his computer. :wink: :smile:

PR #1369, opened 2018-11-27
Docs for onboarding Amazon Corretto as an official docker image.

EricEdens:
This PR adds a new official image containing [Amazon Corretto 8](https://aws.amazon.com/corretto/) installed into [Amazon Linux 2](https://hub.docker.com/_/amazonlinux/).

The image contains the full JDK 8, at the latest security update 8u192, with common commands such as `java` and `javac` symlinked into the user's path.


EricEdens:
PR for library definitions: https://github.com/docker-library/official-images/pull/5113

amazoncorretto/issues.md
```diff
@@ -0,0 +1 @@
+https://github.com/corretto/corretto-8-docker/issues
```
tianon: Since this is equivalent to the default (https://github.com/docker-library/docs/blob/0ae2dfcd0df23423f769d7bd9a27675df0ed190d/.template-helpers/issues.md), you should likely simply remove it, especially so that it gets linkified properly. :+1:

EricEdens:
> logo.png looks like it's just the standard AWS logo -- is that intentional?

For the moment that's what we have. We'll make sure to keep it updated, though.

> Since this is equivalent to the default (https://github.com/docker-library/docs/blob/0ae2dfcd0df23423f769d7bd9a27675df0ed190d/.template-helpers/issues.md), you should likely simply remove it, especially so that it gets linkified properly.

All done!

amazoncorretto/issues.md
EricEdens: All done!

PR #1370, opened 2018-11-29
Remove references to the ibmjava version, add a link instead.

dinogun:


dinogun:
@tianon No need to update the ibmjava docs with every release after this change !

tianon:
We'll hit the bump PR later today (I'm on my phone ATM) but this looks great and is now independent of the bump 👌👍💜💚🎉🎊

PR #1371, opened 2018-11-30
[percona] update Bug Report instruction

delgod:


tianon:
Maybe a link off to https://github.com/percona/percona-docker#submitting-bug-reports would be sufficient? (Then you can maintain more verbose details there instead?)

delgod:
I am simplified things a bit ;)

PR #1375, opened 2018-12-08
Updated content.md

### LINKED ISSUE #317: Debian Dockerfiles not under docker-library?
Why is the documentation for the official Debian images located here, but pointing at https://github.com/tianon/docker-brew-debian for the actual Dockerfiles?

Shouldn't it be a separate repository under docker-library like most of the other stuff?

(Yes, I have noticed that tianon is a contributor to this repository as well, but IMO it *feels better* if it's all under one umbrella.)

propri:
Fixes #317

Wrong parenthesis for JSON Wikipedia link


tianon:
Nice catch, thanks!

PR #1381, opened 2018-12-14
[ROS] Add Official Docker image for ROS2

ruffsl:
In preparation for the release of for adding ROS2 to images to the Official Docker Library: https://github.com/docker-library/official-images/pull/5183

This README is derived from that of ROS1, but has been adapted to reflect ROS2. Changes include:
* Added Dockerfile examples for both installing and building ROS packages
* Removed more elementy/repetitive Docker CLI walkthroughs-
* Inplace with concise docker-compose examples
* Added example using ROS2 with security enabled

Let me know if you have any thoughts or suggestions.
ping @tfoote @nuclearsandwich @sloretz


ros2/content.md
```diff
@@ -0,0 +1,265 @@
...
+    command: ros2 run demo_nodes_py listener
+```
+
+> Use docker-copose inside the same directory to launch our ROS nodes. Given the containers created derive from the same docker compose project, they will coexist on shared project network:
+
+```console
+$ docker-compose up -d
+```
+
+> Notice that a new network named `ros2_demos` has been created, as can be shown further with:
```
ruffsl: Is this little trailing intro to docker-compose worth keeping, or should we assume the basics of docker-compose is common knowledge?

ros2/content.md
```diff
@@ -0,0 +1,265 @@
...
+Lets build upon the example above by adding authenticated encryption to the message transport. This is done by leveraging [Secure DDS](https://www.omg.org/spec/DDS-SECURITY). We'll use the same ROS 2 docker image to bootstrap the PKI, CAs, and Digitally Signed files.
+
+> Create a script at `~/ros2_demos/keystore/bootstrap_keystore.bash` to bootstrap a keystore and add entries for each node:
+
+``` shell
+#!/usr/bin/env bash
+# Bootstrap ROS 2 keystore
+ros2 security create_keystore ./
+ros2 security create_key ./ talker
+ros2 security create_key ./ listener
```
ruffsl: created keystore file system may still be owned by root afterward, perhaps it's worth changing the permission recursively to that of the parent directory, and marking all key material read only.

ros2/content.md
```diff
@@ -0,0 +1,265 @@
...
+``` command
+$ docker-compose up
+```
+
+Note: So far this has only added authenticated encryption, i.e. only  participants with public certificates signed by a trusted CA may join the domain. To enable access control within the secure domain, i.e. restrict which and how topics may be used by participants, more such details can be found [here](https://github.com/ros2/sros2/).
+
+# More Resources
+
+[ROS.org](http://www.ros.org/): Main ROS website  
+[Docs](https://docs.ros2.org/): Core Documentation
```
ruffsl: I replaced the wiki that is ROS1 related to avoid misleading traffic to older unrelated content. Are there any other links we'd like to add here?

ros2/content.md
```diff
@@ -0,0 +1,265 @@
+# What is [ROS 2](https://index.ros.org/doc/ros2)?
+
+The Robot Operating System (ROS) is a set of software libraries and tools that help you build robot applications. From drivers to state-of-the-art algorithms, and with powerful developer tools, ROS has what you need for your next robotics project. And it's all open source. ROS 2 is the next-generation ROS: building upon modern middleware for decentralized discovery, message serialization, and secure transport with quality of service enabling high performance, low latency, language agnostic distributed computing for robotic systems.
+
+> [wikipedia.org/wiki/Robot_Operating_System](https://en.wikipedia.org/wiki/Robot_Operating_System)
+
+[%%LOGO%%](https://index.ros.org/doc/ros2)
```
ruffsl: I pulled the new logo from the ROS2 github org gravatar, but perhaps there is a different logo we'd like to use for branding?

ros2/content.md
nuclearsandwich: https://index.ros.org/doc/ros2/ might be a better entry point than docs.ros2.org.

It's the new home of the ROS2 wiki which moved into a generated site so changes were easier to manage with a pull-request workflow.

tianon:
I'm totally lost as to why this is being proposed as a new image instead of
being tags of the existing "ros" image?

ruffsl:
> I'm totally lost as to why this is being proposed as a new image instead of being tags of the existing "ros" image?

Very understandable; the confusion perhaps stems from that the projects are so similarly named but in many ways separate parallel efforts. Among the differences between ROS (hereon referred to as ROS1) and ROS2: separate package ecosystem, non-interchangeable APIs, different wire-transport implementation, new build system, modified message types, different CLI syntax, etc... Although ROS1 and ROS2 share similar concepts of topics/services/actions/parameters/nodes/namespaces, they are not compatible.

I suppose one may argue that such different projects sharting the same naming could collocate under the same repo using only tags, like with python and version 2 and 3. Yet this comparison falls short in few ways:
* In Python it is possible write a package that support both a v2 and v3 runtime
  * Not possible with ROS1 and ROS2 as api/apt/build/install/launching of packages are different
* Python is a programming language and is released by versions
  * ROS1 and ROS2 are more like frameworks or ecosystems, each with distro generations

Another difference is that both ROS1 and ROS2 are ongoing projects. Though the distro names of the two do not collide, I would suspect it confusing to folks reading tags and attempting to determine which are ROS1 compient and which are ROS2, or what is the latest LTS version of either?:
* `ros:indigo`
* `ros:kinetic`
* `ros:bouncy`
* `ros:lunar`
* `ros:crystal`
* `ros:melodic`

But suppose we could add another sub-super prefix to the tags to denote ROS1 vs ROS2, but at that point I think the repo name would be nice to do that with. 

Another is that currently the `latest` tag currently points to the most recent LTS distro. How should that be reconciled when each has its own latest latest distro? If someone reads a historic `FROM ros` in a dockerfile, a run script, or a docker compose project, how are they to know if it refers to a ROS1 distro or a ROS2 distro?

Lastly, the projects necessitate different instructions in how they are containerized, again due to the build/install and run CLIs, but also as the transport implementation,  e.g. XMLRPC vs DDS, centralized master broker vs disitrolized discovery, requires different user documentation, thus the differences between the ROS1 and ROS2 README docs I've written.

For more background on branding, see the communities 'Why ROS 2.0' design artificial for further motivation:
https://design.ros2.org/articles/why_ros2.html

Perhaps @nuclearsandwich @tfoote or @sloretz could speak to this effect or suggest alternatives.

ros2/content.md
ruffsl: If we plan on merging with the current `ros` repo, then I'll just go back to the version agnostic logo.

ros2/content.md
```diff
@@ -0,0 +1,265 @@
+# What is [ROS 2](https://index.ros.org/doc/ros2)?
```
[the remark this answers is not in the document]
ruffsl: Cool.

ros2/content.md
ruffsl: Well, docs.ros2.org has all the nicely organized sphinx pages. I feel the index url is still a jumble.
Is the intent that the index will become a ordered nesting of sphinx project docs, like say python's official doc pages?

ruffsl:
> We talked about this internally and I think that although we've been hosting many ROS2 resources in separate locations this is just another version of ROS so it does make sense to just use new tags under the existing repository.

@tfoote , I'd fine with merging the repos. I'll have to rework the CI, but doable.
Still, I have some follow up questions:

#### Tagging

What should be tagged as latest? Use the latests supported ROS1 LTS, or the most recent release of ROS2?

>  There will be overlapping alphabetic listings and latest will give you the latest LTS which for now will stay ROS1 until we have an official LTS for ROS2 then we can switch it over.

I'm leaning towards the most release of ROS2 support migration (or perhaps the latest stable release of ROS2, depending on if we follow through with rolling releases), but is sounds like you'd recommend leaving it to ROS1 for now until a ROS2 LTS is released. 

#### Documentation
Should we still overhaul the documentation to include ROS2 though? I think mediating the nuances between ROS1 and ROS2 deployment would be out of scope in the simple introductory readme. As mentioned above, the details one has to consider when containizing ROS is rather different between versions.

ros2/content.md
```diff
@@ -0,0 +1,265 @@
...
+
+> Use a temporary container to run the keystore bootstraping script in the keystore directory:
+
+
+```console
+$ docker run -it --rm \
+    --env-file ./config.env \
+    --volume ./keystore:/keystore:rw \
+    --workdir /keystore \
+    ros2 bash bootstrap_keystore.bash
```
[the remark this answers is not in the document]
ruffsl: Fixed in 347e32d

ros2/content.md
ruffsl: Fixed in 347e32d

ros/content.md
```diff
@@ -1,28 +1,104 @@
...
+      ros-${ROS_DISTRO}-demo-nodes-py \
+      ros-${ROS_DISTRO}-sros2 && \
+    rm -rf /var/lib/apt/lists/*
+
+# run ros package launch file
+CMD ["ros2", "launch", "demo_nodes_cpp", "talker_listener.launch.py"]
 ```
 
-You can then build and run the Docker image:
+Note: all ROS images include a default entrypoint that sources the ROS environment setup before exiting the configured command, in this case the demo packages launch file. You can then build and run the Docker image like so:
```
sloretz: exiting -> executing

ruffsl:
> Since the ros image will have tags for ros 1 and ros2 distros, can we keep the content from both?

I don't know if the docker repo documentation page would be the best place to expand upon the nuances between ros versions. Perhaps we could outsource that to an external reference? Migrate the old ros1 docker examples to the ROS 1 wiki and just link to the old docs for past versions.

I was thinking we could wait on merging on this docs pr until a ROS2 LTS is released to take over the 'latest' tag. In the meantime, I can include these new examples with an announcement when the ros2 images are released.

tianon:
@ruffsl looks like this maybe needs some updates? Is it still relevant?

ruffsl:
> @ruffsl looks like this maybe needs some updates? Is it still relevant?

Its relatively polished and ready, I'd say. I'm just waiting to get some final feedback from other ROS maintainer as well as migrating the `latest` tag to point to the newest ROS2 LTS release. I'm also kind of waiting until the ROS snapshot repos are ready before decommissioning the EOL'd tags for transitioning. 

https://github.com/osrf/docker_images/issues/282#issuecomment-504232106

ros/content.md
```diff
@@ -1,28 +1,99 @@
-# What is [ROS](http://www.ros.org/)?
+# What is [ROS](https://index.ros.org/doc/ros2)?
```
mikaelarguedas: I understand the conclusion of merging repos and documentation, but is there a way to specify in some relevant places when the 2 versions vary. I'm thinking expecially here for documentation entry point and core licensing.

ros/content.md
```diff
@@ -1,28 +1,99 @@
...
+
+# build package workspace
+RUN . /opt/ros/$ROS_DISTRO/setup.sh && \
+    colcon build \
+      --packages-select \
+        demo_nodes_cpp \
+      --mixin release
+
+# create new multi-stage
+FROM %%IMAGE%%:dashing-ros-core AS runner
```
mikaelarguedas: Could we end up with missing runtme dependencies by having the runner being based on a different image than the builder?
Maybe the runner should use rosdep to install runtime dependencies?

ros/license.md
```diff
@@ -1,3 +1,3 @@
-The core of ROS is licensed under the standard three-clause BSD license. This is a very permissive open license that allows for reuse in commercial and closed source products. You can find more about the BSD license from the Opensource.org [BSD 3-Clause](http://opensource.org/licenses/BSD-3-Clause) page and Wikipedia's [BSD Licenses](http://en.wikipedia.org/wiki/BSD_licenses) entry.
+The core of ROS is licensed under the Apache License 2.0. This is a very permissive open license that allows for modification, distribution, commercial use, patent use, and private use in open and closed source products. You can find more about the license from the [Apache License](https://www.apache.org/licenses) website or Opensource.org [Apache License 2.0](https://opensource.org/licenses/Apache-2.0) page.
```
mikaelarguedas: Not sure how to deal with this but the 2 project (ROS and ROS 2) have different license for their core and this change would apply only to ROS 2

ros/content.md
```diff
@@ -31,16 +102,16 @@ Developing such complex systems with cutting edge implementations of newly publi
...
 
+For a complete listing of supported architecture and base images for each ROS Distribution Release, please read the official REP on target platforms for either [ROS 1](https://www.ros.org/reps/rep-0003.html) or for [ROS 2](https://www.ros.org/reps/rep-2000.html).
+
 ## Deployment suggestions
 
 The available tags include supported distros along with a hierarchy tags based off the most common meta-package dependencies, designed to have a small footprint and simple configuration:
 
--	`ros-core`: barebone ROS install
+-	`ros-core`: barebone ROS 2 install
 -	`ros-base`: basic tools and libraries (also tagged with distro name with LTS version as `latest`)
```
mikaelarguedas: This assumes the latest ROS 2 LTS is tagged as latest. This is not currently the case

ros/content.md
```diff
@@ -31,16 +102,16 @@ Developing such complex systems with cutting edge implementations of newly publi
...
 With the advancements and standardization of software containers, roboticists are primed to acquire a host of improved developer tooling for building and shipping software. To help alleviate the growing pains and technical challenges of adopting new practices, we have focused on providing an official resource for using ROS with these new technologies.
 
+For a complete listing of supported architecture and base images for each ROS Distribution Release, please read the official REP on target platforms for either [ROS 1](https://www.ros.org/reps/rep-0003.html) or for [ROS 2](https://www.ros.org/reps/rep-2000.html).
+
 ## Deployment suggestions
 
 The available tags include supported distros along with a hierarchy tags based off the most common meta-package dependencies, designed to have a small footprint and simple configuration:
 
--	`ros-core`: barebone ROS install
+-	`ros-core`: barebone ROS 2 install
```
mikaelarguedas: If the goal is to conflate ROS 1 and 2 in this doc, is there a need to call out the version here?

ros/content.md
```diff
@@ -31,16 +102,16 @@ Developing such complex systems with cutting edge implementations of newly publi
...
 -	`ros-base`: basic tools and libraries (also tagged with distro name with LTS version as `latest`)
--	`robot`: basic install for robots
--	`perception`: basic install for perception tasks
 
-The rest of the common meta-packages such as `desktop` and `desktop-full` are hosted on automatic build repos under OSRF's Docker Hub profile [here](https://hub.docker.com/r/osrf/ros/). These meta-packages include graphical dependencies and hook a host of other large packages such as X11, X server, etc. So in the interest of keep the official images lean and secure, the desktop packages are just be hosted with OSRF's profile.
+The rest of the common meta-packages such as `desktop` and `ros1-bridge` are hosted on automatic build repos under OSRF's Docker Hub profile [here](https://hub.docker.com/r/osrf/ros/). These meta-packages include graphical dependencies and hook a host of other large packages such as X11, X server, etc. So in the interest of keep the official images lean and secure, the desktop packages are just be hosted with OSRF's profile.
```
mikaelarguedas: Same here, ros1-bridge being ROS 2 only maybe its not worth pointing it out here.

ros/content.md
```diff
@@ -58,185 +129,106 @@ Some application may require device access for acquiring images from connected c
...
-> If we then use `rostopic` to list published message topics, we should see something like this:
+### ROS 1 Bridge
 
-```console
-$ rostopic list
-/chatter
-/rosout
-/rosout_agg
-```
+To ease ROS 2 migration, [`ros1_bridge`](https://index.ros.org/p/ros1_bridge/github-ros2-ros1_bridge) is a ROS 2 package that provides bidirectional communication between ROS 1 and ROS 2. As a minimal example, given the ROS 2 Dockerfile above, we'll create the ROS 1 equivalent bellow, and name the Dockerfile appropriately.
```
mikaelarguedas: typo: bellow -> below

ros/content.md
```diff
@@ -58,185 +129,106 @@ Some application may require device access for acquiring images from connected c
...
-> Ok, now that we see the two nodes are communicating, let get inside one of the containers and do some introspection what exactly the topics are:
-
-```console
-$ docker exec -it master bash
-$ source /ros_entrypoint.sh
-```
+> Note: the auto-generated network, `ros_demos`, will persist until you explicitly remove it using `docker-compose down`.
 
-> If we then use `rostopic` to list published message topics, we should see something like this:
+### ROS 1 Bridge
```
mikaelarguedas: Thanks for adding this :+1:

ros/content.md
mikaelarguedas: Linking to the reps defining ROS variants would be useful here

ros/content.md
```diff
@@ -58,185 +129,106 @@ Some application may require device access for acquiring images from connected c
...
+> Create the directory `~/ros_demos` and add the first `Dockerfile` example from above. In the same directory, also create file `docker-compose.yml` with the following that runs a C++ publisher with a Python subscriber:
 
-#### Run services
+```yaml
+version: '3'
 
-> To create a container for the ROS master and advertise it's service:
+services:
+  talker:
+    build: ./Dockerfile
```
mikaelarguedas:
```suggestion
    build:
      context: .
      dockerfile: Dockerfile
```

ros/content.md
```diff
@@ -58,185 +129,106 @@ Some application may require device access for acquiring images from connected c
...
+    command: ros2 run demo_nodes_cpp talker
 
-```console
-$ docker run -it --rm \
-    --net foo \
-    --name master \
-    %%IMAGE%%:ros-tutorials \
-    roscore
+  listener:
+    build: ./Dockerfile
```
mikaelarguedas:
```suggestion
    build:
      context: .
      dockerfile: Dockerfile
```

ros/content.md
mikaelarguedas: This is to avoid errors like:
```
  File "/usr/lib/python3/dist-packages/docker/api/build.py", line 142, in build
    raise TypeError("You must specify a directory to build in path")
TypeError: You must specify a directory to build in path
```

(on docker 19.03.2)

ros/content.md
gavanderhoorn: @mikaelarguedas' question is actually partly why I posted [Generating ‘dev’ and runtime artefacts from ROS packages](https://discourse.ros.org/t/generating-dev-and-runtime-artefacts-from-ros-packages/12448).

ruffsl: Looks like using rosdep to install runtime dependencies has been ticketed here:
https://github.com/ros-infrastructure/rosdep/pull/727

Once rosdep gets that feature, we could update example to use it. In the meantime, we'll just make a note to make sure the user is aware of need for runtime depencicies.

gavanderhoorn: Doesn't this basically mean that these `Dockerfile`s are currently broken? Without the runtime dependencies, what's in `install` most likely doesn't run in the `runner` container.

mikaelarguedas: > Once rosdep gets that feature, we could update example to use it. In the meantime, we'll just make a note to make sure the user is aware of need for runtime depencicies.

IMO this doesnt justify not installing runtime dependencies in the runner image. So I would argue that rosdep install should be used in the runner image and once rosdep allows to restrict the type of dependency to install, we can update the example to install "only runtime" dependencies.

These example dockerfiles work but only out of luck because the talker listener happen to have all their dependencies satisfied by `ros_core`. This will most likely not be the case for most use cases making this example confusing for users.

---

This also brings another point that is not explained here that is "why use ros-base a the builder base when ros-core is enough and consistent with the runner?"
This needs some comment explaining that the ros-core image doesnt come with colcon and other build tools anymore so ros_base needs to be used instead.

ruffsl: > This needs some comment explaining that the ros-core image doesnt come with colcon and other build tools anymore so ros_base needs to be used instead.

Yep, this'll need to be updated.

> These example dockerfiles work but only out of luck because the talker listener happen to have all their dependencies satisfied by ros_core. This will most likely not be the case for most use cases making this example confusing for users.

What you alternative example package would you recommend? I only chose the talker and listener example as it's the classic demo, and its simple/quick to build, launch, and verify.

mikaelarguedas: I think it's good to keep the `demo_nodes_cpp` package as the example :+1:

A good way to illustrate the need for installing runtime dependencies in the runner would be to launch the `add_two_ints.launch.py` example instead. As that uses insterfaces from `example_interfaces` that are not in ros_core, the container will fail.

mikaelarguedas: In other words here is the runner I'm suggesting:

```
# create new multi-stage
FROM %%IMAGE%%:dashing-ros-core-bionic AS runner

# copy package install
ENV ROS_WS /opt/ros_ws
COPY --from=builder  $ROS_WS/install $ROS_WS/install

# install package dependencies
RUN apt-get update && apt-get install -y python3-rosdep && \
    apt-mark manual sudo && \
    rosdep init && \
    rosdep update --rosdistro $ROS_DISTRO && \
    . $ROS_WS/install/setup.sh && \
    rosdep install -y \
      --from-paths \
        $ROS_WS/install/share \
      --ignore-src && \
    apt-get remove -y python3-rosdep && apt-get autoremove -y && \
    rm -rf /var/lib/apt/lists/*

# source entrypoint setup
RUN sed --in-place --expression \
      '$isource "$ROS_WS/install/setup.bash"' \
      /ros_entrypoint.sh
# run launch file
CMD ["ros2", "launch", "demo_nodes_cpp", "add_two_ints.launch.py"]
```

ros2/content.md
ruffsl: See https://github.com/docker-library/docs/pull/1381/commits/6d09179854033de0cc43bbf426832b327c6575f5

ros/content.md
ruffsl: How about splitting the the to into separate bullet points: https://github.com/docker-library/docs/pull/1381/commits/6d09179854033de0cc43bbf426832b327c6575f5

ros/content.md
ruffsl: This will change with the new LTS foxy https://github.com/osrf/docker_images/pull/395

ros/license.md
ruffsl: We can simply follow with other library images and link to external resources for finding licences of included packages: https://github.com/docker-library/docs/pull/1381/commits/23bf362b2369c36acefb18fa7d85559fe1cf72aa

ros/content.md
ruffsl: Keeping ros org landing page as main link 966f6ae

ros/content.md
ruffsl: Updated examples 66047eb

ros/content.md
ruffsl: Updated examples 66047eb

ros/content.md
ruffsl: I think having to install rosdep back into ros-core is a little counter to its purpose. If rosdep could pipe a list of package names to a files that satisfy runtime requirements, then that file could be copied and piped back into apt-get in the runner image.

Although, this is still blocked by rosdep's feature set, so until then I've updated the docs to make careful note of this in 66047eb

ros/content.md
ruffsl: Switching to be version generic only when needed. 2cd312f

ros/content.md
mikaelarguedas: > this is still blocked by rosdep's feature set

the rosdep feature set allows to do it. It's just not as optimized as it could be and we can easily add a flag once there is one to restrict the dependency set while preserving the logic of "automatically installing the dependencies you need" rather than having to figure it out and install them by hand.

I still think that this example is incorrect without it, either you're lucky and your dependencies are already in the base image, at which point rosdep is not necessary in the builder image either. Or (what I would guess is the vast majority of the cases) you need to install extra dependencies and ROS provides a tool for that called rosdep that we should use.

So I'm still in favor of using rosdep in the runner image as of now, uninstalling rosdep after was just an extra cleanup that may or may not be justified as rosdep has a pretty small footprint.

> If rosdep could pipe a list of package names to a files that satisfy runtime requirements, then that file could be copied and piped back into apt-get in the runner image.

As you use a different base image for building and running, the list of missing dependencies to install will differ between the builder or the runner so it doesnt seem bulletproof.

> helping to reconcile runtime vs development dependency discrepancies you may have otherwise forgotten to declare in your `package.xml` manifest

If the package.xml is not used to install dependencies, this approach cannot help catch missing or mismatching dependency declarations.

ruffsl: > As you use a different base image for building and running, the list of missing dependencies to install will differ between the builder or the runner so it doesnt seem bulletproof.

That's where I think there should be a dry run option for rosdep where it just returns all the resolved runtime dependencies, regardless if they are installed or not, which would be complete enough for the runner to apt install. We do this for the CI in the navigation stack, but with a few more POSIX tools.

> If the package.xml is not used to install dependencies, this approach cannot help catch missing or mismatching dependency declarations.

That out of band dependency management then just falls onto the developer doesn't it?
rosdep can't help the developer if the developer doesn't also help rosdep.

mikaelarguedas: > That out of band dependency management then just falls onto the developer doesn't it?
rosdep can't help the developer if the developer doesn't also help rosdep.

That's why I'm arguing for using rosdep in the example :wink: 
Otherwise this statement is misleading because the example does *not* allow catching mismatching dependency declarations because it ignores the declared dependencies (by not using rosdep)

> That's where I think there should be a dry run option for rosdep where it just returns all the resolved runtime dependencies

There is currently a dry run option, that lists all the installation commands it would run, (but it only lists the missing packages).
Is there an issue on rosdep to do the same for all (even already installed) packages ? it doesnt look like it would be a difficult feature to add.

ruffsl: I don't know of such a ticket, but yeah, it should only need a small change to support. Then we can pipe those apt package names to a placeholder file to share across multistage.

mikaelarguedas: > Then we can pipe those apt package names

I think piping the list of commands is better than the apt package names. Otherwise how would you handle other installers (e.g. pip packages)? trying to apt install those will likely not work.

> I don't know of such a ticket

I encourage you to open one if that's a feature you need and end up working around in other places.

In the meantime, can we agree on updating the example to install dependencies in the runner to move this forward? once extra rosdep features are available we can see how to update/optimize this example

gavanderhoorn: The `--simulate` option of `rosdep` could *almost* do what you are asking for I believe.

See also [How can generate a list of dependencies from the src folder of a workspace?](https://answers.ros.org/question/251732) on ROS Answers.

mikaelarguedas: :+1: yes that's the one I had in mind in:

> There is currently a dry run option, that lists all the installation commands it would run, (but it only lists the missing packages).

ros/content.md
ruffsl: 50a2c27

ros/content.md
ruffsl: > I encourage you to open one if that's a feature you need

On a closer search, looks like it's been open for while:

https://github.com/ros-infrastructure/rosdep/issues/587
https://github.com/ros-infrastructure/rosdep/pull/727

I've changed the multistage example to demo build and dependency caching with 2316f47 . We can always append a runtime stage later when rosdep is better up to the task. I think it best to keep the advance example clean and development oriented, so we can leave it at the builder stage to install exec dependencies for now.

ros/content.md
```diff
@@ -1,28 +1,116 @@
-# What is [ROS](http://www.ros.org/)?
+# What is [ROS](https://www.ros.org/)?
```
mikaelarguedas, line 1: As these docs now default to ROS2, we could use this link instead http://www.ros2.org/ (which redirects to https://index.ros.org/doc/ros2/)

ros/content.md
```diff
@@ -1,28 +1,116 @@
...
+ENV OVERLAY_WS $OVERLAY_WS
+RUN sed --in-place --expression \
+      '$isource "$OVERLAY_WS/install/setup.bash"' \
+      /ros_entrypoint.sh
+
+# run launch file
+CMD ["ros2", "launch", "demo_nodes_cpp", "talker_listener.launch.py"]
 ```
 
+The example above starts by using [`vcstools`](https://github.com/vcstools/vcstools) to clone source repos of interest into the cacher stage. One could similarly `COPY` code from the local build context into the source directory as well. Package manifest files are then cached in a temporary directory where the following builder stage may copy from to install necessary dependencies with [`rosdep`](https://github.com/ros-infrastructure/rosdep). This is done prior copying the rest of the source files to preserve the multi stage build cache, given unaltered manifests do not altered declared dependencies, saving time and bandwidth. The overlay is then built using [`colcon`](https://colcon.readthedocs.io/en/released/), the entrypoint updated to source the workspace, and the default command set to launch the demo.
```
mikaelarguedas, line 108: Nit^n: Can we make this one sentence per line ?

mikaelarguedas, line 108:
```suggestion
The example above starts by using [`vcstool`](https://github.com/dirk-thomas/vcstool) to clone source repos of interest into the cacher stage. One could similarly `COPY` code from the local build context into the source directory as well. Package manifest files are then cached in a temporary directory where the following builder stage may copy from to install necessary dependencies with [`rosdep`](https://github.com/ros-infrastructure/rosdep). This is done prior copying the rest of the source files to preserve the multi stage build cache, given unaltered manifests do not altered declared dependencies, saving time and bandwidth. The overlay is then built using [`colcon`](https://colcon.readthedocs.io/en/released/), the entrypoint updated to source the workspace, and the default command set to launch the demo.
```

We're actually using vcstool

ros/content.md
```diff
@@ -31,16 +119,18 @@ Developing such complex systems with cutting edge implementations of newly publi
 
 With the advancements and standardization of software containers, roboticists are primed to acquire a host of improved developer tooling for building and shipping software. To help alleviate the growing pains and technical challenges of adopting new practices, we have focused on providing an official resource for using ROS with these new technologies.
 
+For a complete listing of supported architecture and base images for each ROS Distribution Release, please read the official REP on target platforms for either [ROS1](https://www.ros.org/reps/rep-0003.html) or for [ROS2](https://www.ros.org/reps/rep-2000.html).
```
mikaelarguedas, line 122:
```suggestion
For a complete listing of supported architectures and base images for each ROS Distribution Release, please read the official REP on target platforms for either [ROS1](https://www.ros.org/reps/rep-0003.html) or for [ROS2](https://www.ros.org/reps/rep-2000.html).
```

ros/content.md
```diff
@@ -58,185 +148,115 @@ Some application may require device access for acquiring images from connected c
...
+
+Alternatively, more permissive network setting can be use to share all host network interfaces with the container, such as [`host` network driver](https://docs.docker.com/network/host/), simplifying connectivity with external network participants. Be aware however that this removes the networking namespace separation between containers, and can affect the ability of DDS participants communicate between containers, as documented [here](https://community.rti.com/kb/how-use-rti-connext-dds-communicate-across-docker-containers-using-host-driver).
```
mikaelarguedas, line 153:
```suggestion
Alternatively, more permissive network setting can be used to share all host network interfaces with the container, such as [`host` network driver](https://docs.docker.com/network/host/), simplifying connectivity with external network participants. Be aware however that this removes the networking namespace separation between containers, and can affect the ability of DDS participants to communicate between containers, as documented [here](https://community.rti.com/kb/how-use-rti-connext-dds-communicate-across-docker-containers-using-host-driver).
```

ros/content.md
ruffsl: Is that going to be the new landing page?
I just figured the top most link should point to the top most domain.

ros/content.md
ruffsl: See abf5b03

ros/content.md
ruffsl: I'm going to keep this as is. We can revisit this when updating the docs adding the ros1-bridge tag.

ruffsl:
Ping @tianon @yosifkit , with the release of ROS2 Foxy, this is ready to be merged, as we'd like the docs to be update to reflect the move to ROS2 and provide relevant examples.

* https://github.com/osrf/docker_images/pull/395
* https://discourse.ros.org/t/ros-foxy-fitzroy-released/14495

ros/content.md
```diff
@@ -1,50 +1,158 @@
-# What is [ROS](http://www.ros.org/)?
+# What is [ROS](https://www.ros.org/)?
 
-The Robot Operating System (ROS) is a set of software libraries and tools that help you build robot applications. From drivers to state-of-the-art algorithms, and with powerful developer tools, ROS has what you need for your next robotics project. And it's all open source.
+The Robot Operating System (ROS) is a set of software libraries and tools that help you build robot applications.  
+From drivers to state-of-the-art algorithms, and with powerful developer tools, ROS has what you need for your next robotics project.  
+And it's all open source.
```
tianon, line 5: I think you probably want to take a look at the "rich diff" in GitHub for this file -- using one-sentence-per-line results in literal line breaks in the output on GitHub (and Docker Hub's rendering is less consistent, which is why we tend to use one-line-per-paragraph even though it makes the diffs kind of annoying).

ruffsl: I've reverted the one sentence per line commit.

ruffsl:
> so it might make sense to move a lot of this content to somewhere

I think we could trim down the discussion bit that's not as essential as the use examples, but I'll do that in a separate PR, as we'll need to update the tag listing once we've added support for the ros1-bridge tag later on anyway.

PR #1384, opened 2018-12-20
Updating Open Liberty logo

arthurdm:
The current Open Liberty logo on Docker Hub has the words `Open Liberty` and the spaceship, but the spaceship looks [squashed](https://hub.docker.com/_/open-liberty).  

I am replacing it with a logo that has just the spaceship, from https://developer.ibm.com/wasdev/blog/2017/10/17/microprofile-1-2-liberty-17-0-0-3/ol_logomark_132_wht/

fyi @NottyCode 

Signed-off-by: Arthur De Magalhaes <ademagalhaes@gmail.com>


yosifkit:
(Ignoring Travis failure, since it is irrelevant)

While this is a fine change, this does not directly control the image at the top of the page (or on search results). I'll slack message some Docker folks to see if they can also update their copy.

PR #1386, opened 2018-12-21
Added static-file-server to official image documentation.

halverneus:


halverneus:
Closing without merging. See https://github.com/docker-library/official-images/pull/5220 for comments.

PR #1387, opened 2018-12-24
Remove Ubuntu's explicit "sources.list" contents

tianon:
These don't really add any value, and are a source of confusion if/when things change (as seen in https://github.com/tianon/docker-brew-ubuntu-core/issues/138).


tianon:
@mwhudson does this seem sane to remove to you?

mwhudson:
+1, what was that even doing there in the first place...

Issue #1136, opened 2018-01-30
remove the deprecated and removed 8083 Administrator interface port


rdxmb:
> ' 8083 Administrator interface port, if it is enabled ' 

should be removed in the docs, for

> 'The administrator interface is deprecated as of 1.1.0 and will be removed in 1.3.0.'

yosifkit:
Oh, you mean [here](https://github.com/docker-library/docs/blob/e8dfc2f2217fac8ac72428afcbb21cf674dd6b87/influxdb/content.md#exposed-ports) for influxdb?  That would be up to the maintainers of the image.

CC @jsternberg

rdxmb:
yes. Is there another repository for that?

yosifkit:
This is the correct repo for that doc, but I was at first unsure which image you were talking about since there are docs for 136 Docker image repos.  It is up to the maintainer of the image on changes to their docs (which is why I pinged them).

jsternberg:
I've seen this message. I'm going to take a look, but have been unable to.

Just replying to confirm that I've seen this.

tianon:
Closing old issue.  If this is still a problem, a PR would be excellent. 👍

rdxmb:
confused. There **is** a PR linked above, open for about one year.

tianon:
Doh, thanks -- @jsternberg, mind taking a look at #1141?

jsternberg:
Sure.

rdxmb:
wait a second, there is something missing cc @jsternberg

rdxmb:
just created a new PR, so please look at them separately
@jsternberg

PR #1141, opened 2018-02-02
telegraf: remove deprecated port forwarding

petetnt:
8083 was the port for the deprecated admin UI which has since then been replaced with Chronograf.

See https://docs.influxdata.com/influxdb/v1.4/tools/web_admin/ and https://docs.influxdata.com/chronograf/v1.4/guides/transition-web-admin-interface/ for more information

PR #1390, opened 2019-01-04
Influxdb remove admin interface

rdxmb:
fixes https://github.com/docker-library/docs/issues/1136


tianon:
cc @jsternberg :pray: (I guess this conflicts with / replaces #1141?  happy to help resolve these however you see fit)

rdxmb:
@tianon why that?
https://github.com/docker-library/docs/pull/1141 -> telegraf image
this PR -> influxdb image

They do not conflict or replace, but complement each other.

yosifkit:
Sorry, I think @tianon and I were both assuming without fully comprehending that they were for the same image, since it was the same port deprecation. :blush:

This PR has an extra, unrelated line at the end (it looks like text from a merge conflict line :wink:), and there is a useful blank line being removed at [132](https://github.com/docker-library/docs/pull/1390/files#diff-dbc83212c60edbaea4d9c39c6a0a8523L132).

rdxmb:
> This PR has an extra, unrelated line at the end (it looks like text from a merge conflict line wink), and there is a useful blank line being removed at 132.

Oops. Fixed.

rdxmb:
@Cultti fixed. Thanks!

PR #1402, opened 2019-01-15
kong: update README for Kong 1.0 and above

thibaultcha:


truongsinh96:
Hi @thibaultcha , sorry for my stupid question, but could you tell me the reason why we need PostgreSQL 9.6+? Because currently, I check official document about requirements of Kong, it said that it supports Postgres 9.5+.
Thank you.

PR #1408, opened 2019-01-22
fix(#1406): set Mediawiki MYSQL_RANDOM_ROOT_PASSWORD boolean to a string

### LINKED ISSUE #1406: [Mediawiki] Docker compose : services.database.environment.MYSQL_RANDOM_ROOT_PASSWORD contains true, which is an invalid type, it should be a string, number, or a null
Hi,

If you try to run the docker-compose for Mediawiki it does not work because a boolean is not allowed in docker-compose : https://devops.stackexchange.com/questions/537/why-is-one-not-allowed-to-use-a-boolean-in-a-docker-compose-yml

Setting it to a string fix the issue.

Atem18:
https://devops.stackexchange.com/questions/537/why-is-one-not-allowed-to-use-a-boolean-in-a-docker-compose-yml

Fixes #1406

### DIFF
--- a/mediawiki/stack.yml
+++ b/mediawiki/stack.yml
@@ -25,4 +25,4 @@ services:
       MYSQL_DATABASE: my_wiki
       MYSQL_USER: wikiuser
       MYSQL_PASSWORD: example
-      MYSQL_RANDOM_ROOT_PASSWORD: yes
+      MYSQL_RANDOM_ROOT_PASSWORD: 'yes'


tianon:
Ouch, looks like Travis is failing on this due to https://bitbucket.org/ruamel/yaml/issues/181/preserve-quotes-on-boolean-values (`ymlfmt` is parsing it as YAML 1.2 and `docker-compose` appears to be using YAML 1.1). :disappointed:

Atem18:
That's sad :( . So I should close the issue here and open one here : https://github.com/docker/compose ?

tianon:
No, I think I need to fix `ymlfmt` instead (since compose/stack is really
the source of truth on how to parse these files). ♥️

Atem18:
Ok, have fun then :p

tianon:
Ok, I _think_ I've managed it: https://github.com/tianon/dockerfiles/commit/f363a3df4c398632c301615a110b2ccc6bd31c35 :sweat_smile:

(I used this change to test and make sure it does the right thing here -- before this change, it wants to convert that `yes` into `true`, and after this change it leaves the file alone, and it doesn't suggest any changes to other YAML files.)

I've restarted CI which should _hopefully_ come back green this time. :crossed_fingers: (if it does, I'll be merging :+1:)

tianon:
(Rebased and swapped from `"` to `'`, and now we're green! :+1: :tada:)

PR #1414, opened 2019-01-31
add note for enterprise containers

alvin-huang:
cc: @pearkes

### DIFF
--- a/consul/content.md
+++ b/consul/content.md
@@ -5,6 +5,9 @@ Consul is a distributed, highly-available, and multi-datacenter aware tool for s
 -	[Consul documentation](https://www.consul.io/)
 -	[Consul on GitHub](https://github.com/hashicorp/consul)
 
+## Notice for (x.y.z-ent) Image Variants
+This HashiCorp enterprise software is for use only by customers who have a valid and active license agreement with HashiCorp or an authorized HashiCorp reseller. If you do not have a valid license to use this software, you may not download or otherwise use it. All use of this software is subject to the terms and conditions of your license agreement, and all other rights are expressly reserved.
+
 %%LOGO%%
 
 # Consul and Docker


alvin-huang:
@yosifkit I should've added a bit of context to the PR. We are in the process of releasing the enterprise binaries to Dockerhub here https://github.com/docker-library/official-images/pull/5378 but wanted to get messaging in the docs to go with it so no one is caught off guard. I am fine with waiting until that PR is merged before getting the doc update in, just wanted to keep the gap as small as possible.

consul/content.md
```diff
@@ -5,6 +5,9 @@ Consul is a distributed, highly-available, and multi-datacenter aware tool for s
 -	[Consul documentation](https://www.consul.io/)
 -	[Consul on GitHub](https://github.com/hashicorp/consul)
 
+## Notice for (x.y.z-ent) Image Variants
+This HashiCorp enterprise software is for use only by customers who have a valid and active license agreement with HashiCorp or an authorized HashiCorp reseller. If you do not have a valid license to use this software, you may not download or otherwise use it. All use of this software is subject to the terms and conditions of your license agreement, and all other rights are expressly reserved.
```
tianon: Continuing my comment from https://github.com/docker-library/official-images/pull/5378 (now with the added clarification that `consul:1.4.2-ent` is intended to stay after https://github.com/docker-library/official-images/pull/5378 backfills older versions which then get dropped), it still seems strange to me to include these here if their use is intended to be controlled, limited, or even discouraged, and it seems like a separate `consul/ent` repo would make more sense (and be more easily controlled), or even taking advantage of Docker Store's ability to control image distribution of images/repos via licensing / permissions?

tianon:
Closing, given https://github.com/docker-library/official-images/pull/5378#issuecomment-612254309

PR #1420, opened 2019-02-07
Also mention deprecation in opensuse/README-short.txt

Vogtinator:
https://github.com/docker-library/official-images/issues/5371

### DIFF
--- a/opensuse/README-short.txt
+++ b/opensuse/README-short.txt
@@ -1,1 +1,1 @@
-DEPRECATED - for current official images see opensuse/leap and opensuse/tumbleweed.
+DEPRECATED - for current images by the openSUSE Project see opensuse/leap and opensuse/tumbleweed.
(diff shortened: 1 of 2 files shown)


opensuse/README-short.txt
```diff
@@ -1 +1 @@
-This project contains the stable releases of the openSUSE distribution.
+DEPRECATED - for current official images see opensuse/leap and opensuse/tumbleweed.
```
tianon: I think using the term "official images" here might cause some confusion; how about this instead?

```suggestion
DEPRECATED - for current openSUSE images see opensuse/leap and opensuse/tumbleweed.
```

Vogtinator: The main point of this is to tell users that the official images made by the openSUSE project can be found elsewhere - so I'm against removing "official images" in the short description.

Vogtinator: I reworded it, should be fine now

PR #1424, opened 2019-02-08
[nats] Release v1.4.1

kozlovic:
Details can be found [here](https://github.com/nats-io/gnatsd/releases/tag/v1.4.1)

### DIFF
--- a/nats/content.md
+++ b/nats/content.md
@@ -30,7 +30,7 @@
 # Check "docker run" for more information.
 
 $ docker run -d --name nats-main -p 4222:4222 -p 6222:6222 -p 8222:8222 %%IMAGE%%
-[INF] Starting nats-server version 1.4.0
+[INF] Starting nats-server version 1.4.1
 [INF] Git commit [ce2df36]
 [INF] Starting http monitor on 0.0.0.0:8222
 [INF] Listening for client connections on 0.0.0.0:4222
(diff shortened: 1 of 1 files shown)


kozlovic:
@yosifkit @tianon Hey guys, sorry to bother you, would you know how we can change the logos that appear at top of the page [here](https://hub.docker.com/_/nats) and [here](https://hub.docker.com/_/nats-streaming).

As you can see in each of those repos, we have provided our `logo.png` and it is correctly displayed when one scrolls down past the supported tags and quick reference sections, but not at top of the page or as the result of a search for `nats` like this: `https://hub.docker.com/search?q=nats&type=image`.

Thanks!

yosifkit:
Not a problem! We have been working with the folks on Docker Hub to better use the logo files from docs, but it hasn't happened yet. For now the best way is to go through support@docker.com.

kozlovic:
Thanks!

derekcollison:
I just submitted a support ticket as well. This is still not done. Let us know what we need to do different to get this done. Thanks.

tianon:
It's unfortunately completely outside our control; as far as we know, this is still something that has to change in the Hub directly (and there's no API for us to hit), so Docker's support channels are all we've got.

derekcollison:
We did support email as you suggested, got nothing back. Was hoping you all would have more pull maybe. Been multiple months outstanding.

derekcollison:
ok sent a new support email.. This is kinda silly IMO.

tianon:
I'm definitely sorry about it, and wish there was more we could do, but it's fully outside our control. :disappointed:

PR #1430, opened 2019-02-15
Restricting access only to localhost for non-encrypted access to nextcloud containers

christianlupus:
In the docker-compose examples, the httpd containers (be it apache or nginx) are publicly available on port 8080. In the text it is already mentioned that an appropriate reverse proxy should be used in front of it.
My modifications restrict the ips the containers listen to only localhost. So as long as the reverse proxy is on the same machine, this is perfectly secure without worrying about an additional firewall layer to protect direct access to 8080 from the internet.

### DIFF
--- a/nextcloud/content.md
+++ b/nextcloud/content.md
@@ -168,7 +168,7 @@ services:
   app:
     image: %%IMAGE%%
     ports:
-      - 8080:80
+      - 127.0.0.1:8080:80
     links:
       - db
     volumes:
@@ -217,7 +217,7 @@ services:
   web:
     image: nginx
     ports:
-      - 8080:80
+      - 127.0.0.1:8080:80
     links:
       - app
     volumes:


yosifkit:
This technically works, but only on `docker-compose`; once you move to `docker stack deploy` on swarm mode this would not work (https://github.com/moby/moby/issues/32299). As you note, this requires the reverse proxy to be on the same machine (where one could use a CDN like Cloudflare for the reverse proxy). :man_shrugging: In either case it is partially wrong.

Ping nextcloud image maintainer @tilosp. What would you like to do here?

tianon:
Friendly ping @tilosp

tilosp:
@christianlupus could you create a pr against the readme in https://github.com/nextcloud/docker, because this readme there is basically the canonical source for the docs.

christianlupus:
@tilosp, as you requested I opened a PR on the main repository. Will this repo here be updated if the PR is accepted?

tilosp:
it's currently not automated, but thats on my todo list.
After it's merged into the main repo some maintainer will normally open a pr to keep this repo in sync with the main one.
But if you want you can also leave this pr open to be merged once the one in the main repo is.

christianlupus:
I am ok with closing this PR for now if there is a manual regular sync in place (or you get the automatic sync running).

PR #1431, opened 2019-02-15
Corrected repo reference

jethr0null:
Apologies, the last PR only included updates for the issue text and not the repo ref itself.


tianon:
👍

PR #1433, opened 2019-02-19
Add Source Mage deprecation notice

tianon:
Refs https://github.com/vaygr/docker-sourcemage/issues/2 (thanks @lag-linaro :heart:)

cc @vaygr -- if the `sourcemage` Docker official image is still something you'd like to continue maintaining, please speak up :wink:

(This also isn't permanent: as the wording implies, we'd be happy to have continued maintenance of this image if it's going to be more active, but even upstream appears to only be slightly more fresh than our image with 0.62-11 in 2017.)

### DIFF
--- a/sourcemage/deprecated.md
+++ b/sourcemage/deprecated.md
@@ -1,0 +1,3 @@
+This image is deprecated due to maintainer inactivity (last updated Dec 2016; [docker-library/official-images#2403](https://github.com/docker-library/official-images/pull/2403)).
+
+See [docker-library/docs#1433](https://github.com/docker-library/docs/pull/1433) for discussion around an in-progress upstream update.


vaygr:
Hey @tianon, yeah. However we should move it under @sourcemage umbrella first.

We're mostly based on a rolling release model, so I'd say we do not update our ISO/chroot/docker images very often. If you think there are CVEs we absolutely need to cover and bump our docker image -- please let me know, and I'll do my best to make that happen ASAP.

What you pointed out is a stable grimoire. And due to other priorities and infrastructure migration that's been happening for the last few months, it's not something we can afford to pursue atm, so I think our activity in this period is best tracked via https://github.com/sourcemage/grimoire/commits/master.

tianon:
Thanks @vaygr, that's really helpful context.  I would _definitely_ recommend rebuilding an updated tarball, even if only to signal that this is still active, but also to make sure anyone building `FROM sourcemage` gets a cache bust and thus rebuild triggers.

Additionally, it has been a _long_ time since this image was updated, and it's ~723MB, so I have hard time imagining that there isn't a CVE or two in there somewhere. :smile:

Do you have any idea what the timeline for moving the repo under @sourcemage might be?

lag-linaro:
@vaygr if this project is maintained, could you take a look at https://github.com/vaygr/docker-sourcemage/issues/1 please?

vaygr:
@tianon yeah, that makes sense.

We have some plans to finish infrastructure migration in the next couple of weeks, so I think by the end of March we should be ok to focus on some important stuff, and our Docker image is something we'd like to keep in a good shape.

@lag-linaro sure thing.

tianon:
Hey @vaygr, it's that time again 😀

Any luck getting the aforementioned bits in shape? 👍

tianon:
Ping @vaygr

tianon:
We're about halfway though April now, any luck? 😅

tianon:
Ping @vaygr?

vaygr:
@tianon unfortunately this is taking longer than we expected.

The infrastructure work is still in progress, and on top of that we now have some organizational changes.

So before we can make an update to the docker image, we need to cut the new stable grimoire / a stable chroot image first. I know this is something @wich is actively working on, but we'll still be a couple of months away once it's finished.

So I don't think I can give you solid ETA, not until we're done with the stable cut.

If you're finding it critical for the Hub, please move on with deprecation process and we'll resubmit the image when we're ready again.

tianon:
I wouldn't say it's *hyper* urgent, but I think it does look bad for Source Mage (which I think we can both agree isn't great), hence why I'm trying to keep on top of it and make sure it's still something you're thinking about and working on.  I'll adjust the frequency of my pings accordingly. :heart: :+1:

vaygr:
Thanks for understanding @tianon 
Fully agree here.

Yeah, we're working hard on a new chroot image (in fact the test one was released several days ago) and about to start testing process.

Once that's done, it'll be only a matter of days to get that ready for Docker.

tianon:
Hopefully the chroot tarball testing is going well?

(For context, it's been ~930 days:exclamation: since the last `sourcemage` update :sweat_smile:)

tianon:
Hey, it's me again :smile:

According to https://sourcemage.org/Download, it appears the latest "test" Source Mage chroot image hasn't changed since 20190513 -- is that perhaps because the testing has gone well and no further changes have been necessary? :pray: :heart:

tianon:
Hey @vaygr, we're well over 1000 days now since the last `sourcemage` image update -- any chance this will be resolved soon?

tianon:
We're now at just shy of three and a half years since `library/sourcemage` has had an update (https://github.com/docker-library/official-images/commit/c56a8f8eacb1069ad25423b3347130a80e74a63d). :scream: 

If we don't go with a proper full "deprecation" notice, perhaps we could at least update the image description to make it clear to users that it's not actively updated?

tianon:
Updated the wording to more accurately reflect the discussion in this thread so we can get this merged. :heart:

PR #1435, opened 2019-02-21
WIP: add archlinux image

SantiagoTorres:
This is the first attempt at adding an archlinux official base image


SantiagoTorres:
cc @shibumi @pierres

shibumi:
Seems like there is something wrong with out markdown format

archlinux/content.md
```diff
@@ -0,0 +1,32 @@
...
+# About this image
+
+The base filesystem tarball for this image is auto-generated in Arch Linux infrastructure at the beginning of each month. Given the rolling-release nature of Arch Linux, images are tagged with the timestamp of the date they were generated. For example, `%%IMAGE`:20181107 was generated the Seventh of November of 2018. The latest tag will contain the latest build always.
+
+This image is intended to serve the following goals:
+
+	- Provide the Arch experience in a Docker Image
+	- Provide the most simple but complete image to base every other upon
+	- pacman needs to work out of the box
+	- All installed packages have to be kept unmodified
```
tianon: I think this will probably render as a code block instead of a list, which likely isn't what you intended.  To appease `markdownfmt`, you'll want this to be `-<tab>text` (instead of `<tab>- text` as it is now) and that should properly create a markdown list.

archlinux/content.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Base packages
+
+Given that this image doesn't intend to contain an init system, several packages of the `base` group have been removed.
```
tianon: Is the list of packages removed static enough that they could reasonably be listed here?  If not, is there perhaps somewhere in the GitHub repository we could link for users curious which ones this includes? (perhaps https://github.com/archlinux/archlinux-docker/pull/6 or https://github.com/archlinux/archlinux-docker/blob/master/packages)

Looking at that `packages` file, it does seem to contain the `systemd` package (and I understand why it would probably have to at this point), but it does seem like the wording here is slightly misleading as a result. :sweat_smile:

archlinux/content.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Updating
+
+Arch Linux is a rolling release distribution, so a full update is recommended when installing new packages.
```
tianon: I'm not sure what this is trying to say -- that folks should update all installed packages in the `RUN` lines of `Dockerfile`s which are `FROM` this one?  Would it perhaps instead make sense to recommend that folks pull and rebuild on top of the latest `archlinux:latest` image on a regular basis to ensure they get the full benefits of the rolling release nature of Arch?

archlinux/github-repo
```diff
@@ -0,0 +1 @@
+https://github.com/archlinux/archlinux-docker/
```
tianon: I think this trailing slash is going to lead to some harmless but cosmetically funny-looking links in some generated bits (like `https://github.com/archlinux/archlinux-docker//issues`, for example).

archlinux/license.md
```diff
@@ -0,0 +1 @@
+The Docker image creation scripts contained under the repository %%REPO%% are licensed under GPLv3. However, all applications contained in each tag are licensed under their specific licenses.
```
tianon: Just to be clear, this content is going to have https://github.com/docker-library/docs/blob/73b89b28e5133c925a8dd6104d16ebaf01f5fc46/.template-helpers/license-common.md appended after it (in case that changes your intended wording here).

archlinux/content.md
```diff
@@ -0,0 +1,32 @@
...
+
+Currently we have official packages optimized for the x86-64 architecture. We complement our official package sets with a community-operated package repository that grows in size and quality each and every day.
+
+Our strong community is diverse and helpful, and we pride ourselves on the range of skillsets and uses for Arch that stem from it. Please check out our forums and mailing lists to get your feet wet. Also glance through our wiki if you want to learn more about Arch.
+
+%%LOGO%%
+
+# About this image
+
+The base filesystem tarball for this image is auto-generated in Arch Linux infrastructure at the beginning of each month. Given the rolling-release nature of Arch Linux, images are tagged with the timestamp of the date they were generated. For example, `%%IMAGE`:20181107 was generated the Seventh of November of 2018. The latest tag will contain the latest build always.
```
tianon: Once we're getting ready to merge, I think you'll want to update this example to whatever's the "latest" snapshot at the time so it's not referencing a tag that never existed. :smile:

Also, you'll want to make sure `%%IMAGE` is the full `%%IMAGE%%` and probably should include the tag inside the code markers, ala:

```markdown
For example, `%%IMAGE%%:20190905` was generated ...
```

archlinux/README-short.txt
```diff
@@ -0,0 +1 @@
+Archlinux is a simple, lightweight Linux distribution aimed for flexibility.
```
tianon: Was "Archlinux" here intentional? (I would've expected either "archlinux" or "Arch Linux", but I'll admit I'm not familiar with what's canonical within the Arch community here :sweat_smile:)

archlinux/content.md
```diff
@@ -0,0 +1,32 @@
+# What is Arch Linux?
+
+Arch Linux, is a lightweight and flexible Linux® distribution that tries to Keep It Simple.
+
+Currently we have official packages optimized for the x86-64 architecture. We complement our official package sets with a community-operated package repository that grows in size and quality each and every day.
+
+Our strong community is diverse and helpful, and we pride ourselves on the range of skillsets and uses for Arch that stem from it. Please check out our forums and mailing lists to get your feet wet. Also glance through our wiki if you want to learn more about Arch.
```
tianon: It might be a good idea to turn "our wiki" here into a link to the (unparalleled) Arch wiki so the uninitiated can find it easier. :smile:

archlinux/content.md
SantiagoTorres: Ok! will fix.

archlinux/content.md
SantiagoTorres: Right! I'll review this and make sure it's up to date. I'll also add a forward reference to the packages list.

archlinux/README-short.txt
SantiagoTorres: You're right. I quickly sketched this out and muscle memory has made me make mistakes more often than not ;P

archlinux/content.md
SantiagoTorres: Noted. I'll keep this in mind.

archlinux/license.md
SantiagoTorres: Ah, well yeah that's a little redundant then. Let me reword with this into consideration.

archlinux/github-repo
SantiagoTorres: No harm in removing it then :)

archlinux/content.md
SantiagoTorres: Good point! I'll add the link.

SantiagoTorres:
I think I addressed the comments, but let me know if I missed anything! :+1:

archlinux/README-short.txt
tianon: I think you missed this one. :sweat_smile:

archlinux/content.md
```diff
@@ -0,0 +1,35 @@
...
+# About this image
+
+The base filesystem tarball for this image is auto-generated in Arch Linux infrastructure at the beginning of each month. Given the rolling-release nature of Arch Linux, images are tagged with the timestamp of the date they were generated. For example, `%%IMAGE%%:20190905` was generated the Seventh of November of 2018. The latest tag will contain the latest build always.
+
+This image is intended to serve the following goals:
+
+-   Provide the Arch experience in a Docker Image
+-   Provide the most simple but complete image to base every other upon
+-   pacman needs to work out of the box
+-   All installed packages have to be kept unmodified
```
tianon: Close, but not quite -- this needs to be literal tabs to satisfy `markdownfmt`.  If you trust the tool, you can use the script at the root of the repo to have it apply the change directly (`./markdownfmt.sh -w archlinux/content.md`), or I'd be happy to deal with `markdownfmt` changes if you'd rather. :+1:

```suggestion
-	All installed packages have to be kept unmodified
```

archlinux/content.md
```diff
@@ -0,0 +1,35 @@
...
+## Base packages
+
+Given that that the image is intended to be light-weight several packages of the `base` group have been removed. Please take a look at the [packages](https://github.com/archlinux/archlinux-docker/blob/master/packages) file to see what packages are included on this release.
+
+## Updating
+
+Arch Linux is a rolling release distribution, so a full update is recommended
+when installing new packages. In other words, we suggest to either execute `RUN
+pacman -Syu` immediately after your `FROM` statement or as soon as you docker
+run into a container.
```
tianon: I'm guessing your editor (or muscle-memory) auto-wrapped this paragraph for you? (it's making `markdownfmt` grumpy -- I'm happy to do the `markdownfmt` dancing if you'd rather)

SantiagoTorres:
Thanks for helping me it! :+1:  I just noticed as I was about to push

archlinux/content.md
```diff
@@ -0,0 +1,32 @@
...
+
+Currently we have official packages optimized for the x86-64 architecture. We complement our official package sets with a community-operated package repository that grows in size and quality each and every day.
+
+Our strong community is diverse and helpful, and we pride ourselves on the range of skillsets and uses for Arch that stem from it. Please check out our forums and mailing lists to get your feet wet. Also glance through our [wiki](https://wiki.archlinux.org) if you want to learn more about Arch.
+
+%%LOGO%%
+
+# About this image
+
+The base filesystem tarball for this image is auto-generated in Arch Linux infrastructure at the beginning of each month. Given the rolling-release nature of Arch Linux, images are tagged with the timestamp of the date they were generated. For example, `%%IMAGE%%:20190905` was generated the Seventh of November of 2018. The latest tag will contain the latest build always.
```
yosifkit:
```suggestion
The base filesystem tarball for this image is auto-generated in Arch Linux infrastructure at the beginning of each month. Given the rolling-release nature of Arch Linux, images are tagged with the timestamp of the date they were generated. For example, `%%IMAGE%%:20191006` was generated the Seventh of October of 2019. The `latest` tag will always contain the most recent build.
```
Minor edit and date adjustment.

PR #1440, opened 2019-03-03
Align content with https://github.com/xwiki-contrib/docker-xwiki/commits/master/README.md

vmassol:

PR #1442, opened 2019-03-05
Link to the source documentation to avoid out of sync doc

vmassol:


tianon:
As I noted in https://github.com/docker-library/docs/pull/1440#issuecomment-469295893, I'd recommend keeping at least a bare minimum example here (see https://hub.docker.com/_/registry for one that's doing essentially exactly that), but I won't push too hard if you're set on this (just wanted to throw it out there one more time).

vmassol:
@tianon Thanks. Yes I had seen your comment and thought about it but in the end decided to not go with it because:
1) I don't want that part to become out of sync either and it would be duplicate content to maintain 
2) I'd like to bring users to the doc page and not let them think it's enough to look at this example
3) I don't see any simple example that would satisfactory. I don't wan't push to MySQL before PostgreSQL for example.

Thanks again for suggesting it though! :)

tianon:
Fair enough!

PR #1449, opened 2019-03-17
Update content.md

MattMorgis:
Adds some details to the `README.md` on Docker Hub that may help someone when using the image for the first time.

I had followed the command to run a dev server within a container, however, in order to view the UI on my host machine and to run commands against it from another container, I needed to expose the default `8200` port.

Next, when running `docker run vault status`, I would get:
```
Error checking seal status: Get https://127.0.0.1:8200/v1/sys/seal-status: dial tcp 127.0.0.1:8200: connect: connection refused
```

However, running `docker exec -it <container-id> vault status` would succeed.

It took me ~a minute~ longer than I'd like to admit until I realized the former was being run in a second container, and I needed to pass the `VAULT_ADDR` environment variable, which also needed to include the local IP address of the docker container running the dev server. 

From there, `docker run -e VAULT_ADDR='http://<container-ip>:8200' vault <command>` worked perfectly. _Edit:_ To store a secret, I also needed to pass `VAULT_TOKEN`

### DIFF
--- a/vault/content.md
+++ b/vault/content.md
@@ -59,6 +59,11 @@ Run `vault status` in a second container:
 $ docker run -e VAULT_ADDR='http://<container-ip>:8200' --cap-add=IPC_LOCK vault status
 ```
 
+Store a secret:
+```console
+$ docker run -e VAULT_ADDR='http://<container-ip>:8200' -e VAULT_TOKEN='myroot' --cap-add=IPC_LOCK vault kv put secret/hello foo=world excited=yes
+```
+
 ## Running Vault in Server Mode
 
 ```console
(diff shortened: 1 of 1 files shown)


tianon:
cc @chrishoffman @jefferai @meirish (`vault` image maintainers)

tianon:
> cc @chrishoffman @jefferai @meirish (`vault` image maintainers)

tianon:
(I'm happy to help fix the CI formatting if the content is approved/desired by an image maintainer.)

tianon:
Unfortunately, the `vault` DOI image has since been deprecated :heart:

PR #1452, opened 2019-03-19
Add warning about using PG env vars

dkulp2:
I discovered that the image will not successfully fully initialize and will crash if the environment variables PGHOST, PGDATABASE, PGUSER and PGPASSWORD are included. I thought I'd use the same configmap for client and server in a k8s configuration so that changing, say, a password, would occur in just one place, but I discovered that the Postgres image hangs and then crashes when these environment variables are set. I removed them and created a separate client configmap. All is working OK again.

### DIFF
--- a/postgres/README.md
+++ b/postgres/README.md
@@ -131,6 +131,8 @@ The PostgreSQL image uses several environment variables which are easy to miss. 
 
 **Warning**: the Docker specific variables will only have an effect if you start the container with a data directory that is empty; any pre-existing database will be left untouched on container startup.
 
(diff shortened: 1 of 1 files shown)


tianon:
Sorry for the delay :bow:

I appreciate the sentiment here, but I feel like this might be a little heavy, and I believe the note in https://github.com/docker-library/docs/pull/1129/files#diff-232262385f7ea6d785355ea2345980d0R60 ought to be sufficient. :heart:

However, thank you for the contribution!

PR #1455, opened 2019-03-27
fix docker-erlang-otp github address

getong:


tianon:
Nice! :tada: :birthday: :balloon:

PR #1459, opened 2019-04-01
[Swift] Update content to work with current Docker images

shahmishal:


tianon:
https://github.com/apple/swift-docker/issues/9#issuecomment-328205958 is still accurate:

> Sounds like more caps are now required than before -- definitely time to
> update the description, but we need to figure out the minimal set of
> required caps first (because adding "--privileged" to that description is a
> 100% no-go; you might as well not use a container at that point).

tianon:
I'm disappointed to see https://github.com/apple/swift-docker/issues/9 is now completely gone (not just closed/locked), since it had a lot of good discussion around this issue that I hope is saved somewhere persistent.

(Google's cache has a copy, for now: https://webcache.googleusercontent.com/search?q=cache:ttCmdcBE9RgJ:https://github.com/apple/swift-docker/issues/9)

As noted above and in that issue discussion, `--privileged` is not the correct solution.  In my testing (as I noted there), taking Docker's default seccomp profile and adding `personality` as an allowed syscall and using `--cap-add sys_ptrace` was sufficient.  I wish there were an easier way to tell Docker to just apply a slight modification to the default seccomp profile instead of having to replace it completely or disable it, but even doing `--security-opt seccomp:unconfined` is going to be more secure than `--privileged`.

---

For posterity, here's a copy of my relevant comment from that issue:

> Ok, here's what I've had success with: (where `default.json` comes from https://raw.githubusercontent.com/docker/docker-ce/v17.06.2-ce/components/engine/profiles/seccomp/default.json)
>
> ```diff
> --- default.json	2017-09-08 15:02:37.061507178 -0700
> +++ swift.json	2017-09-08 15:02:54.821989684 -0700
> @@ -216,6 +216,7 @@
>  				"open",
>  				"openat",
>  				"pause",
> +				"personality",
>  				"pipe",
>  				"pipe2",
>  				"poll",
> ```
> 
> ```console
> $ docker run -it --rm --security-opt seccomp=swift.json --cap-add sys_ptrace swift swift
> Welcome to Swift version 3.1 (swift-3.1-RELEASE). Type :help for assistance.
>   1>  
> ```

shahmishal:
@tianon I re-enabled issues on swift-docker, the main reason for disabling the issues was to avoid from having new issues created on GitHub.com. 

https://github.com/apple/swift-docker/issues/9#issuecomment-446785754

PR #1466, opened 2019-04-21
FIX: Cross-compile

formych:
Cross-compile your app inside the Docker container

I have tried to run this code， but it did not work. 
So I changed in my way.

### DIFF
--- a/golang/content.md
+++ b/golang/content.md
@@ -59,6 +59,7 @@ Alternatively, you can build for multiple platforms at once:
 $ docker run --rm -it -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%IMAGE%%:1.8 bash
 $ for GOOS in darwin linux; do
 >   for GOARCH in 386 amd64; do
+>     export GOOS GOARCH
 >     go build -v -o myapp-$GOOS-$GOARCH
 >   done
 > done


golang/content.md
```diff
@@ -59,7 +59,7 @@ Alternatively, you can build for multiple platforms at once:
 $ docker run --rm -it -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%IMAGE%%:1.8 bash
 $ for GOOS in darwin linux; do
 >   for GOARCH in 386 amd64; do
->     go build -v -o myapp-$GOOS-$GOARCH
+>     GOOS=$GOOS GOARCH=$GOARCH go build -v -o myapp-$GOOS-$GOARCH
```
tianon: Oh good catch, these need to be exported.  In the interest of shorter line length and making it more obvious what's happening here, I think I'd prefer to go for an explicit `export`, ala:

```suggestion
>     export GOOS GOARCH
>     go build -v -o myapp-$GOOS-$GOARCH
```

formych: Oh, what you said is right! But i think  this change the global env GOOS and GOARCH, maybe  we should unexport then in the end.

tianon: This is just an example, so I think that's best left as an exercise for the reader.

formych: Okey,  i write as u recommened, maybe i think too more, ^__^

tianon:
Awesome, thanks!

tianon:
I guess the simpler option here would've been to just use `bash -a` (https://www.gnu.org/software/bash/manual/bashref.html#The-Set-Builtin). 😅

PR #1468, opened 2019-04-26
update configuration file link

siyuan-li:
update configuration file link from https://raw.githubusercontent.com/containous/traefik/master/traefik.sample.toml to https://raw.githubusercontent.com/containous/traefik/v1.7/traefik.sample.toml

### DIFF
--- a/traefik/README.md
+++ b/traefik/README.md
@@ -61,7 +61,7 @@ Telling Traefik where your orchestrator is could be the *only* configuration ste
 
 # Example usage
 
-Grab a [sample configuration file](https://raw.githubusercontent.com/containous/traefik/master/traefik.sample.toml) and rename it to `traefik.toml`. Enable `docker` provider and web UI:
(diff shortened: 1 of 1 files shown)


tianon:
Sorry for the delay!  A couple issues I have with this:

1. it's a change to `README.md` (which is a generated file, per the comment at the top of it), so would need to move to `content.md`

2. the top of https://raw.githubusercontent.com/containous/traefik/master/traefik.sample.toml is pretty clear that it's for Traefik v2, and links to https://github.com/containous/traefik/blob/v1.7/traefik.sample.toml explicitly for Traefik v1 users

3. Traefik v2 *is* available and supported under https://hub.docker.com/_/traefik (even if it's still in Alpha releases), so I think it's reasonable for it to be preferred, but I'd leave that up to the Traefik maintainers to decide: cc @emilevauge @vdemeester @ldez @juliens @nmengin @mmatur @dduportal

PR #1484, opened 2019-05-26
php: optimize layers order

glensc:
optimize commands order that they can be cached if build context (project source) changes.

it might seem a bit illogical to use commands that don't exist yet, but the result is the same and avoids adding `WORKDIR` and `CMD` layers over and over when rebuilding with layers cache.

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -12,9 +12,9 @@ PHP is a server-side scripting language designed for web development, but which 
 
 ```dockerfile
 FROM %%IMAGE%%:7.2-cli
-COPY . /usr/src/myapp
 WORKDIR /usr/src/myapp
 CMD [ "php", "./your-script.php" ]
+COPY . .
 ```
 
 Then, run the commands to build and run the Docker image:


php/content.md
```diff
@@ -12,9 +12,9 @@ PHP is a server-side scripting language designed for web development, but which
 
 ```dockerfile
 FROM %%IMAGE%%:7.2-cli
-COPY . /usr/src/myapp
 WORKDIR /usr/src/myapp
 CMD [ "php", "./your-script.php" ]
+COPY . .
```
tianon: Given that this example is mostly illustrative (not a real example), that conceptually these make more sense in the other order, and that the difference from a `docker pull` perspective is negligible (it's all metadata, no extra layer tarballs), I'm not real keen on moving this all the way down, but I do definitely agree with moving it after `WORKDIR` so we don't have to repeat ourselves on the full path.

glensc: It is a real example, just rest of the commands are not included ;)

The pattern I'm following is to order least changed commands first unless they have an effect (can't do `USER` and then `RUN apt install`). Surprisingly zero length layers do take time when doing local rebuilds, so better have them cached.

so, I have directives like WORKDIR, EXPOSE before COPY commands:

```dockerfile
...
# assemble runtime image
# most often changed items came last to use layers cache more efficiently
FROM base

WORKDIR /app
ENV NODE_ENV production
ENV PORT 32645
EXPOSE $PORT
CMD ["node_modules/.bin/forever", "forever.json"]
USER node

COPY --from=build /app .
```

while typically such dockerfile is written as `COPY` being first layer, and then `ENV`, `EXPOSE`, `CMD` and `USER`.

the only problem I see, that it might be not very logical to specify `CMD` while the files are not yet copied to the image :)

PR #1486, opened 2019-05-28
Fixed instructions in Redis container docs

sbkok:
Updated docs as the current instructions are not working.

Added instructions to:
* Connect to a redis server running in docker from the local machine.
* Connect with `redis-cli` to a redis server when both run in docker using docker networks.

### DIFF
--- a/redis/content.md
+++ b/redis/content.md
@@ -14,6 +14,12 @@ Redis is an open-source, networked, in-memory, key-value data store with optiona
 $ docker run --name some-redis -d %%IMAGE%%
 ```
 
+Connect to it locally on your machine:
+
+```console
+$ docker run --name some-redis -p 127.0.0.1:6379:6379 -d %%IMAGE%%
+```
+
 ## start with persistent storage
 
 ```console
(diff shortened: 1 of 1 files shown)


redis/content.md
```diff
@@ -14,6 +14,12 @@ Redis is an open-source, networked, in-memory, key-value data store with optiona
 $ docker run --name some-redis -d %%IMAGE%%
 ```
 
+Connect to it locally on your machine:
+
+```console
+$ docker run --name some-redis -p 127.0.0.1:6379:6379 -d %%IMAGE%%
+```
```
tianon: I'd rather not suggest this to users, since this is really something that should be easily inferred by someone familiar with Docker and/or how `-p` works (and we'd prefer to keep the documentation somewhat succinct and specific to Redis instead of a general guide to using Docker).

redis/content.md
```diff
@@ -26,10 +32,24 @@ For more about Redis Persistence, see [http://redis.io/topics/persistence](http:
 
 ## connecting via `redis-cli`
 
+Create the network first:
+
 ```console
-$ docker run -it --network some-network --rm %%IMAGE%% redis-cli -h some-redis
+$ docker network create some-network
+```
```
tianon: Along similar lines, the `docker network create` bits were intentionally left out of https://github.com/docker-library/docs/pull/1441 (as implied by the title) -- we'd prefer not to document Docker basics here and instead provide more reference/tips for using Redis with this image.

sbkok:
Thanks for the review @tianon. I agree that we should not completely duplicate the instructions as provided by the docker docs, however, at the moment the instructions in the doc make it look like it should have worked if you first run the server followed by running the redis cli.

In my opinion, we should at least point the user in the right direction by referencing the docs on how to set it up if you don’t want to include step-by-step instructions. What do you think about this?

PR #1500, opened 2019-06-08
Fix broken link to Docker documentation (--device)

lindhe:
The link to the official Docker documentation of the `docker run` command was broken. Also, I took the liberty to update it so that it points specifically to the section about the `--device` option.


yosifkit:
Thanks @lindhe!

LGTM. cc @ruffsl in case you use this link elsewhere in OSRF.

PR #1503, opened 2019-06-12
Remove "Dockerfile path" from tags listing

tianon:
For context, this was added initially because the previous rendering of Markdown on the Hub was such that code in a link would render without underline and barely noticeable blue color (and no hover behavior either besides the normal cursor change), so it wasn't clear it was even a link without additional non-code text. Now that that has been corrected/adjusted, I think it's totally reasonable to remove those (since they're included in the link already).

For reference/comparison, this takes the `open-liberty` readme from ~39k characters down to ~33k characters.

See https://github.com/docker-library/official-images/issues/6030 for discussion.


tianon:
More precisely, `open-liberty` goes from 39263 down to 33809.

yosifkit:
Should line 66 be changed as well for Shared Tags?

https://github.com/docker-library/docs/blob/c87a68de3ea202ab548124afae051649d0453756/.template-helpers/generate-dockerfile-links-partial.tmpl#L65-L67

tianon:
Doh, yep good catch -- update pushed.

PR #1505, opened 2019-06-17
Update Bonita doc for 7.9

JeremJR:
related to https://github.com/docker-library/official-images/pull/6097

### DIFF
--- a/bonita/content.md
+++ b/bonita/content.md
@@ -167,9 +167,9 @@ The `-v /my/own/datadir:/opt/bonita` part of the command mounts the `/my/own/dat
 
 		```console
 		cd bonita_migration
-		wget https://release.ow2.org/bonita/bonita-migration-distrib-2.40.0.zip
+		wget https://release.ow2.org/bonita/bonita-migration-distrib-2.41.0.zip
 		wget https://download.forge.ow2.org/bonita/BonitaBPMCommunity-7.2.4-Tomcat-7.0.67.zip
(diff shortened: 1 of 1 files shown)


bonita/content.md
```diff
@@ -258,22 +258,26 @@ The `-v /my/own/datadir:/opt/bonita` part of the command mounts the `/my/own/dat
...
 
 	```console
-	$ docker restart bonita_7.8.4_postgres
+	$ docker restart bonita_7.9.0_postgres
 	```
 
-For more details regarding Bonita migration, see the [documentation](https://documentation.bonitasoft.com/bonita/7.8/migrate-from-an-earlier-version-of-bonita-bpm).
+-	Specific consideration regarding migration to Java 11 in Bonita 7.9
+
+	Bonita 7.9 docker image runs with Java 11. If you are migrating from an earlier version which runs Java 7.8 you should validate on a test environment that your custom code is compatible. Aside from just code incompatibility, special attention has to be given to the dependencies of the custom code, as they might either not work in Java 11.
```
yosifkit:
```suggestion
	Bonita 7.9 docker image runs with Java 11. If you are migrating from an earlier version which runs Java 8, you should validate on a test environment that your custom code is compatible. Aside from just code incompatibility, special attention has to be given to the dependencies of the custom code, as they might not work in Java 11.
```

JeremJR:
thank you @yosifkit

PR #1506, opened 2019-06-19
SONAR-12067 Update docker image documentation about graceful shutdown

julienlancelot:


tianon:
Instead of documenting flags for `docker stop`, I'd suggest you recommend that users add `--stop-timeout` on their container, especially since there are a number of different situations that might try to stop a container where `--stop-timeout` will apply (but a `--time` flag on `docker stop` won't because the stop wasn't manual).

julienlancelot:
Thanks @tiangolo for your feedback, we're gonna see how to handle that.

dbmeneses:
@tiangolo I've updated the P/R based on your feedback.

tiangolo:
I think you intended to tag @tianon, not me. But I'll take it as a compliment :smile:

sonarqube/content.md
```diff
@@ -92,6 +92,16 @@ $ docker build --tag=sonarqube-custom .
 $ docker run -ti sonarqube-custom
 ```
 
+### Avoid hard termination of SonarQube
+
+Starting from SonarQube 7.8, SonarQube stops gracefully, waiting for any tasks in progress to finish.
+Waiting for in-progress tasks to finish can take a large amount of time which the docker does not expect by default when stopping.
+To avoid having the SonarQube instance killed by the Docker daemon after 10 seconds, it is best to configure a timeout to stop the container with `--stop-timeout`. For example:
```
yosifkit: The travis test that runs the `markdownfmt` check needs these three lines to be combined to a single line. The other travis error about a changed `README.md` is false and so no action need to be taken for that.

```diff
diff ./sonarqube/content.md markdownfmt/./sonarqube/content.md
--- /tmp/markdownfmt986287976
+++ /tmp/markdownfmt671901095
@@ -94,9 +94,7 @@
 
 ### Avoid hard termination of SonarQube
 
-Starting from SonarQube 7.8, SonarQube stops gracefully, waiting for any tasks in progress to finish.
-Waiting for in-progress tasks to finish can take a large amount of time which the docker does not expect by default when stopping.
-To avoid having the SonarQube instance killed by the Docker daemon after 10 seconds, it is best to configure a timeout to stop the container with `--stop-timeout`. For example:
+Starting from SonarQube 7.8, SonarQube stops gracefully, waiting for any tasks in progress to finish. Waiting for in-progress tasks to finish can take a large amount of time which the docker does not expect by default when stopping. To avoid having the SonarQube instance killed by the Docker daemon after 10 seconds, it is best to configure a timeout to stop the container with `--stop-timeout`. For example:
```

dbmeneses: Alright, fixed.

sonarqube/content.md
```diff
@@ -92,6 +92,14 @@ $ docker build --tag=sonarqube-custom .
 $ docker run -ti sonarqube-custom
 ```
 
+### Avoid hard termination of SonarQube
+
+Starting from SonarQube 7.8, SonarQube stops gracefully, waiting for any tasks in progress to finish. Waiting for in-progress tasks to finish can take a large amount of time which the docker does not expect by default when stopping. To avoid having the SonarQube instance killed by the Docker daemon after 10 seconds, it is best to configure a timeout to stop the container with `--stop-timeout`. For example:
+
+```console
+docker run --stop-timeout 3600
```
yosifkit: Did you want a `%%IMAGE%%` at the end like line 16? (it'll get autofilled when README.md is generated)

pierre-guillot-sonarsource: Yes. I added it, among other changes

tianon:
Travis is confused, merging

PR #1507, opened 2019-06-19
adoptopenjdk: add logo + formatting

gdams:


gdams:
thanks for landing this, I've seen that the doc is now updated at https://hub.docker.com/_/adoptopenjdk but the main logo at the top is still the default one. Do I need to change something else?

gdams:
If we have a choice it would be good to add https://adoptopenjdk.net/dist/assets/social-logo.png

yosifkit:
Unfortunately the logo at the top is not controlled by us. I'll try poking someone at Docker Inc to see if they can tell the right people.

gdams:
Thanks @yosifkit! It would be great to get the logo added! :)

PR #1514, opened 2019-06-26
Adjust long-description trimming to prefer trimming tag listing

tianon:
Instead of cutting off the end of the image documentation, this change makes the code prefer to trim the entire "Supported tags" list, which for all our images which currently go over the limit brings them down enough to be within.  The note about where to get the full `README.md` file (with the "Supported tags" list intact) still stands, although could maybe do with some better styling and/or wording to make it more obvious (TBD).

I also verified that this change does the right thing in the fallback case by making the `postgres` image description intentionally extra long and verified that it simply trims the description content as before (since taking off the tags doesn't make the description short enough to be under the limit in that case).

Refs https://github.com/docker-library/official-images/issues/6030, https://github.com/docker/hub-feedback/issues/238

### DIFF
--- a/push.pl
+++ b/push.pl
@@ -4,7 +4,7 @@ use warnings;
 use 5.010;
 use open ':encoding(utf8)';
 
-use File::Basename qw(fileparse);
+use File::Basename qw(basename fileparse);
 use File::Temp;
 use Getopt::Long;
 use Mojo::UserAgent;
@@ -14,7 +14,7 @@ use Term::UI;
 use Term::ReadLine;
 
 my $hubLengthLimit = 25_000;
-my $githubBase = 'https://github.com/docker-library/docs/tree/master';
+my $githubBase = 'https://github.com/docker-library/docs/tree/master'; # TODO point this at the correct "dist-xxx" branch based on "namespace"
 
 my $username;
 my $password;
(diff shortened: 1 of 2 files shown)


tianon:
FYI @gdams this should help make `adoptopenjdk`'s description trimming more friendly :+1:

push.pl
```diff
@@ -81,17 +81,32 @@ sub prompt_for_edit {
 		# TODO https://github.com/docker/hub-beta-feedback/issues/238
 		my $fullUrl = "$githubBase/$proposedFile";
 		my $note = "**Note:** the description for this image is longer than the Hub length limit of $lengthLimit, so has been trimmed.  The full description can be found at [$fullUrl]($fullUrl).  See [docker/hub-beta-feedback#238](https://github.com/docker/hub-beta-feedback/issues/238) for more information.\n\n";
-		$proposedText = $note . substr $proposedText, 0, ($lengthLimit - length($note));
+		
+		# if our text is too long for the Hub length limit, let's first try removing the "Supported tags" list and add our note and see if that's enough to let us put the full image documentation
+		my $trimmedText = $proposedText;
+		$trimmedText =~ s%^# Supported tags and respective `Dockerfile` links\n.*?\n(?=# |\[)%$note%ms;
```
tianon: Perhaps we just change the note for this version to include an extra blurb explicitly pointing out where to find the now replaced "Supported tags and respective `Dockerfile` links" section?

tianon: I suppose it could include an even more targeted link directly to the anchor, ala https://github.com/docker-library/docs/blob/master/adoptopenjdk/README.md#supported-tags-and-respective-dockerfile-links :thinking:

gdams:
Thanks for your work on this @tianon :)

push.pl
tianon: Updated with a more specific/relevant note for this case (and it now keeps the heading). :+1:

tianon:
image

PR #1519, opened 2019-07-03
Add/varnish

gquintard:
companion PR to https://github.com/docker-library/official-images/pull/5855, supersedes #1241
Thanks go to @teohhanhui for the initial work. I clipped down the vmods section as the instructions vary a bit depending on the vmod author, so I want to unify this first before exposing "easy" instructions to the world


varnish/content.md
```diff
@@ -0,0 +1,49 @@
...
+
+backend default {
+  .host = "www.nytimes.com:80";
+}
+```
+
+Then run:
+
+```console
+$ docker run --name my-running-varnish -v /path/to/default.vcl:/usr/local/etc/varnish/default.vcl:ro --tmpfs /usr/local/var/varnish:exec -d %%IMAGE%%
```
yosifkit: Should this be `/etc/varnish/default.vcl` like the Dockerfile example and the default `CMD` of the container?

```Dockerfile
CMD ["varnishd", "-F", "-f", "/etc/varnish/default.vcl"]
```

gquintard: it should, good catch. fixed

gquintard:
> do you want to mention that if users replace the CMD, they need to make sure to use -F?

You tell me :-) My assumption was that once you start changing `CMD`, you are on your own and should know what you are doing. But if you think the comment is warranted, I can add it.

> You could also wait and adjust the entrypoint at some point to ensure -F is in the arguments to varnishd (though might be a little complex).

I've hack my share of "magic" interfaces that do crazy things for "ease of use", and they very consistently exploded in very subtle ways. I'd rather keep things dumb and debuggable (can you tell I'm an Arch user??)

tianon:
(Travis is confused and thinks `README.md` has changed here, which is not true.)

PR #1525, opened 2019-07-09
docker: add a note about TLS

AkihiroSuda:
TCP connection without TLS is completely insecure and can easily result
in container breakout.

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>


AkihiroSuda:
@tianon @justincormack @thaJeztah PTAL

justincormack:
I think we should remove this default from the `dind` containers.

tianon:
@justincormack if we remove the default, the container will default to unix-socket only, which makes for a pretty irritating user experience from a container :disappointed:  Any ideas for how to resolve that?

AkihiroSuda:
@tianon 
> Any ideas for how to resolve that?

My suggestion is to let dind create TLS files on start up
https://github.com/docker-library/docker/issues/164

```console
$ docker run -d --privileged --name some-docker \
  -e DIND_TLS_SAN=DNS:docker,DNS:localhost docker:dind
$ docker cp some-docker:/certs/for-client ./certs-for-client
$ docker run --rm --link some-docker:docker -v $(pwd)/certs-for-client:/root/.docker \
  -e DOCKER_HOST=tcp://docker:2376 docker info
```

AkihiroSuda:
Can we get this merged?

docker/content.md
```diff
@@ -19,17 +19,26 @@ If you are still convinced that you need Docker-in-Docker and not just access to
...
+$ docker run --privileged --name some-docker -d -v /daemon-secret:/secret %%IMAGE%%:dind \
+ dockerd -H tcp://0.0.0.0:2376 --tlsverify --tlscacert /secret/ca.pem --tlscert /secret/cert.pem --tlskey /secret/key.pem
 ```
 
 **Note:** `--privileged` is required for Docker-in-Docker to function properly, but it should be used with care as it provides full access to the host environment, as explained [in the relevant section of the Docker documentation](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities).
 
-By default, the `dind` variants of this image add `--host=tcp://0.0.0.0:2375` (on top of the explicit default of `--host=unix:///var/run/docker.sock`) in order to allow external containers to access `dockerd` appropriately (as the following examples illustrate). If you use `--network=host` or other methods of sharing network namespaces (such as Kubernetes pods, for example), this might be a security issue. To disable this image behavior, simply override the container command or entrypoint to run `dockerd` directly (`... docker:dind dockerd ...` or `... --entrypoint dockerd docker:dind ...`).
```
tianon: My preference is to make this existing paragraph a little bit stronger (explicitly recommending TLS with some clues how to use it and making it more clear that this is a potential security issue with more than just `--network=host`) in addition to implementing https://github.com/docker-library/docker/issues/164 for 19.03+:

> **Warning:** by default, the `dind` variants of this image add `--host=tcp://0.0.0.0:2375` (on top of the explicit default of `--host=unix:///var/run/docker.sock`) in order to allow other containers to access `dockerd` (as the following examples illustrate). If you use `--network=host`, shared network namespaces (as in Kubernetes pods), or otherwise have network access to the container (including containers started within the `dind` instance via their gateway interface), this is a potential security issue (which can lead to access to the host system, for example). To disable this image behavior, simply override the container command or entrypoint to run `dockerd` directly (`... docker:dind dockerd ...` or `... --entrypoint dockerd docker:dind ...`). It is recommended to implement TLS (`... docker:dind dockerd --host tcp://0.0.0.0:2376 --tlsverify ...`) if network access to the `dind` instance is required.

AkihiroSuda: updated

tianon: I meant I'd rather adjust *just* this paragraph for now -- especially given how much simpler https://github.com/docker-library/docker/pull/166 will make enabling TLS (so that the docs updates don't need to be quite so extensive).

AkihiroSuda: updated

PR #1528, opened 2019-07-10
emphasis on recommended solutions

alcohol:


composer/content.md
```diff
@@ -91,42 +92,42 @@ Our image is aimed at quickly running Composer without the need for having a PHP
...
 	  "config": {
 	    "platform": {
-	      "php": "7.1.3"
+	      "php": "MAJOR.MINOR.PATCH"
 	    }
 	  }
 	}
 	```
 
+-	use [`--ignore-platform-reqs`](https://getcomposer.org/doc/03-cli.md#install-i):
```
derrabus: Can we add a warning? Something like "This option is potentially harmful. You might get dependencies that are incompatible with your target environment."

composer/content.md
```diff
@@ -91,42 +92,42 @@ Our image is aimed at quickly running Composer without the need for having a PHP
...
 Suggestions:
 
+-	(recommended) create your own build image and [install](https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md) Composer inside it.
+
+	**Note:** Docker 17.05 introduced [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/), simplifying this enormously:
+
+	```dockerfile
+	COPY --from=%%IMAGE%% /usr/bin/composer /usr/bin/composer
+	```
+
```
derrabus: `config.platform` should work for extensions as well. Maybe we can add this to the list of suggestions?

composer/content.md
alcohol: Labeled the options in a suggestive manner. Good enough?

exussum12:
Looks perfect. Great work!

composer/content.md
derrabus: I'd still prefer a warning, but your solution works for me. Thank you!

composer/content.md
```diff
@@ -23,7 +23,8 @@ $ docker run --rm --interactive --tty \
 You can bind mount the Composer home directory from your host to the container to enable a persistent cache or share global configuration:
 
 ```console
-$ docker run --rm --interactive --tty \
+$ COMPOSER_HOME=$HOME/.composer \
+  docker run --rm --interactive --tty \
   --volume $PWD:/app \
   --volume $COMPOSER_HOME:/tmp \
```
tianon: This PR looks great except this bit -- I don't think this does what you want/intend:

```console
$ FOO=bar echo $FOO

$ FOO=bar; echo $FOO
bar
```

Perhaps you really intended for something like `--volume ${COMPOSER_HOME:-$HOME/.composer}:/tmp`?

alcohol: You are right

PR #1532, opened 2019-07-14
Replaced php logo with a squared one

SerhiiCho:
The problem is that php logo is not a square and therefore looks incorrect on [docker hub php page](https://hub.docker.com/_/php). 

### Before
DeepinScreenshot_select-area_20190714212437
### After
DeepinScreenshot_select-area_20190714212852


tianon:
Unfortunately, that copy of the logo is not part of this repository -- it is maintained by the Hub team directly.  We're happy to (and have offered to them to) regenerate or resize our logos in an automated way via ImageMagick or similar and even would be happy to implement an API to update them on the Hub directly, but we don't currently have that.  (In other words, updating our logo will not fix the logo at the top of the page which is the problem this is trying to solve, so I'm -1 on this change.)

cc @programmerq (perhaps you'd be willing to prod internally and see if there's anyone with some sympathy for PHP?) :pray: :smile: :heart:

tianon:
This is now fixed! :+1: :tada:

SerhiiCho:
> This is now fixed! +1 tada

That's awesome. I don't know why the wrong version was gross to me :elephant:

PR #1538, opened 2019-07-19
Revamp "docker" image docs, especially for TLS

tianon:
This is the companion to https://github.com/docker-library/docker/pull/166. :+1:

### DIFF
--- a/docker/content.md
+++ b/docker/content.md
@@ -18,9 +18,9 @@ If you are still convinced that you need Docker-in-Docker and not just access to
 
 ## TLS
 
-Starting in 18.09+, the `dind` variants of this image will automatically generate TLS certificates the directory specified by the `DOCKER_TLS_CERTDIR` environment variable.
+Starting in 18.09+, the `dind` variants of this image will automatically generate TLS certificates in the directory specified by the `DOCKER_TLS_CERTDIR` environment variable.
 
-**Warning:** in 18.09, this behavior is disabled by default (for compatibility) but can be enabled by setting the variable to an appropriate value (`-e DOCKER_TLS_CERTDIR=/certs` or similar). In 19.03+, this behavior is enabled by default.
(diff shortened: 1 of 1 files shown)


docker/content.md
```diff
@@ -16,127 +16,176 @@ If you are still convinced that you need Docker-in-Docker and not just access to
 
 [![asciicast](https://asciinema.org/a/24707.png)](https://asciinema.org/a/24707)
 
+## TLS
+
+Starting in 18.09+, the `dind` variants of this image will automatically generate TLS certificates the directory specified by the `DOCKER_TLS_CERTDIR` environment variable.
```
yosifkit:
```suggestion
Starting in 18.09+, the `dind` variants of this image will automatically generate TLS certificates in the directory specified by the `DOCKER_TLS_CERTDIR` environment variable.
```

docker/content.md
```diff
@@ -16,127 +16,176 @@ If you are still convinced that you need Docker-in-Docker and not just access to
...
+
+**Warning:** in 18.09, this behavior is disabled by default (for compatibility) but can be enabled by setting the variable to an appropriate value (`-e DOCKER_TLS_CERTDIR=/certs` or similar). In 19.03+, this behavior is enabled by default.
```
yosifkit: Maybe keep the previous security warning:
```suggestion
**Warning:** in 18.09, this behavior is disabled by default (for compatibility). If you use `--network=host`, shared network namespaces (as in Kubernetes pods), or otherwise have network access to the container (including containers started within the `dind` instance via their gateway interface), this is a potential security issue (which can lead to access to the host system, for example). It is recommended to enable TLS by setting the variable to an appropriate value (`-e DOCKER_TLS_CERTDIR=/certs` or similar). In 19.03+, this behavior is enabled by default.
```

PR #1539, opened 2019-07-23
Add a note about HOME wrt wordpress:cli

tianon:
https://github.com/docker-library/wordpress/issues/417#issuecomment-514358315 :sweat_smile:


tianon:
Without this, sometimes `wp` will try to write to `/etc/X11/fs/.wp-cli/cache` (since `/etc/X11/fs` is `HOME` for UID 33 in Alpine).

PR #1543, opened 2019-07-26
improve sonarqube documentation about embedded elasticsearch requirements

pierre-guillot-sonarsource:


pierre-guillot-sonarsource:
Hello team, looks like the travis job is red, i can't figure out why :
`The command ".travis/check-markdownfmt.sh" exited with 1.`

I can't run this script on MacOS because of `readlink` :
`readlink: illegal option -- f`

pretty hard to have the check green, then :(

tianon:
@pierre-guillot-sonarsource I'd recommend running the `markdownfmt.sh` script at the root of this repository instead -- its usage is very similar to `gofmt` (`./markdownfmt.sh -d foo/bar.md` to see a diff, `./markdownfmt.sh -w foo/bar.md` to apply all suggested changes automatically)

Also, the output in the Travis logs is a normal unified diff, so technically should be applyable directly via something like `patch -p1`.

pierre-guillot-sonarsource:
Thanks @tianon , `./markdownfmt.sh -w foo/bar.md` did worked. Travis is now green :)

PR #1547, opened 2019-08-01
Add a short blurb linking to the "dind-rootless" PR

tianon:
See https://github.com/docker-library/docker/pull/174.

Given that this is an experimental feature, I think it's reasonable to link to the PR itself as documentation (I've marked as "resolved" several distracting discussions there so the discussion and examples are easy to follow).


tianon:
(This depends on https://github.com/docker-library/official-images/pull/6400)

PR #1549, opened 2019-08-07
Updated mysql content.md

poldridge:
This is a convenient way to delay starting other services until MySQL is accepting connections, when using docker-compose v2.1 or higher.


mysql/content.md
```diff
@@ -152,6 +152,24 @@ If there is no database initialized when the container starts, then a default da
...
+    environment:
+      MYSQL_ROOT_PASSWORD: example
+    healthcheck:
+      test: 'mysql -e SELECT\ 1'
+                                                                                 
+  app:
+    ...
+    depends_on:
+      mysql:
+        condition: service_healthy
```
tianon: Given the caveats at https://docs.docker.com/compose/compose-file/#depends_on, especially the following one, I'm wary of adding anything that references `depends_on:` in the documentation:

> - Version 3 no longer supports the `condition` form of `depends_on`.
> 
> - The `depends_on` option is ignored when deploying a stack in swarm mode with a version 3 Compose file.

poldridge: That's a good point, this only works with compose file formats 2.1 - 2.4. I think it is worth sharing, but this probably isn't the best way to do that.

PR #1550, opened 2019-08-10
perl: Document PID 1 signal handling behavior

zakame:
Describe how Perl behaves when being run as PID 1 under containers,
particularly on receiving SIGTERM/SIGINT, showing example for explicit
signal handling.

Ref Perl/docker-perl#44

### DIFF
--- a/perl/content.md
+++ b/perl/content.md
@@ -32,6 +32,32 @@ For many simple, single file projects, you may find it inconvenient to write a c
 $ docker run -it --rm --name my-running-script -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%IMAGE%%:5.20 perl your-daemon-or-script.pl
 ```
 
+## Signal handling behavior notice
+
+As Perl will run as PID 1 by default in containers (unless an [ENTRYPOINT](https://docs.docker.com/engine/reference/builder/#entrypoint) is set,) special care needs to be considered when expecting to send signals (particularly SIGINT or SIGTERM) to it. For example, running
+
+```console
+$ docker run -it --name sleeping_beauty --rm %%IMAGE%%:5.20 perl -E 'sleep 300'
+```
+
+and doing on another terminal,
+
+```console
+$ docker exec sleeping_beauty kill 1
+```
+
+will *not* stop the perl running on the `sleeping_beauty` container (it will keep running until the `sleep 300` finishes.) To do so, one must set a signal handler like this:
+
+```console
+$ docker run -it --name quick_nap --rm %%IMAGE%%:5.20 perl -E '$SIG{TERM} = sub { say "recv TERM" }; sleep 300'
+```
+
+so doing `docker exec quick_nap kill 1` (or the simpler `docker stop quick_nap`) will immediately stop the container, and print `recv TERM` in the other terminal.
+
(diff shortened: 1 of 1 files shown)


perl/content.md
```diff
@@ -32,6 +32,32 @@ For many simple, single file projects, you may find it inconvenient to write a c
...
+$ docker exec kill-me-plz kill 1
+```
+
+will *not* stop the perl running on the `kill-me-plz` container (it will keep running until the `sleep 300` finishes.) To do so, one must set a signal handler like this:
+
+```console
+$ docker run -it --name kill-me-now --rm %%IMAGE%%:5.20 perl -E '$SIG{TERM} = sub { say "recv TERM" }; sleep 300'
+```
+
+so doing `docker exec kill-me-now kill 1` (or the simpler `docker stop kill-me-now`) will immediately stop the container, and print `recv TERM` in the other terminal.
```
yosifkit: Just two mior comments: maybe not name the containers about killing, I don't want to unintentionally upset or hurt someone; use `docker kill -s TERM some-perl` instead of `docker exec kill-me-now kill 1`.

zakame: Thanks, updated with more benign names :muscle:

perl/content.md
```diff
@@ -32,6 +32,32 @@ For many simple, single file projects, you may find it inconvenient to write a c
...
+$ docker exec sleeping_beauty kill 1
+```
+
+will *not* stop the perl running on the `sleeping_beauty` container (it will keep running until the `sleep 300` finishes.) To do so, one must set a signal handler like this:
+
+```console
+$ docker run -it --name quick_nap --rm %%IMAGE%%:5.20 perl -E '$SIG{TERM} = sub { say "recv TERM" }; sleep 300'
+```
+
+so doing `docker exec quick_nap kill 1` (or the simpler `docker stop quick_nap`) will immediately stop the container, and print `recv TERM` in the other terminal.
```
tianon: IMO this is kind of misleading -- your `$SIG{TERM}` function doesn't kill the program, but the program stops gracefully given the detail of `sleep` being interrupted by signals, right?  In other words, even `SIGUSR1` should "stop" this particular program, right?

I think that behavior is more obvious if the program is changed to `$SIG{TERM} = sub { say "recv TERM" }; sleep 300 while 1` -- then every `SIGTERM` received simply prints `recv TERM` and the program continues.  Maybe the example `sub` should also `exit`/`die`?

zakame: No, there's no corresponding `$SIG{USR1}` handler explicitly set (nor an entrypoint handling it as well,) so `quick_nap` will still sleep even when USR1 is received.

Yes, using `die` instead of `say` could be more appropriate.

tianon: Oh interesting, the docs for `sleep` are a little misleading on that point 😅

(But I think you catch my point that the current wording implies it's the handler that's killed the container and in this case it isn't 😇)

zakame: Yep, currently the handler simply interrupts the `sleep` and since it doesn't go back, perl just stops.  It'd be more clear if something else followed after the sleep, like

```perl
$SIG{INT} = sub { $sig++; say "recv INT" }; sleep 300; say "waking up" if $sig
```

zakame: Updated the example and description to be more clear, thanks!

tianon:
👍🤘

PR #1552, opened 2019-08-10
new example for configuring redis-server

mostafahussein:
As redis-server has support for adding the configuration using `stdin` according to the help below, I thought it would be helpful for others to get the idea about how to use it with docker in case someone wanted to configure redis without building a custom image or mounting a configuration file.
Feel free to update the content as you see fit or let me know what needs to be done so I can update it.

Thanks
```
Usage:
       ./redis-server - (read config from stdin)
...
```

### DIFF
--- a/redis/content.md
+++ b/redis/content.md
@@ -48,12 +48,12 @@ $ docker run -v /myredis/conf/redis.conf:/usr/local/etc/redis/redis.conf --name 
 
 Where `/myredis/conf/` is a local directory containing your `redis.conf` file. Using this method means that there is no need for you to have a Dockerfile for your redis container.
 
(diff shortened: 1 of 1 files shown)


yosifkit:
There is no need to go through all the work to pass a bunch of environment values as a file to stdin.

> Since Redis 2.6 it is possible to also pass Redis configuration parameters using the command line directly.
>
> \- https://redis.io/topics/config#passing-arguments-via-the-command-line

We show this above in the ["start with persistent storage"](https://github.com/docker-library/docs/blob/e3b9843c41f31b147c0dbcd4f80288500b2a7276/redis/content.md#start-with-persistent-storage): 

```console
$ docker run --name some-redis -d redis redis-server --appendonly yes
```

mostafahussein:
Thanks for the heads up ! I missed the point that all parameters can be passed through the command line options not for specific configs

mostafahussein:
@yosifkit do you mind if i updated the example and reopened this PR ?

PR #1557, opened 2019-08-16
sonarqube: Add Docker Secrets

adgerson:


tianon:
cc `sonarqube` image maintainers @janos-ss @julienlancelot @christophelevis @pierre-guillot-sonarsource @michal-duda-sonarsource (can we get some review here, please?)

pierre-guillot-sonarsource:
Hello @tianon, i guess this is related to a PR opened on our side as well : https://github.com/SonarSource/docker-sonarqube/pull/306

As we do not know if this is the correct way of using docker secrets, we do not plan to accept this PR for now.

tianon:
Ah, makes sense; thanks for adding the missing context!

In that case, let's reconsider if/when that functionality (or something else like it) makes it to https://github.com/docker-library/official-images. :+1:

PR #1567, opened 2019-09-08
[PHP] Document default extensions

J0WI:
https://github.com/friendica/docker/pull/78#issuecomment-529210776

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -49,7 +49,7 @@ RUN docker-php-source extract \
 
 ### PHP Core Extensions
 
-For example, if you want to have a PHP-FPM image with `iconv` and `gd` extensions, you can inherit the base image that you like, and write your own `Dockerfile` like this:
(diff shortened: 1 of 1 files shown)


php/content.md
```diff
@@ -66,6 +65,45 @@ Remember, you must install dependencies for your extensions manually. If an exte
 
 See ["Dockerizing Compiled Software"](https://tianon.xyz/post/2017/12/26/dockerize-compiled-software.html) for a description of the technique Tianon uses for determining the necessary build-time dependencies for any bit of software (which applies directly to compiling PHP extensions).
 
+### Default extensions
+
+Some extensions are compiled by default. As of PHP 7.3 these are:
```
tianon: Is there a command we could suggest users run instead to get this list for themselves?

J0WI: I got the list with `php -m`

yosifkit: I'd rather document the command instead of the list; this list will too easily be out of date.

J0WI: Done.

PR #1569, opened 2019-09-10
Update README.md

archlemon:


archlemon:
Oh, diff is showing only the line change.

`then a putting a connect-retry` I removed the `a` after `then` in the first sentence.

Issue #1572, opened 2019-09-13
Prefer os tags in How to use this Image sections


dogweather:
**Problem:** There's a mis-match between the description sections and how the images are actually managed. This creates unrealistic expectations by users. 

We're telling users, e.g.:

> ruby:version
> This is the defacto image. If you are unsure about what your needs are, you probably want to use this one. It is designed to be used both as a throw away container (mount your source code and start the container to start your app), as well as the base to build other images off of.

However, that simplified tag doesn't reference a stable or immutable image. E.g., recently many "older" images were updated to base off newer or different OS's. 

And so, a mountain of help docs, tutorials, etc. all advise to use that simple label type. And all the mysql-based ones [are broken now](https://stackoverflow.com/questions/57048428/e-package-mysql-client-has-no-installation-candidate-in-php-fpm-image-build-u). Since many people followed this advice, many production and development systems are broken as well.

**Proposed Solution:** All examples and instructions should use fully qualified labels such as `ruby:2.6.3-buster` instead of the currently advised style `ruby:2.6.3`.

wglambert:
We do have a section in the faq about tags https://github.com/docker-library/faq#whats-the-difference-between-shared-and-simple-tags

Maybe we should make a link to that, currently there's no mention of the faq in any image documentation. I think the only link we have to it is in the [official-images](https://github.com/docker-library/official-images/#more-faqs) repo

wglambert:
Relevant to https://github.com/docker-library/official-images/issues/6282

tianon:
We also do already have a blurb about these OS-specific variants, but we should probably improve it and move it up to be more prominent: (https://github.com/docker-library/docs/blob/b6419071dbd29a7cf1f693edc102fe0c8443e394/ruby/README.md#rubyversion)

> ## `ruby:<version>`
> 
> This is the defacto image. If you are unsure about what your needs are, you probably want to use this one. It is designed to be used both as a throw away container (mount your source code and start the container to start your app), as well as the base to build other images off of.
> 
> This tag is based off of [`buildpack-deps`](https://hub.docker.com/_/buildpack-deps/). `buildpack-deps` is designed for the average user of Docker who has many images on their system. It, by design, has a large number of extremely common Debian packages. This reduces the number of packages that images that derive from it need to install, thus reducing the overall size of all images on your system.
> 
[...]

PR #1574, opened 2019-09-13
Add a link to our FAQ entry about Shared Tags when we have them

tianon:
> # Supported tags and respective `Dockerfile` links
>
> ## Simple Tags
>
> ...
>
> ## Shared Tags
>
> ...
>
> (See also ["What's the difference between 'Shared' and 'Simple' tags?" in the FAQ](https://github.com/docker-library/faq#whats-the-difference-between-shared-and-simple-tags).)

Refs https://github.com/docker-library/official-images/issues/6282, https://github.com/docker-library/docs/issues/1572#issuecomment-531366628

### DIFF
--- a/.template-helpers/generate-dockerfile-links-partial.tmpl
+++ b/.template-helpers/generate-dockerfile-links-partial.tmpl
@@ -8,7 +8,10 @@ This template defines the "Supported tags and Dockerfile links" portion of an im
 
 {{- $sharedTagGroups := .Manifest.GetSharedTagGroups -}}
 {{- if (len $sharedTagGroups) -}}
-	{{- "## Simple Tags\n\n" -}}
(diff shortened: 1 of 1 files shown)


tianon:
Maybe we should move it up before "Simple Tags" instead so it's more obvious?

yosifkit:
> Maybe we should move it up before "Simple Tags" instead so it's more obvious?

Yeah, I think so. When there are shared tags there are usually quite a few tags and users may miss it when scrolling past them.

tianon:
Updated:

> # Supported tags and respective `Dockerfile` links
>
> (See ["What's the difference between 'Shared' and 'Simple' tags?" in the FAQ](https://github.com/docker-library/faq#whats-the-difference-between-shared-and-simple-tags).)
>
> ## Simple Tags
>
> ...
>
> ## Shared Tags
>
> ...

PR #1585, opened 2019-10-11
Add a separate explicit message for architectures which are not supported for an image

tianon:
This is to handle explicitly the case of images that were supported once but are no longer supported (like "openjdk" on s390x and ppc64le).

Ref https://github.com/docker-library/openjdk/issues/364#issuecomment-541193861 (where this idea was proposed :+1:)

---

# Supported tags and respective `Dockerfile` links

**WARNING:** THIS IMAGE *IS NOT SUPPORTED* ON THE `s390x` ARCHITECTURE

[s390x/openjdk build status badge](https://doi-janky.infosiftr.net/job/multiarch/job/s390x/job/openjdk/)

# Quick reference

...

### DIFF
--- a/.template-helpers/generate-dockerfile-links-partial.sh
+++ b/.template-helpers/generate-dockerfile-links-partial.sh
@@ -13,6 +13,11 @@ if [ -z "${BASHBREW_LIBRARY:-}" ]; then
 	repo="https://github.com/docker-library/official-images/raw/master/library/$repo"
 fi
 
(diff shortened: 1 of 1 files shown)


james-crowley:
@tianon Thanks for your help! Also, do you happen to be on any IRC channels? Would love to talk more informally about a couple things.

yosifkit:
:+1:

> seek us out on Freenode IRC in channel `#docker-library`

https://github.com/docker-library/official-images/tree/5dd0367d4cd41e4fc5ce4cd4c0b9c882935190ac#contributing-to-the-standard-library

PR #1586, opened 2019-10-14
postgresql: link the correct /dev/shm compose docs

luto:
The current docs link to the _build time_ `/dev/shm` setting ([here](https://docs.docker.com/compose/compose-file/#shm_size)), but should link to the [run time setting](https://docs.docker.com/compose/compose-file/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir).


yosifkit:
Thanks!

PR #1588, opened 2019-10-15
Add docs for installing additional commands in telegraf image

danielnelson:
Adds a basic example of how to extend the official image with new packages and use them with Telegraf.

### DIFF
--- a/telegraf/content.md
+++ b/telegraf/content.md
@@ -207,3 +207,42 @@ $ docker run -d --name=telegraf \
 ```
 
 Refer to the docker [plugin documentation](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/docker/README.md) for more information.
+
+### Install Additional Packages
+
+Some plugins require additional packages to be installed.  For example, the `ntpq` plugin requires `ntpq` command.  It is recommended to create a custom derivative image to install any needed commands.
+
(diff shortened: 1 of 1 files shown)


yosifkit:
CC maintainer @jsternberg for content.

@danielnelson, it looks like there are a couple formatting issues caught by travis. It outputs a diff of what needs to change. It looks like it just wants an extra newline between a few parts.

I'd recommend removing `DEBIAN_FRONTEND` since it is unnecessary.

danielnelson:
Thanks for the suggestions, I've got the travis checks passing now and removed the `DEBIAN_FRONTEND` bit.

jsternberg:
@danielnelson is the official maintainer for Telegraf. Is there a way we can list him as a maintainer or change him to the primary maintainer of telegraf so he doesn't have to get docs changed approved through me?

yosifkit:
Adding him to the maintainers list (or replacing yourself) in the library file will ensure that we don't forget next time: https://github.com/docker-library/official-images/blob/48777e39e39e013bc97a47624b7046059a5b52f1/library/telegraf#L1


[Cassandra](https://github.com/docker-library/official-images/blob/48777e39e39e013bc97a47624b7046059a5b52f1/library/cassandra#L3-L4) is a good example if you want a list of maintainers

PR #1603, opened 2019-11-19
Unify shell command syntax

vuchl:


tianon:
Most other image descriptions go the other direction and include `$` explicitly everywhere to make it clear it's a command to run at a shell prompt. :confused:

vuchl:
@tianon I refactored to `$` and also edited the content.md file

PR #1604, opened 2019-11-22
Ensure cache is kept between builds

0b10011:
The default for `CARGO_HOME` is `$HOME_DIR/.cargo`, but that isn't mounted, so it is discarded after every build. This moves `CARGO_HOME` to `/usr/src/myapp/.cargo` within the container, or `$PWD/.cargo` on the host.

### DIFF
--- a/rust/content.md
+++ b/rust/content.md
@@ -35,7 +35,7 @@ $ docker run -it --rm --name my-running-app my-rust-app
 There may be occasions where it is not appropriate to run your app inside a container. To compile, but not run your app inside the Docker instance, you can write something like:
 
 ```console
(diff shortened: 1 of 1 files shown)


yosifkit:
cc Rust image maintainer @sfackler for content approval.

sfackler:
It seems like it'd be better to mount the user's real cargo home in place of the container's cargo home rather than making a new one in a nonstandard location in the host.

tianon:
Closing for now -- if someone wants to make a new PR which fits what's described as the desired change (or make a stronger case for this particular change), feel free to do so. :smile:

PR #1617, opened 2019-12-08
Slack URL updated

shahariaazam:
Previous URL was 404


tianon:
Awesome catch, thank you! :+1:

PR #1621, opened 2019-12-16
Update Bonita and Migration Tool versions

educhastenier:


educhastenier:
Relates to https://github.com/docker-library/official-images/pull/7137

PR #1633, opened 2020-01-08
Rewrite some bits of Debian image's "How It's Made" to clarify

tianon:
Per [some feedback from Holger](https://lists.reproducible-builds.org/pipermail/rb-general/2020-January/001779.html), I took the opportunity to rewrite/rephrase some of this to try and make it more clear how to reproduce the Debian tarballs with the `debuerreotype` tool, specifically around where to look to get the information required (timestamp, etc) to do so.

cc @h01ger (and fyi @paultag) would you be willing to take a look at this and see if I've done a better job? :innocent:
(I'd love to improve it further if you've got more feedback! :heart:)

Also, feel free to circulate this PR (or ask me to do so) on that r-b.o list if you think there'd be value in doing so, since I'd love to get more feedback on this! :+1:


tianon:
I've just pushed a new commit (685c4c1) linking to the new https://debuerreotype.github.io/ site I just set up (which is already pre-seeded with the result of merging https://github.com/docker-library/official-images/pull/7386, so the images aren't actually up on Docker Hub just yet :innocent:).

h01ger:
On Fri, Jan 31, 2020 at 10:37:37AM -0800, Tianon Gravi wrote:
> I've just pushed a new commit (685c4c1) linking to the new https://debuerreotype.github.io/ site I just set up (which is already pre-seeded with the result of merging https://github.com/docker-library/official-images/pull/7386, so the images aren't actually up on Docker Hub just yet :innocent:).

nice!


-- 
cheers,
	Holger

-------------------------------------------------------------------------------
               holger@(debian|reproducible-builds|layer-acht).org
       PGP fingerprint: B8BF 5413 7B09 D35C F026 FE9D 091A B856 069A AA1C

Issue #1634, opened 2020-01-10
Rust compiled binary example is incorrect


elpiel:
The issues with the example below is that the binary installation directory is actually `$CARGO_HOME/bin`, in the example it copies the Source code directory (`COPY --from=builder /usr/src/myapp /usr/local/bin/myapp`) which is not executable and thus this won't work.

The example:
```
FROM rust:1.39 as builder
WORKDIR /usr/src/myapp
COPY . .
RUN cargo install --path .

FROM debian:buster-slim
RUN apt-get update && apt-get install -y libssl
COPY --from=builder /usr/src/myapp /usr/local/bin/myapp
CMD ["myapp"]
```

tianon:
Good catch!

cc @mkinney (last to touch this section / author of this bit), @sfackler (`rust` image maintainer)

PR #1638, opened 2020-01-15
update doc to show working example

mkinney:
Resolves: #1634

### DIFF
--- a/rust/content.md
+++ b/rust/content.md
@@ -40,7 +40,7 @@ RUN cargo install --path .
 
 FROM debian:buster-slim
 RUN apt-get update && apt-get install -y extra-runtime-dependencies
-COPY --from=builder $CARGO_HOME/bin/myapp /usr/local/bin/myapp
+COPY --from=builder /usr/local/cargo/bin/myapp /usr/local/bin/myapp
 CMD ["myapp"]
 ```
 
--- a/rust/content.md
+++ b/rust/content.md
@@ -39,8 +39,8 @@ COPY . .
 RUN cargo install --path .
 
 FROM debian:buster-slim
-RUN apt-get update && apt-get install -y libssl
-COPY --from=builder /usr/src/myapp /usr/local/bin/myapp
+RUN apt-get update && apt-get install -y libssl-dev
+COPY --from=builder /usr/src/myapp/target/release/myapp /usr/local/bin/myapp
 CMD ["myapp"]
 ```
 
--- a/rust/content.md
+++ b/rust/content.md
@@ -33,14 +33,14 @@ $ docker run -it --rm --name my-running-app my-rust-app
 This creates an image that has all of the rust tooling for the image, which is 1.8gb. If you just want the compiled application:
 
 ```dockerfile
-FROM rust:1.39 as builder
+FROM rust:1.40 as builder
 WORKDIR /usr/src/myapp
 COPY . .
 RUN cargo install --path .
 
 FROM debian:buster-slim
 RUN apt-get update && apt-get install -y extra-runtime-dependencies
-COPY --from=builder /usr/src/myapp/target/release/myapp /usr/local/bin/myapp
+COPY --from=builder $CARGO_HOME/bin/myapp /usr/local/bin/myapp
 CMD ["myapp"]
 ```
 
--- a/rust/content.md
+++ b/rust/content.md
@@ -39,12 +39,12 @@ COPY . .
 RUN cargo install --path .
 
 FROM debian:buster-slim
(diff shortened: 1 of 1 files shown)


tianon:
LGTM, cc @sfackler

rust/content.md
```diff
@@ -39,8 +39,8 @@ COPY . .
 RUN cargo install --path .
 
 FROM debian:buster-slim
-RUN apt-get update && apt-get install -y libssl
-COPY --from=builder /usr/src/myapp /usr/local/bin/myapp
+RUN apt-get update && apt-get install -y libssl-dev
```
sfackler: I don't think the development package is required in this image.

mkinney: What is the non-dev version of libssl-dev? There is no libssl.

mkinney: I think the line is there to show how one could install required packages. A simple myapp probably would not require openssl.

tianon: So maybe both should be replaced by a dummy package name that doesn't exist to further illustrate that it's merely an example?

sfackler: Sorry - `libssl1.1`. Yeah, the line is intended to show how you'd pull in extra runtime dependencies, OpenSSL just being one relatively common example.

sfackler: Sure, we could swap it out to `extra-runtime-dependencies` or something like that.

mkinney: Ok. Makes sense.

rust/content.md
```diff
@@ -39,12 +39,12 @@ COPY . .
 RUN cargo install --path .
 
 FROM debian:buster-slim
-RUN apt-get update && apt-get install -y libssl
-COPY --from=builder /usr/src/myapp /usr/local/bin/myapp
+RUN apt-get update && apt-get install -y extra-runtime-dependencies
+COPY --from=builder /usr/src/myapp/target/release/myapp /usr/local/bin/myapp
```
elpiel: `cargo install` install the binary on the host machine and it doesn't create a release target.
the `myapp` binary will be stored in `$CARGO_HOME/bin/myapp`
`$CARGO_HOME` is an env. variable from the `Dockerfile` of Rust (see here: https://hub.docker.com/layers/rust/library/rust/latest/images/sha256-8b43199e12b23fba6643472c526ce757ad9fe34e89e3485af6cb588b47745959)

E.g. `/home/username/.cargo/bin/validator_worker`

Not sure if it's possible to share an env. variable for use from the builder :thinking:

mkinney: I don't think it is possible to get a variable from one docker multi-stage to the other.

I created a simple hello world to validate values. See https://github.com/mkinney/rust_hello_world

Dockerfile:
```
FROM rust:1.40 as builder
WORKDIR /usr/src/myapp
COPY . .
RUN cargo install --path .

FROM debian:buster-slim
COPY --from=builder /usr/local/cargo/bin/hello_world /usr/local/bin/myapp
CMD ["myapp"]
```

Here is the build output:
```
$ docker build -t myrust/hello .
Sending build context to Docker daemon  1.492MB
Step 1/7 : FROM rust:1.40 as builder
 ---> a18c58410bcf
Step 2/7 : WORKDIR /usr/src/myapp
 ---> Using cache
 ---> ad1e60e2d4a6
Step 3/7 : COPY . .
 ---> dd1317607ed3
Step 4/7 : RUN cargo install --path .
 ---> Running in 08cfb8068905
  Installing hello_world v0.1.0 (/usr/src/myapp)
   Compiling hello_world v0.1.0 (/usr/src/myapp)
    Finished release [optimized] target(s) in 0.19s
  Installing /usr/local/cargo/bin/hello_world
   Installed package `hello_world v0.1.0 (/usr/src/myapp)` (executable `hello_world`)
Removing intermediate container 08cfb8068905
 ---> 2f6313763dbf
Step 5/7 : FROM debian:buster-slim
 ---> 2dbffcb4f093
Step 6/7 : COPY --from=builder /usr/local/cargo/bin/hello_world /usr/local/bin/myapp
 ---> Using cache
 ---> 0c1c6ecaaa90
Step 7/7 : CMD ["myapp"]
 ---> Using cache
 ---> 5a1f47e07eac
Successfully built 5a1f47e07eac
Successfully tagged myrust/hello:latest
```

Here is testing it:
```
$ docker run -it --rm myrust/hello
Hello, world!
```

elpiel: Yeah, that's something I tried as well and it didn't work.
For the example: The only thing I am worried about is `$CARGO_HOME` changing in the Rust docker image, which might break the path if you hardcode it in the other stage.
But that's also good IMO now

PR #1639, opened 2020-01-15
[PHP] Add link to install-php-extensions

mlocati:
Is this change acceptable?
I received many requests to add a link to my [`install-php-extensions`](https://github.com/mlocati/docker-php-extension-installer) script, many people find it very useful...

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -61,7 +61,9 @@ RUN apt-get update && apt-get install -y \
 	&& docker-php-ext-install -j$(nproc) gd
 ```
 
-Remember, you must install dependencies for your extensions manually. If an extension needs custom `configure` arguments, you can use the `docker-php-ext-configure` script like this example. There is no need to run `docker-php-source` manually in this case, since that is handled by the `configure` and `install` scripts.
+Remember, you must install dependencies for your extensions manually. If you are having difficulty figuring out which Debian or Alpine packages need to be installed before `docker-php-ext-install`, then have a look at [`install-php-extensions`](https://github.com/mlocati/docker-php-extension-installer). This script builds upon the `docker-php-ext-*` scripts and simplifies the installation of PHP extensions by automatically adding and removing Debian (apt) and Alpine (apk) packages. For example, to install the memcached extension you simply have to run `install-php-extensions gd`. This tool is contributed by community members and is not included in the images, please refer to their git repo for installation, usage, and issues.
+
+If an extension needs custom `configure` arguments, you can use the `docker-php-ext-configure` script like this example. There is no need to run `docker-php-source` manually in this case, since that is handled by the `configure` and `install` scripts.
 
 See ["Dockerizing Compiled Software"](https://tianon.xyz/post/2017/12/26/dockerize-compiled-software.html) for a description of the technique Tianon uses for determining the necessary build-time dependencies for any bit of software (which applies directly to compiling PHP extensions).
 
--- a/php/content.md
+++ b/php/content.md
@@ -49,7 +49,7 @@ RUN docker-php-source extract \
 
 ### PHP Core Extensions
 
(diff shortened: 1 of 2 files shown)


yosifkit:
We should probably not put it in its own section. This tool is currently a special case to help users install dependencies and I don't want to encourage more tools to add links to their own at this time.

Maybe just swap to a paragraph about finding what dependencies to install? Something similar to this? Probably up after line 148?

If you are having difficulty figuring out which Debian or Alpine packages need to be installed before `docker-php-ext-install`, then have a look at [`install-php-extensions`](https://github.com/mlocati/docker-php-extension-installer).  This script builds upon the `docker-php-ext-*` scripts and simplifies the installation of PHP extensions by automatically adding and removing Debian (`apt`) and Alpine (`apk`) packages. For example, to install the `memcached` extension you simply have to run `install-php-extensions memcached`.  This tool is contributed by community members and is not included in the images, please refer to their git repo for installation, usage, and issues.

yosifkit:
(Also, thanks for maintaining this tool! :heart: It is very helpful.)

mlocati:
> Maybe just swap to a paragraph about finding what dependencies to install? Something similar to this?

Yep, much better! What about https://github.com/docker-library/docs/pull/1639/commits/ea98192a45b85783296cf36df311b0b5b9e76e64 ([preview](https://github.com/mlocati/docs/blob/ea98192a45b85783296cf36df311b0b5b9e76e64/php/README.md#php-core-extensions)) ?

> (Also, thanks for maintaining this tool! ❤️ It is very helpful.)

Thanks! I was quite tired to google for required packages every time I needed a new extension. Furthermore, if more and more people use this script, it can attract more contributors, so that it gets improved covering new extensions.

mlocati:
(typos fixed in ba43d0e472c6899ca053abb7e0199bd2ad28bd94)

mlocati:
Is there anything else that should be fixed?

php/content.md
```diff
@@ -61,7 +61,9 @@ RUN apt-get update && apt-get install -y \
...
+
+If an extension needs custom `configure` arguments, you can use the `docker-php-ext-configure` script like this example. There is no need to run `docker-php-source` manually in this case, since that is handled by the `configure` and `install` scripts.
 
 See ["Dockerizing Compiled Software"](https://tianon.xyz/post/2017/12/26/dockerize-compiled-software.html) for a description of the technique Tianon uses for determining the necessary build-time dependencies for any bit of software (which applies directly to compiling PHP extensions).
```
tianon: I think I'd rather keep all the "third party" mentions down here separate from the "included bits" description.

As in:

```markdown
Remember, you must install dependencies for your extensions manually. If an extension needs custom `configure` arguments, you can use the `docker-php-ext-configure` script like this example. There is no need to run `docker-php-source` manually in this case, since that is handled by the `configure` and `install` scripts.

If you are having difficulty figuring out which Debian or Alpine packages need to be installed before `docker-php-ext-install`, then have a look at [the `install-php-extensions` project](https://github.com/mlocati/docker-php-extension-installer). This script builds upon the `docker-php-ext-*` scripts and simplifies the installation of PHP extensions by automatically adding and removing Debian (apt) and Alpine (apk) packages. For example, to install the GD extension you simply have to run `install-php-extensions gd`. This tool is contributed by community members and is not included in the images, please refer to their Git repository for installation, usage, and issues.

See also ["Dockerizing Compiled Software"](https://tianon.xyz/post/2017/12/26/dockerize-compiled-software.html) for a description of the technique Tianon uses for determining the necessary build-time dependencies for any bit of software (which applies directly to compiling PHP extensions).
```

mlocati:
> I'm also happy to make any of these adjustments if you'd rather. 👍

Sure!

PR #1642, opened 2020-01-16
Add phpMyAdmin

J0WI:
https://github.com/docker-library/official-images/pull/7288
cc @williamdes


phpmyadmin/content.md
```diff
@@ -0,0 +1,133 @@
+# What is Postfix Admin?
```
williamdes: "Postfix Admin?" ?

phpmyadmin/content.md
```diff
@@ -0,0 +1,133 @@
...
+The official MySQL and MariaDB use following environment variables to define these:
+
+-	`MYSQL_ROOT_PASSWORD` - This variable is mandatory and specifies the password that will be set for the `root` superuser account.
+-	`MYSQL_USER`, `MYSQL_PASSWORD` - These variables are optional, used in conjunction to create a new user and to set that user's password.
+
+## Supported Docker hub tags
+
+The following tags are available:
+
+-	`latest`, `fpm`, and `fpm-alpine` are always the most recent released version
```
tianon, line 26: https://github.com/docker-library/official-images/pull/7288 doesn't have any of these tags -- is that an oversight here, or there? :smile:

phpmyadmin/content.md
```diff
@@ -0,0 +1,133 @@
...
+-	Major versions, such as `5`, `5-fpm`, and `5-fpm-alpine`
+-	Specific minor versions, such as `5.0`, `5.0-fpm`, and `5-fpm-alpine`
+-	Specific patch versions, such as `5.0.0`, `5.0.0-fpm`, and `5.0.0-alpine`. Note that, on rare occasion, here may be an intermediary "docker-only" release, such as 4.9.2-1
+-	`edge`, `edge-fpm`, `edge-fpm-alpine`, a development version build from the daily snapshot
```
tianon, line 30: Ditto here.

phpmyadmin/content.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Image variants
```
tianon, line 32: This section will be redundant with a similar section the overall template will also add (https://github.com/docker-library/docs/blob/55c3c55f78abe73947c4e268a24c01ac60228724/.template-helpers/variant.md and associated files); if this content is important/useful, it should probably be moved into `variant-foo.md` files to leverage/override that existing behavior.

phpmyadmin/content.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Variables that can be read from a file using `_FILE`
+
+-	PMA_PASSWORD
+-	MYSQL_ROOT_PASSWORD
+-	MYSQL_PASSWORD
+-	PMA_HOSTS
+-	PMA_HOST
+
+For more detailed documentation see https://docs.phpmyadmin.net/en/latest/setup.html#installing-using-docker
```
tianon, line 133: Given that this has much more detailed documentation about how to use the image, would it make sense to put less information into this image description so that the same information isn't being maintained in two different places?

The `registry` image/directory is a good example of taking this to the extreme (https://hub.docker.com/_/registry) but I think there's probably a happy medium that could be valuable here.

phpmyadmin/stack.yml
```diff
@@ -0,0 +1,15 @@
...
+
+services:
+  db:
+    image: mariadb:10.3
+    restart: always
+    environment:
+      MYSQL_ROOT_PASSWORD: notSecureChangeMe
+
+  phpmyadmin:
+    image: phpmyadmin
```
tianon, line 11: It seems that perhaps `:latest` being missing from https://github.com/docker-library/official-images/pull/7288 was the oversight? :sweat_smile:

(Did you want to include `restart: always` on this container as well so that it matches the `db` container?)

williamdes:
I added https://github.com/phpmyadmin/docker/commit/27f6f7e5aecc56b3a7231be3fd416ba5278f2837 to fix the missing tags @tianon

phpmyadmin/content.md
williamdes: I fixed it ;)

phpmyadmin/content.md
J0WI: Removed for now, I don't think that the library is suitable for daily snapshots.

phpmyadmin/content.md
J0WI: I'd keep it for now, because it gives a nice overview of available environment variables.
@williamdes do you may consider to change this in future?

williamdes: I think official phpMyAdmin documentation should always be linked, right ?

phpmyadmin/content.md
J0WI: I totally agree on this, but I wonder why the "apache vs. fpm" variant is gone from PHP images?

phpmyadmin/content.md
J0WI: True, but we don't need to write things here again that are already covered in the phpMyAdmin docs.

williamdes: Sure

But IMO the ENVs should be documented in the place near the implementation because our docs could diverge
So maybe remove the link ?

ibennetch: Why not link here from the official documentation? The official documentation could say something like "There is a Docker image, for simple cases you can use `docker run phpmyadmin/phpmyadmin`, for anything more advanced see the full list of supported options at Docker Hub?

I feel strongly we should only document things like this in only one place, and it makes more sense to me to put it on Docker Hub than docs.phpmyadmin.net.

J0WI: Sure, it doesn't matter where the documentation lives but duplication should be avoided. Anyway, I think this is not a blocker to get https://github.com/docker-library/official-images/pull/7288 done.

tianon: Let's definitely revisit this post-merge :+1:

PR #1646, opened 2020-01-23
Optimize PNG logos

J0WI:
I used OptiPNG, but feel free to use [anything else](https://en.wikipedia.org/wiki/Portable_Network_Graphics#Optimizing_tools).


tianon:
I understand the theory here, but in practice all these files are already pretty small and due to the way Git works we'll be downloading the larger size anyways when we clone this repository, so I'm not sure what the benefit of doing this is?

(Especially since most of these are provided directly by the relevant image maintainer or upstream, and thus we'll need to add some kind of CI to enforce this if we want to keep them optimized.)

J0WI:
This aims to reduce the bandwidth usage of network requests to Docker Hub.

tianon:
For what gets uploaded to Docker Hub, these get converted to be exactly 120x120 first:

https://github.com/docker-library/docs/blob/b34c47bea4b4ae5582c07f13e901da52304caf35/push.pl#L167-L172

So the original artifact size still isn't terribly relevant (and the concern about automatically enforcing "optimized images" still stands).

Additionally, the total size of *all* current `logo.png` files is 3.2MB (smaller than `alpine:latest` :sweat_smile:).

tianon:
This is interesting, and I'd normally be on board (assuming we could resolve some kind of CI-based reproducible check for it, which I'm sure we could), but *all* our current PNG files in this repository are a total of 3.4MB all together, so it's not really worth the churn.

Thanks! :heart:

PR #1658, opened 2020-02-21
Add warning about versions and breaking changes

mltsy:
This PR can be considered an alternative to https://github.com/docker-library/postgres/pull/689/files.

I don't know whether it makes more sense to add the warning here or there.  In either case, one should probably link to the other, for the full description.

Note: I added the warning text at the very top of the file mostly because the contents of this file appear directly below the "supported tags" section of the README after templates are applied... if there's a better way to arrange this, I'm open to suggestions, obviously ;)

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -1,3 +1,7 @@
+**Tags and versioning warning:**
+
+Pinning a tag (even to the patch level) does not prevent breaking changes.  See [Versioning Warning] below.
+
 # What is PostgreSQL?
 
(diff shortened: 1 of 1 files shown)


tianon:
I agree it would be good to document this better; thanks for taking a first pass!

That being said, I don't think a specific image is the right place for it -- I think our FAQ is probably a better place, since it's really a generic thing across all the official images: https://github.com/docker-library/faq

mltsy:
@tianon That would be great!  Although that buries it a little bit...

I'm torn on this because, as much as I would love to believe that everyone knows to RTFM before using new technology - in practice, I know a lot of people (including me) come to Docker not as a "student of new technology", but because they are working in places it's already implemented, and make what seem like "reasonable" assumptions until something breaks.  This particular issue does (at least to some degree) violate "reasonable assumptions" in a way that can surprise you by breaking a production environment when you least expect it, so it seems important to have more visible as a warning.  The docs for a particular image seem much more visible and likely to be read and/or referenced when something goes wrong with an image than the docs for the docker official images library as a whole.  But that's just my opinion... still open to suggestions.

Maybe a good compromise would be to add this documentation to the main FAQ or overview, and then allow image maintainers to put a note/link to that section (for visibility)?

PR #1668, opened 2020-03-05
✍️ Improve Postgres docs on password/auth environment vars

babakks:
Current documentations on PostgreSQL image, are a bit unclear about the **required**, **non-empty** `POSTGRES_PASSWORD` environment variable.

Now, more explanations are added and clearer emphasis on important notes are inserted.

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -95,7 +95,18 @@ See the PostgreSQL documentation on [`pg_hba.conf`](https://www.postgresql.org/d
 
 ### `PGDATA`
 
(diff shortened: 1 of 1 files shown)


riconnon:
This tripped me up too when upgrading from `9.6.16-alpine` to `9.6.17-alpine` I did not expect to find a significant change in behaviour.
This docs change would definitely help.

yosifkit:
Rebased and moved a bit around. Only document `POSTGRES_PASSWORD` being not required in the `POSTGRES_HOST_AUTH_METHOD` section.  Also added some `PGDATA` improvements.

PR #1672, opened 2020-03-10
Add EspoCRM docs

tmachyshyn:
This is the documentation for EspoCRM image, https://github.com/docker-library/official-images/pull/7603


espocrm/stack.yml
```diff
@@ -0,0 +1,60 @@
+version: '3.8'
+
+services:
+
+  mysql:
+    image: mysql:8
+    container_name: mysql
+    command: --default-authentication-plugin=mysql_native_password
```
LaurentGoderre, line 8: This line should be removed because it would break `mysql:8` (Would work with 8.0.x but not 8.4.x)

espocrm/stack.yml
```diff
@@ -0,0 +1,60 @@
...
+      ESPOCRM_ADMIN_PASSWORD: password
+      ESPOCRM_SITE_URL: "http://localhost:8080"
+    volumes:
+      - espocrm:/var/www/html
+    restart: always
+    ports:
+      - 8080:80
+
+  espocrm-daemon:
+    image: espocrm/espocrm
```
LaurentGoderre, line 36: same here

espocrm/stack.yml
```diff
@@ -0,0 +1,60 @@
...
+    container_name: espocrm-daemon
+    volumes:
+      - espocrm:/var/www/html
+    restart: always
+    entrypoint: docker-daemon.sh
+
+  espocrm-websocket:
+    image: espocrm/espocrm
```
LaurentGoderre, line 44: same here

tmachyshyn:
@LaurentGoderre thanks for your recommendations. That's all fixed.

LaurentGoderre (approved):

tmachyshyn:
@LaurentGoderre I fixed some ymlfmt & markdownfmt validation errors in the last commit.
If any other changes are needed please let me know.

tianon:
(force push: just a rebase to kick CI and see where we're at status-wise)

tianon:
There might be other things you want to update here (it's probably been a minute since you looked at it :smile:) but here are a few notes I saw while reviewing it:

---

The `docker run` examples use `--link`, which has been deprecated. The current pattern across these docs is `--network some-network` (see e.g. `drupal`, `joomla`):

```console
$ docker run --name mysql --network some-network -e MYSQL_ROOT_PASSWORD=password -d mysql:8
$ docker run --name some-%%REPO%% --network some-network -d %%IMAGE%%
```

```console
$ docker run --name some-%%REPO%% --network some-network -p 8080:80 -d %%IMAGE%%
```

```console
$ docker run --name some-%%REPO%% --network some-network -e ESPOCRM_SITE_URL=http://172.20.0.100:8080 -p 8080:80 -d %%IMAGE%%
```

(The container name `mysql` matches the default `ESPOCRM_DATABASE_HOST`, so no extra env var is needed.)

---

In `stack.yml`:

```yaml
ESPOCRM_ADMIN_PASSWORD: password
```

And in the prose examples:

> access it via `http://localhost:8080` with credentials "admin" and "password"

The image does the right thing -- it emits an explicit startup warning when default credentials are detected. But the docs present them as the happy path with no caveat. The stack.yml value should use a placeholder like `your_admin_password`, and the prose examples should note that those credentials need to be changed before any real use.

---

`ESPOCRM_DATABASE_PORT` is a supported install-time option (it's in the image's defaults) but doesn't appear in the environment variables section.

---

Two small prose fixes:

- "This image requires to run MySQL server" → "This image requires a running MySQL server"
- "These environment variables are using to define" → "are used to define"

espocrm/compose.yaml
```diff
@@ -0,0 +1,93 @@
...
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD", "bin/command", "app-check"]
+      start_period: 20s
+      interval: 180s
+      timeout: 20s
+      retries: 3
+
+  espocrm-websocket:
+    image: espocrm/espocrm:latest
```
tianon: This file should reference only images from DOI :eyes:

What's the purpose of each espocrm container having the same exact health check?  Does it actually check the running service in some way?

```suggestion
    image: espocrm:latest
    container_name: espocrm
    environment:
      ESPOCRM_DATABASE_HOST: espocrm-db
      ESPOCRM_DATABASE_USER: espocrm
      ESPOCRM_DATABASE_PASSWORD: your_database_password
      ESPOCRM_ADMIN_USERNAME: admin
      ESPOCRM_ADMIN_PASSWORD: your_admin_password
      ESPOCRM_SITE_URL: "http://localhost:8080"
    volumes:
      - espocrm-data:/var/www/html/data
      - espocrm-custom:/var/www/html/custom
      - espocrm-custom-client:/var/www/html/client/custom
    restart: unless-stopped
    depends_on:
      espocrm-db:
        condition: service_healthy
    healthcheck:
      test: ["CMD", "bin/command", "app-check"]
      start_period: 20s
      interval: 60s
      timeout: 20s
      retries: 3
    ports:
      - 8080:80

  espocrm-daemon:
    image: espocrm:latest
    container_name: espocrm-daemon
    volumes_from:
      - espocrm
    restart: unless-stopped
    entrypoint: docker-daemon.sh
    depends_on:
      espocrm:
        condition: service_healthy
    healthcheck:
      test: ["CMD", "bin/command", "app-check"]
      start_period: 20s
      interval: 180s
      timeout: 20s
      retries: 3

  espocrm-websocket:
    image: espocrm:latest
```

tmachyshyn:
@tianon thank you for your feedback. We have reviewed your comments and implemented several improvements. 
We did not reach out earlier, as this was still a draft version. Now, it's the final version of documentation.

Please feel free to share any further suggestions.

espocrm/content.md
```diff
@@ -0,0 +1,238 @@
...
+  -v espocrm-custom-client:/var/www/html/client/custom \
+  -p 8080:80 \
+  -d %%REPO%% && \
+docker run \
+  --name espocrm-daemon \
+  --network espocrm-network \
+  --restart unless-stopped \
+  --volumes-from espocrm \
+  --entrypoint docker-daemon.sh \
+  -d %%REPO%%
```
tianon: These should both still use `%%IMAGE%%` (which on the per-arch repo pages will expand to something like `amd64/espocrm` instead of just `espocrm` like `%%REPO%%` does -- they're the same on the library/ repos though).

espocrm/content.md
```diff
@@ -0,0 +1,238 @@
...
+docker run \
+  --name espocrm \
+  --network espocrm-network \
+  -p 8080:80 \
+  -e ESPOCRM_DATABASE_USER=espocrm \
+  -e ESPOCRM_DATABASE_PASSWORD=your_database_password \
+  -e ESPOCRM_ADMIN_USERNAME=admin \
+  -e ESPOCRM_ADMIN_PASSWORD=your_admin_password \
+  -e ESPOCRM_SITE_URL=http://192.168.0.100:8080 \
+  -d %%REPO%%
```
tianon: Same (`%%IMAGE%%`)

espocrm/content.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Docker Run
+
+To upgrade EspoCRM when using Docker Run:
+
+1\. Pull the latest images:
+
+```bash
+docker pull mariadb
+docker pull %%REPO%%
```
tianon: Ditto (`%%IMAGE%%`)

espocrm/logo.png
tianon, line 1: I don't know if you've got an SVG version of this, but it seems like it'd lend itself well to an SVG format (`logo.png` can be replaced by `logo.svg` and then our tooling will generate crisp versions of whatever sizes Docker Hub needs, both now and in the future).

tmachyshyn:
@tianon thank you for your review.
The `image` definition has been fixed, and the logo has been added in SVG format.

tianon (approved):

espocrm/content.md
```diff
@@ -0,0 +1,238 @@
...
+docker run \
+  --name espocrm-db \
+  --network espocrm-network \
+  --restart unless-stopped \
+  -e MARIADB_DATABASE=espocrm \
+  -e MARIADB_USER=espocrm \
+  -e MARIADB_PASSWORD=your_database_password \
+  -e MARIADB_ROOT_PASSWORD=your_root_password \
+  -v espocrm-db:/var/lib/mysql \
+  -d mariadb && \
```
yosifkit, line 32: Minor, non-blocking, style nit (perhaps even moved as the first argument to `docker run`):
```suggestion
  -d \
  mariadb && \
```
With all the other lines following the `--flag value` pattern, having the image name next to `-d` might incorrectly lead someone to believe that the image is a value passed to the `-d` flag. Applies to other `docker run` lines as well.

yosifkit (approved):

PR #1675, opened 2020-03-11
Adding Hitch

ThijsFeryn:
This is a companion PR to https://github.com/docker-library/official-images/pull/8674

### DIFF
--- a/hitch/content.md
+++ b/hitch/content.md
@@ -40,9 +40,9 @@ docker run hitch "--backend=[varnish]:8443"
 
 ## Setting the certificate
 
-The Hitch Docker image comes with a self-signed certificate that is stored in `/etc/hitch/certs/default`. 
+The Hitch Docker image comes with a self-signed certificate that is stored in `/etc/hitch/certs/default`.
 
-This certificate is automatically created during *Hitch* package install, and is a self-signed certificate using 2048-bit RSA-encrypted cipher. It is set up for the `localhost` hostname, with an expiration date 30 years in the future. 
+This certificate is automatically created during *Hitch* package install, and is a self-signed certificate using 2048-bit RSA-encrypted cipher. It is set up for the `localhost` hostname, with an expiration date 30 years in the future.
 
 This certificate is only suited for testing. Using a bind mount, you can override the value of the certificate and use your own certificate, which is advisable.
 
--- a/hitch/content.md
+++ b/hitch/content.md
@@ -7,7 +7,7 @@
 Running a Hitch Docker container can be done by using the following command:
 
 ```console
-docker run --name=hitch -p 443:443 hitch:latest
+$ docker run --name=hitch -p 443:443 %%IMAGE%%
 ```
 
 This container will expose port `443`, which is required for HTTPS traffic.
@@ -17,10 +17,10 @@ This container will expose port `443`, which is required for HTTPS traffic.
 Without any argument, the container will run `hitch --config=/etc/hitch/hitch.conf`. You can mount your own configuration file to replace the default one:
 
 ```console
-docker run -v /path/to/your/config/file:/etc/hitch/hitch.conf hitch
+$ docker run -v /path/to/your/config/file:/etc/hitch/hitch.conf:ro %%IMAGE%%
 ```
 
-You can also change the path of the configuration file by setting the `HITCH_CONFIG_FILE` environment variable. You can set it to an empty string to disable the configuration file altogether.
+You can also change the path of the configuration file by setting the `HITCH_CONFIG_FILE` environment variable.
 
 Note that extra arguments can be added to the command line. If the first argument starts with a `-`, the arguments are added to the default command line, otherwise they are treated as a command.
 
@@ -35,7 +35,7 @@ But you'll probably run your Varnish in a separate container. In that case, you'
 Here's how you set the backend via a *command-line option*:
 
 ```console
-docker run hitch "--backend=[varnish]:8443"
+$ docker run %%IMAGE%% "--backend=[varnish]:8443"
 ```
 
 ## Setting the certificate
@@ -49,7 +49,7 @@ This certificate is only suited for testing. Using a bind mount, you can overrid
 Here's an example:
 
 ```console
-docker run -v /path/to/your/certificate:/etc/hitch/certs/default hitch
+$ docker run -v /path/to/your/certificate:/etc/hitch/certs/default:ro %%IMAGE%%
 ```
 
 You can also override the [`pem-file`](https://github.com/varnish/hitch/blob/master/hitch.conf.man.rst#pem-file--string) configuration setting in your mounted configuration file.
@@ -59,5 +59,5 @@ If you prefer setting the certificate location on the command line, you can add 
 Here's how you do this:
 
 ```console
-docker run hitch "--backend=[varnish]:8443:/path/to/cert.pem"
+$ docker run %%IMAGE%% "--backend=[varnish]:8443:/path/to/cert.pem"
 ```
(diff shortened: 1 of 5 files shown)


hitch/content.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Configuration file and extra options
+
+Without any argument, the container will run `hitch --config=/etc/hitch/hitch.conf`. You can mount your own configuration file to replace the default one:
+
+```console
+docker run -v /path/to/your/config/file:/etc/hitch/hitch.conf hitch
+```
+
+You can also change the path of the configuration file by setting the `HITCH_CONFIG_FILE` environment variable. You can set it to an empty string to disable the configuration file altogether.
```
tianon, line 23: Ah @gquintard, this is the behavior my version of the script lost for you -- however, wouldn't it be easier/more intuitive to suggest users change the command to `hitch` to disable the default behavior?  ie:

```console
$ docker run ... %%IMAGE%% hitch
```

hitch/content.md
```diff
@@ -0,0 +1,63 @@
...
+
+Note that extra arguments can be added to the command line. If the first argument starts with a `-`, the arguments are added to the default command line, otherwise they are treated as a command.
+
+> Our assumption is that your backend, Varnish or other, supports both *HTTP/2* and the *PROXY* protocol.
+
+## Connecting to Varnish
```
tianon, line 29: I realize it's "Hitch by Varnish" but up until this point this reads as a totally generic TLS proxy, so as someone who is admittedly not using Varnish currently, this came as a surprise (and I don't see non-Varnish covered later on either) -- is that intentional?

The Hitch documentation implies even `PROXY` protocol support is optional, but my understanding of `HTTP/2` is that it only supports TLS, so I'm not sure how this could forward to an unencrypted backend properly if it requires `HTTP/2` support?  Is that just referring to the default settings?

hitch/content.md
```diff
@@ -0,0 +1,63 @@
...
+
+By default Hitch will connect to Varnish using `localhost:8843` using the [PROXY protocol](https://github.com/varnish/hitch/blob/master/docs/proxy-protocol.md). If your `varnishd` process has been started with `-a localhost:8443,PROXY`, the two will be able to talk together and Varnish will expose the true client IP as `client.ip` in VCL.
```
tianon, line 31: It's addressed in the following paragraph, but this feels a little misleading (since running in the same Kubernetes Pod is the only common use case that leads to `localhost` meaning much / working the way this is documented, and even that isn't horribly common these days in my experience since it's so limited :grimacing:).

If this is encoded in the image-provided `hitch.conf` (not the compiled-in defaults), maybe it would make sense to do something like `sed -i -e 's/localhost/varnish/g' /etc/hitch/hitch.conf` so it can work OOTB with multiple containers?

(I think it's also fine to leave as-is -- having users supply a command to set an explicit `--backend` is totally reasonable, but in that case I think this wording could maybe use some tweaking. :smile:)

hitch/content.md
```diff
@@ -0,0 +1,63 @@
...
+
+But you'll probably run your Varnish in a separate container. In that case, you'll want to change the backend settings. You can either do that by replacing the [`backend`](https://github.com/varnish/hitch/blob/master/hitch.conf.man.rst#backend--) configuration setting in your mounted configuration file, or by adding a *command-line option*.
+
+Here's how you set the backend via a *command-line option*:
+
+```console
+docker run hitch "--backend=[varnish]:8443"
```
tianon, line 38:
```suggestion
$ docker run %%IMAGE%% "--backend=[varnish]:8443"
```

hitch/content.md
```diff
@@ -0,0 +1,63 @@
...
+The Hitch Docker image comes with a self-signed certificate that is stored in `/etc/hitch/certs/default`.
+
+This certificate is automatically created during *Hitch* package install, and is a self-signed certificate using 2048-bit RSA-encrypted cipher. It is set up for the `localhost` hostname, with an expiration date 30 years in the future.
+
+This certificate is only suited for testing. Using a bind mount, you can override the value of the certificate and use your own certificate, which is advisable.
+
+Here's an example:
+
+```console
+docker run -v /path/to/your/certificate:/etc/hitch/certs/default hitch
```
tianon, line 52:
```suggestion
$ docker run -v /path/to/your/certificate:/etc/hitch/certs/default:ro %%IMAGE%%
```

hitch/content.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+You can also override the [`pem-file`](https://github.com/varnish/hitch/blob/master/hitch.conf.man.rst#pem-file--string) configuration setting in your mounted configuration file.
+
+If you prefer setting the certificate location on the command line, you can add the location as part of the `--backend` option.
+
+Here's how you do this:
+
+```console
+docker run hitch "--backend=[varnish]:8443:/path/to/cert.pem"
```
tianon, line 62:
```suggestion
$ docker run %%IMAGE%% "--backend=[varnish]:8443:/path/to/cert.pem"
```

hitch/maintainer.md
```diff
@@ -0,0 +1 @@
+[the Hitch Docker Community](%%GITHUB-REPO%%)
```
tianon, line 1: Did you want this to be something stronger, given it's officially maintained by/with the Hitch project, or are you happy with this wording? (just to be sure :+1:)

tianon (commented):
A few notes -- notably, using `%%IMAGE%%` instead of `hitch` will ensure that pages for images like `amd64/hitch` will appropriately shift to the specific architecture image name.

It might also be good to use something like `%%IMAGE%%:1` or `%%IMAGE%%:1.7` to encourage users to use something more specific than `:latest`, but that's entirely your call.

hitch/content.md
gquintard: agreed

```suggestion
You can also change the path of the configuration file by setting the `HITCH_CONFIG_FILE` environment variable.
```

hitch/content.md
gquintard: Looooots to unpack here :-) But you are right, it can be confusing.

There's a strong relationship between `hitch` and `varnish`, that you can notably see in the default configuration files:
- https://github.com/varnishcache/pkg-varnish-cache/blob/master/systemd/varnish.service#L26
- https://github.com/varnish/pkg-hitch/blob/master/systemd/hitch.conf#L18

but the two are still completely independent, you can use `HAProxy` to terminate TSL and use the `PROXY` protocol to contact Varnish, or you can use `hitch` in front on any other TCP server that lacks TLS.

Regarding H/2 mandating TLS, yes and no, there are provisions outside of TLS, but that not necessarily the norm.

I think it would make sense to remove the whole section and tackle the whole `hitch`+`varnish` synergy in a separate repo.

hitch/content.md
gquintard: the most common use-case is on the same machine, or in the same pod, so I'd leave `localhost`

hitch/maintainer.md
gquintard: I'd say it's fine, unless there's another angle than glory and recognition I haven't seen :-)

hitch/content.md
ThijsFeryn: @tianon @gquintard In https://github.com/docker-library/docs/pull/1675/commits/0d7fe8213d58cadbe5f06990dcfc20d5ed0396fa I removed references to Varnish and mentioned backend connections in general.

yosifkit (approved):

hitch/content.md
```diff
@@ -0,0 +1,63 @@
...
+
+You can also change the path of the configuration file by setting the `HITCH_CONFIG_FILE` environment variable.
+
+Note that extra arguments can be added to the command line. If the first argument starts with a `-`, the arguments are added to the default command line, otherwise they are treated as a command.
+
+> Our assumption is that your backend, Varnish or other, supports both *HTTP/2* and the *PROXY* protocol.
+
+## Connecting to the backend
+
+By default Hitch will connect to the backend using `localhost:8843` using the [PROXY protocol](https://github.com/varnish/hitch/blob/master/docs/proxy-protocol.md). If your backend server *PROXY*, the two will be able to talk together and backend will be able to expose the true client IP.
```
tianon, line 31: > If your backend server *PROXY*, the two ...

This feels like there's a word missing -- maybe "implements"? :sweat_smile:

PR #1683, opened 2020-03-20
Add wrlx-image for Wind River Linux images

robertlinux:
Signed-off-by: Robert Yang <liezhi.yang@windriver.com>


wrlx-image/content.md
```diff
@@ -0,0 +1,45 @@
...
+# How to use this image
+
+## Usage
+
+Use like you would any other base image:
+
+```dockerfile
+FROM %%IMAGE%%:10.18.0011
+RUN opkg update
+RUN opkg install python3-django
```
tianon, line 16: If `opkg update` is similar to `apt-get update` (and only updates the list of available packages), then these should be combined:

```suggestion
RUN opkg update && opkg install python3-django
```

---

If `opkg update` is more similar to `apt-get upgrade` (updating packages themselves, not just the list of available packages), then it should be dropped instead -- we encourage base image maintainers to update their base image periodically to get updated packages downstream to users:

```suggestion
RUN opkg install python3-django
```

wrlx-image/content.md
```diff
@@ -0,0 +1,45 @@
...
+%%LOGO%%
+
+# How to use this image
+
+## Usage
+
+Use like you would any other base image:
+
+```dockerfile
+FROM %%IMAGE%%:10.18.0011
```
tianon, line 14: You probably want to consider adding more "rolling" tags to these images, so users can do something like `FROM %%IMAGE%%:10.18` (and thus be able to get updated images by rebuilding instead of having to update their `Dockerfile` every time).

PR #1686, opened 2020-03-31
Wordpress: extend doc on PHP extensions

Adrien-Luxey:
The lack of emailing is a good example of an additional extension that is not included in the image. Also referencing [wordpress/issues/30](https://github.com/docker-library/wordpress/issues/30) because it contains good suggestions on the topic.


tianon:
Amended to remove the extraneous extra trailing space (Travis caught it in https://travis-ci.org/github/docker-library/docs/builds/669097752, but didn't notify this PR properly for some reason). :+1:

Thanks for the contribution!

Adrien-Luxey:
Thanks for the correction and for the merge!

PR #1689, opened 2020-04-01
clean dockerfile layer in example

jdrouet:
Following the [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) we should encourage people to remove the cache after installing dependencies.

### DIFF
--- a/rust/content.md
+++ b/rust/content.md
@@ -39,7 +39,7 @@ COPY . .
 RUN cargo install --path .
 
 FROM debian:buster-slim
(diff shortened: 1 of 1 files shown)


tianon:
cc `rust` image maintainer @sfackler

(also, for this to be in a format we can accept, assuming @sfackler is :+1: on the change, you'll need to move the edit from `README.md` over to `content.md` since the former is a generated artifact)

sfackler:
SGTM

tianon:
Rebased and moved the change to `content.md` -- thanks!

PR #1690, opened 2020-04-01
cache dependencies to optimize workflow

jdrouet:
Use `cargo fetch` to download the dependencies and keep them in a layer to improve the development process.

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -98,6 +98,24 @@ Note: Some shared libraries may need to be installed as shown in the installatio
 
 This method will create an image that is less than 200mb. If you switch to using the Alpine-based rust image, you might be able to save another 60mb.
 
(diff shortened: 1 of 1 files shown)


tianon:
cc `rust` image maintainer @sfackler

(also, for this to be in a format we can accept, assuming @sfackler is :+1: on the change, you'll need to move the edit from `README.md` over to `content.md` since the former is a generated artifact)

tianon:
IMO this is a pretty decent case of a good "exercise for the reader" but it's really your call, @sfackler :+1:

sfackler:
I agree that this is more complexity than is necessary for a hello world example.

tianon:
Ok, makes sense. :+1:

Thank you for the contribution! :heart:

PR #1698, opened 2020-04-16
Add proxy section

J0WI:


tianon:
This is the updates that go with https://github.com/docker-library/official-images/pull/7812, right?

@tilosp did you want to take a look over these too before we merge?

PR #1699, opened 2020-04-19
caddy: Add square logo for Hub UI

hairyhenderson:
Adding an additional `logo-120.png` so that the Hub UI can show Caddy's logo!

I figured this out by reading @tianon's `push.pl` script, and I think this is the first instance in the repo, but I assume this is OK - having both a `logo.png` and a `logo-120.png` means that the description page can show the full rectangular logo, while the square logo should get picked up for the Hub UI.

@tianon @yosifkit please correct me if this doesn't make sense! 😅 

Signed-off-by: Dave Henderson <dhenderson@gmail.com>


tianon:
Let's find out together whether my code works properly! :smile:

mholt:
Cool, thanks very much @tianon  and @hairyhenderson !

hairyhenderson:
@tianon BTW - the logo worked, and looks great! 🎉

PR #1702, opened 2020-04-27
Update WebSphere Liberty documentation

arthurdm:
Correcting the order where `configure.sh` must be called.  Also, we need a server.xml in the sample.


tianon:
Same story, sorry for the delay. :disappointed:

PR #1703, opened 2020-04-27
Update Open Liberty docs

arthurdm:
* Moving configure.sh to the correct place
* Add server.xml in the sample snippet


tianon:
Doh, not sure how we missed this; sorry!

PR #1705, opened 2020-04-29
Update README.md

management-dev:
To include the environment variable for url in the docker-compose example.

### DIFF
--- a/ghost/README.md
+++ b/ghost/README.md
@@ -147,6 +147,8 @@ services:
     ports:
       - 8080:2368
     environment:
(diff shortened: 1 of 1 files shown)


yosifkit:
>Error: at least one repo README.md has changed
>These files are autogenerated, so it is unnecessary to modify them
>Please update content.md and docker-library-bot will take care of README.md
>See: https://github.com/docker-library/docs/#image-namereadmemd

tianon:
`url:` was added in https://github.com/docker-library/docs/pull/1722 :+1:

Thanks!

PR #1711, opened 2020-05-01
Add initial GitHub Actions CI

tianon:
I'm not sure whether I have to push this initial branch to the origin for it to work properly, but hopefully not! :smile:


tianon:
Yep, looks like it has to be on the origin. :sweat_smile:

PR #1712, opened 2020-05-01
Add initial GitHub Actions CI

tianon:
#1711, take two


tianon:
Nice, b48dc50 -> https://github.com/docker-library/docs/actions/runs/93259313 :muscle:

PR #1713, opened 2020-05-03
Fix incorrect port mapping of 2368 and update to latest docker image version

hong823:
Ghost's port is displayed as `8080` which is incorrect, port `2368` should be the correct port as there are URLs on the page that is redirecting to port `2368`.

Example:
```
http://localhost:2368/author/ghost/
```

### DIFF
--- a/ghost/README.md
+++ b/ghost/README.md
@@ -144,7 +144,7 @@ version: '3.1'
 services:
 
   ghost:
-    image: ghost:1-alpine
+    image: ghost
     restart: always
     ports:
       - 2368:2368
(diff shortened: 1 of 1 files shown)


tianon:
It's not "incorrect" so much as "intentionally using a different host port than the port inside the container" which is why the line in the YAML has two values (the first being the "outside port" and the second being the "inside port" which is the port Ghost/Node is listening on).

What we should probably do instead is add an example `url` environment variable (which is a commonly requested addition) pointing to the exposed port.

PR #1721, opened 2020-05-20
[ArangoDB] Add section about Limiting resource utilization

Simran-B:
https://www.arangodb.com/docs/devel/programs-arangod-env-vars.html

### DIFF
--- a/arangodb/content.md
+++ b/arangodb/content.md
@@ -130,7 +130,7 @@ See more information about [Configuration](https://www.arangodb.com/docs/stable/
 
 `arangod` checks the following environment variables, which can be used to restrict how much memory and how many CPU cores it should use:
 
--	`ARANGODB_OVERRIDE_DETECTED_TOTAL_MEMORY` _(introduced in v3.6.3)_
+-	`ARANGODB_OVERRIDE_DETECTED_TOTAL_MEMORY` *(introduced in v3.6.3)*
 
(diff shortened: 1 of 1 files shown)


dothebart:
@tianon thanks, I've discussed with my collegue @Simran-B (who is mainly responsible for the arangodb documentation) to create this PR beforehand.

Hope you like it, as we seem to have users having problems about arangodb not adjusting memory usage. 
We found it a bit of a pity to add this workaround, as without docker ArangoDB would use the proc filesystem to retrieve this information. Is there any easy (or default by convention) way for a programm to determine its memory resources in a docker container to have a more common sense or at best automatic solution for this?

Cheers from Cologne, 
Willi

yosifkit:
> Is there any easy (or default by convention) way for a programm to determine its memory resources in a docker container to have a more common sense or at best automatic solution for this?

I think the canonical place is `/sys/fs/cgroup/memory/memory.limit_in_bytes` (https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt), but that may be different in future systems using `cgroup-v2` (https://www.kernel.org/doc/Documentation/cgroup-v2.txt).

tianon:
https://github.com/arangodb/arangodb/pull/11682 looks nice! :+1:

Do you still want to merge this PR in the interim? (just double checking before I hit merge :smile:)

Simran-B:
Yes, please! It will remain relevant for at least 3.6, which may not
receive a backport, and for the 3.7 alpha versions which are already out
there.

>
We will open another PR once the new implementation landed.

PR #1725, opened 2020-05-27
Update Golang documentation to use latest stable version (1.14)

jbleduigou:
This is very similar to my previous PR:
https://github.com/docker-library/docs/pull/1657

I am wondering if we could automate that somehow?


tianon:
Yeah, I guess it would be interesting to create some kind of either `%%LATEST%%` template variable that references the "most specific" tag of the `foo:latest` group or maybe some way to specify a `bashbrew cat --format` string inside the templates (which would be much more flexible).  I bet there are a bunch of images where such a thing would be useful, although it's a bit complicated to implement.

However, thanks for the contributions in the meantime! :+1:

PR #1727, opened 2020-05-29
Update content.md

nastena1606:
Replace MariaDB with Percona Server in the "Connect to Percona Server from the MySQL command line client" section


tianon:
Great catch! :+1:

PR #1738, opened 2020-06-11
Add "Classic Swarm" deprecation notice

tianon:
Rendering preview:

---

# **DEPRECATION NOTICE**

> Classic Swarm has been archived and is no longer actively developed. You may want to use the Swarm mode built into the Docker Engine instead, or another orchestration system.

(https://github.com/docker/classicswarm#readme)

# Quick reference

...


tianon:
Do you think it's also worth linking to https://github.com/docker/classicswarm/pull/2986, or let folks dig that out themselves?  Maybe including more of the text from the PR here?

justincormack:
I think this is probably enough.

PR #1750, opened 2020-06-23
Add a note about world-writeable "/go"

tianon:
Closes https://github.com/docker-library/golang/issues/324


yosifkit:
https://github.com/docker-library/golang/issues/324#issuecomment-642328520 is short enough that we could just put the gist of it here. WDYT?

tianon:
Sure, updated with a second pass

PR #1754, opened 2020-06-28
Correct instructions  for server-client networking

alecthegeek:
1. Added `--network` when starting mongo server
2. Fixed `--network` option when starting client
3. Added additional notes about use of Docker network

### DIFF
--- a/mongo/README.md
+++ b/mongo/README.md
@@ -106,7 +106,7 @@ First developed by the software company 10gen (now MongoDB Inc.) in October 2007
 ## Start a `mongo` server instance
 
 ```console
-$ docker run --name some-mongo --network=some-network-d mongo:tag
+$ docker run --name some-mongo --network some-network -d mongo:tag
 ```
 
 ... where `some-mongo` is the name you want to assign to your container, tag` is the tag specifying the MongoDB version you want, and `some-network` is an existing Docker network. See the list above for relevant tags.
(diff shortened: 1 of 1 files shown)


mongo/README.md
```diff
@@ -106,20 +106,22 @@ First developed by the software company 10gen (now MongoDB Inc.) in October 2007
 ## Start a `mongo` server instance
 
 ```console
-$ docker run --name some-mongo -d mongo:tag
+$ docker run --name some-mongo --network=some-network-d mongo:tag
```
tianon, line 109: I'm OK with this change (assuming we fix the syntax to be consistent/functional as in the suggestion below), but the rest of the changes here are in contrast to the explicit decision in https://github.com/docker-library/docs/pull/1441 to *not* document any details about Docker network creation (since for most users, that's going to be a complicated and individualized topic anyhow, especially if they're using something like Kubernetes instead of Docker), and we're just documenting the image here and don't want to devolve into a full mini Docker tutorial for every official image description. :sweat_smile:

```suggestion
$ docker run --name some-mongo --network some-network -d mongo:tag
```

tianon: (And also, `README.md` is an auto-generated file, so this change needs to be moved over to `content.md` instead, as per the CI failure on this PR.)

alecthegeek:
Closing as not of use

PR #1757, opened 2020-07-01
Add a dedicated "Julia Support Tiers" callout, explicitly noting Alpine/musl

tianon:
This is per https://github.com/docker-library/julia/pull/47#issuecomment-652661869 :heart:


tianon:
Thanks for the suggestion and the review, @fredrikekre! :heart:

PR #1761, opened 2020-07-09
Add notice about /usr/bin/python

wglambert:
Closes https://github.com/docker-library/python/issues/486


tianon:
Definite improvement, thanks :+1:

PR #1767, opened 2020-07-20
Update Telegraf instructions for host monitoring

### LINKED ISSUE #1756: [telegraf] update the docs about monitoring the the host filesystem from within an container
see https://github.com/influxdata/influxdata-docker/issues/399 for details

danielnelson:
Update the general instructions for monitoring the docker engine host system.

closes #1756

### DIFF
--- a/telegraf/content.md
+++ b/telegraf/content.md
@@ -182,11 +182,6 @@ The precise files that need to be made available varies from plugin to plugin. H
 $ docker run -d --name=telegraf \
 	-v $PWD/telegraf.conf:/etc/telegraf/telegraf.conf:ro \
 	-v /:/hostfs:ro \
-	-v /etc:/hostfs/etc:ro \
-	-v /proc:/hostfs/proc:ro \
-	-v /sys:/hostfs/sys:ro \
-	-v /var:/hostfs/var:ro \
-	-v /run:/hostfs/run:ro \
 	-e HOST_ETC=/hostfs/etc \
 	-e HOST_PROC=/hostfs/proc \
 	-e HOST_SYS=/hostfs/sys \
--- a/telegraf/content.md
+++ b/telegraf/content.md
@@ -172,19 +172,28 @@ Check that the measurement `foo` is added in the DB.
 
 -	[Output Plugins](https://docs.influxdata.com/telegraf/latest/plugins/outputs/)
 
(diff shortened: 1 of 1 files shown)


telegraf/content.md
```diff
@@ -172,19 +172,28 @@ Check that the measurement `foo` is added in the DB.
...
-      -v /proc:/host/proc:ro \
-      -v $PWD/telegraf.conf:/etc/telegraf/telegraf.conf:ro \
-      %%IMAGE%%
+	-v $PWD/telegraf.conf:/etc/telegraf/telegraf.conf:ro \
+	-v /:/hostfs:ro \
+	-v /etc:/hostfs/etc:ro \
+	-v /proc:/hostfs/proc:ro \
+	-v /sys:/hostfs/sys:ro \
+	-v /var:/hostfs/var:ro \
+	-v /run:/hostfs/run:ro \
```
tianon, line 189: This seems kind of strange -- shouldn't this just be `-v /:/hostfs:rslave,ro`, and then everything else will be inherited via that one single mount?

danielnelson:
Checked and there doesn't appear to be a reason to mount separately, so I removed these from the docker command. 

Corresponding change in the Telegraf docs: https://github.com/influxdata/telegraf/pull/7868

rdxmb:
@danielnelson Thanks.

When using

`-e HOST_MOUNT_PREFIX=/hostfs`

the docs do not tell about why you should also use

```
-e HOST_ETC=/hostfs/etc \
-e HOST_PROC=/hostfs/proc \
-e HOST_SYS=/hostfs/sys \
-e HOST_VAR=/hostfs/var \
-e HOST_RUN=/hostfs/run \
```

Could you add another sentence to clarify that, please?

tianon:
I looked at the code, and of those only `HOST_PROC` appears to be used, so my guess is that they're intended to be for future-proofing?

danielnelson:
These environment variables are used by the gopsutil library, which Telegraf uses for most of the default plugins, and alter the locations of the filesystems.  Each plugins has different requirements for which variables need set, but these are all the options that may need to be set.

I think the details of these options should probably be documented in the main Telegraf documentation, since the docker documentation is more of a quick overview and getting started guide.

Upstream docs: https://github.com/shirou/gopsutil/#usage

PR #1769, opened 2020-07-23
Update content.md

cheesemacfly:
The `'` characters around the parameter name and value prevent it to be properly passed and should be skipped.


tianon:
This depends on the shell -- in Bash (or any Posix Shell, really), these will work as-is.  I believe they'll also work as-is in PowerShell, although I'm not quite as familiar with that one.

That being said, this change should be reasonably harmless.

cheesemacfly:
You are right. And it does work in PowerShell as well but it doesn't in the basic windows command line tool. And like you said, it's a very small change and without the `'` character it still works everywhere else as far as i can tell.

yosifkit:
Yeah, it should be harmless.  I would have put them there because I assumed that there would be variables that might be valid with whitespace in their value.

PR #1770, opened 2020-07-23
Link to PHP docs for .ini configuration in WordPress

wglambert:
Closes https://github.com/docker-library/wordpress/issues/519

### DIFF
--- a/wordpress/content.md
+++ b/wordpress/content.md
@@ -85,4 +85,4 @@ When running WP-CLI via the `cli` variants of this image, it is important to not
 
 See [the "Configuration" section of the `php` image documentation](https://hub.docker.com/_/php/).
 
-Additionally you could create a `custom.ini` with the desired parameters and place it in the `$PHP_INI_DIR/conf.d/` directory.
+For example, to adjust common `php.ini` flags like `upload_max_filesize`, you could create a `custom.ini` with the desired parameters and place it in the `$PHP_INI_DIR/conf.d/` directory.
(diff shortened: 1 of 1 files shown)


wordpress/content.md
```diff
@@ -80,3 +80,9 @@ Mount the volume containing your themes or plugins to the proper directory; and
 See [the "Running as an arbitrary user" section of the `php` image documentation](https://hub.docker.com/_/php/).
 
 When running WP-CLI via the `cli` variants of this image, it is important to note that they're based on Alpine, and have a default `USER` of Alpine's `www-data`, whose UID is `82` (compared to the Debian-based WordPress variants whose default effective UID is `33`), so when running `%%IMAGE%%:cli` against an existing Debian-based WordPress install, something like `--user 33:33` is likely going to be necessary (possibly also something like `-e HOME=/tmp` depending on the `wp` command invoked and whether it tries to use `~/.wp-cli`). See [docker-library/wordpress#256](https://github.com/docker-library/wordpress/issues/256) for more discussion around this.
+
+## Configuring PHP directives
+
+See [the "Configuration" section of the `php` image documentation](https://hub.docker.com/_/php/).
+
+Additionally you could create a `custom.ini` with the desired parameters and place it in the `$PHP_INI_DIR/conf.d/` directory.
```
tianon, line 88: This is good, but I think we should probably include a few explicit keywords users might <kbd>Ctrl+F</kbd> for, like `php.ini` and `upload_max_filesize`; maybe something like this?

```suggestion
See [the "Configuration" section of the `php` image documentation](https://hub.docker.com/_/php/).

For example, to adjust common `php.ini` flags like `upload_max_filesize`, you could create a `custom.ini` with the desired parameters and place it in the `$PHP_INI_DIR/conf.d/` directory.
```

PR #1780, opened 2020-08-22
add port exposing to readme

bresche:
if you want to connect you have to expose some port


tianon:
What we'd typically recommend to users instead is to use a shared Docker network and connect from their application directly so that the database is not exposed outside the host / Docker.

PR #1781, opened 2020-08-24
Ensure exit code on git diff can cause failure as well

yosifkit:
Found in #1780 (https://github.com/docker-library/docs/pull/1780/checks?check_run_id=1015432599#step:3:5).


tianon:
Yay now it succeeds. :sweat_smile:

PR #1782, opened 2020-08-25
[ArangoDB] Minor adjustments to content.md

Simran-B:
@KVS85 @fceller @joerg84

### DIFF
--- a/arangodb/content.md
+++ b/arangodb/content.md
@@ -153,9 +153,7 @@ See more information about [Configuration](https://www.arangodb.com/docs/stable/
 
 ArangoDB supports two different storage engines from version 3.2 to 3.6. You can choose them while instantiating the container with the environment variable `ARANGO_STORAGE_ENGINE`. With `mmfiles` you choose the classic storage engine (not available in 3.7 and later), `rocksdb` will choose the storage engine based on [RocksDB](http://rocksdb.org/). The default choice is `rocksdb` from version 3.4 on.
 
-ArangoDB use the volume `/var/lib/arangodb3` as database directory to store the collection data and the volume `/var/lib/arangodb3-apps` as apps directory to store any extensions. These directories are marked as docker volumes.
-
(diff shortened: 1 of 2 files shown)


arangodb/content.md
```diff
@@ -151,7 +151,7 @@ See more information about [Configuration](https://www.arangodb.com/docs/stable/
 
 ## Persistent Data
 
-ArangoDB supports two different storage engines as of ArangoDB 3.2. You can choose them while instantiating the container with the environment variable `ARANGO_STORAGE_ENGINE`. With `mmfiles` you choose the classic storage engine, `rocksdb` will choose the newly introduced storage engine based on [RocksDB](http://rocksdb.org/). The default choice is `mmfiles`.
+ArangoDB supports two different storage engines from version 3.2 to 3.6. You can choose them while instantiating the container with the environment variable `ARANGO_STORAGE_ENGINE`. With `mmfiles` you choose the classic storage engine (not available in 3.7 and later), `rocksdb` will choose the storage engine based on [RocksDB](http://rocksdb.org/). The default choice is `rocksdb` from version 3.4 on.
```
KVS85, line 154: `from version 3.5 to 3.6` once 3.4- are EOLed.

Simran-B: If EoL'ed means to treat earlier versions as if they never existed, then the 2.x content also has to go:

> Please note that the old version 2.x used /var/lib/arangodb and /var/lib/arangodb-apps. We will refer to the 3.x variant in this document. In case you are starting a 2.x image just replace it with the 2.x variant.

joerg84: IMO we could remove this or?

joerg84 (approved):

fceller (approved):

fceller (approved):

arangodb/content.md
Simran-B: Removed the 2.x paragraph. I would keep the 3.2-3.6 MMFiles paragraph until 3.6 is EoL, then we can remove it entirely (or have a single sentence about the RocksDB storage engine).

PR #1787, opened 2020-08-28
Change monica logo

asbiin:
Current logo does not work on the https://hub.docker.com/_/monica/ page.
The new one is svg format, which is way better too.


tianon:
:+1:

PR #1789, opened 2020-09-02
Clarifying mysql readme with information about graceful shutdowns 

theothermattm:
Per https://github.com/docker-library/mysql/issues/47

This includes general information about graceful shutdowns and a practical example from @jeremyVignelles from [this comment](https://github.com/docker-library/mysql/issues/47#issuecomment-147397851)

First contribution so suggestions are welcome.

### DIFF
--- a/mysql/content.md
+++ b/mysql/content.md
@@ -186,3 +186,43 @@ For restoring data. You can use `docker exec` command with `-i` flag, similar to
 ```console
 $ docker exec -i some-%%REPO%% sh -c 'exec mysql -uroot -p"$MYSQL_ROOT_PASSWORD"' < /some/path/on/your/host/all-databases.sql
 ```
+
+## Graceful Shutdowns
+
+Running MySQL in a Docker container has some peculiarities around stopping containers that may happen if using this container for local development. For instance, sending a `CTRL+C` to the container may not allow mysqld to shutdown cleanly and may result in corrupt data.
+
+It is recommended that in order to maintain consistent shutdown behavior that you create a custom image with an overridden entrypoint script which will `trap` the `SIGINT` and `SIGTERM` signals and gracefully shutdown mysql. The following is an example which will do this:
+
+Custom Dockerfile:
+
+	FROM mysql:latest
+	
+	ENV MYSQL_ROOT_PASSWORD password
+	ENV MYSQL_DATABASE databasename
+	ENV MYSQL_USER user
+	ENV MYSQL_PASSWORD password
+	
+	# Entrypoint overload to catch the ctrl+c and stop signals
+	ENTRYPOINT ["/bin/bash", "/run.sh"]
+	CMD ["mysqld"]
+	
+	COPY run.sh /run.sh
+
+Custom entrypoint:
+
+	#!/bin/bash
+	# A wrapper around /entrypoint.sh to trap the SIGINT signal (Ctrl+C) and forwards it to the mysql daemon
+	# In other words : traps SIGINT and SIGTERM signals and forwards them to the child process as SIGTERM signals
+	
+	asyncRun() {
+	    "$@" &
+	    pid="$!"
+	    trap "echo 'Stopping PID $pid'; kill -SIGTERM $pid" SIGINT SIGTERM
+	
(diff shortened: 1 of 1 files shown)


yosifkit:
Non-working `ctrl-c` is not significant enough to provide a hacky script workaround in the documentation. `mysqld` is not meant to be run interactively.  The mysql container correctly shuts down when given a `SIGTERM` (the [default "stop signal"](https://docs.docker.com/engine/reference/commandline/stop/)) and so my comment on the issue is still true:

> yes and no. Yes, mysql will receive the `SIGTERM` and begin cleaning up to exit. But if it takes too long (10 seconds by default) it will get a `SIGKILL` from [docker stop](https://docs.docker.com/engine/reference/commandline/stop/).  So if your database needs longer to clean up, either increase the timeout to `docker stop` or use `docker kill -s TERM` and do your own timeout process.
> \- https://github.com/docker-library/mysql/issues/47#issuecomment-372766241

See also https://github.com/docker-library/mysql/issues/47#issuecomment-339242758 and https://github.com/docker-library/mariadb/issues/82#issuecomment-339242301:
> If you are running in non-detached mode, you may send a SIGQUIT signal to the process. This is typically mapped to your `Control-\` key combination.

Although `SIGQUIT` is not documented on their page (https://dev.mysql.com/doc/refman/8.0/en/unix-signal-response.html), it seems to behave the same as `TERM`.

So it might make sense to mention about `Control-\` in the documentation, but it is more of a general Linux thing and not really specific to MySQL or Docker.

theothermattm:
Thanks @yosifkit !

> The mysql container correctly shuts down when given a SIGTERM

Thanks for helping me understand that, with some more testing I'm realizing that whatever problem we're having is not related to this... I'll keep digging, I'm now wondering if in certain circumstances our database is taking too long to shutdown and is corrupting mysql files after a `SIGKILL` is sent.  Our problem is that the issue is very intermittent so it's hard to track down.

> mysqld is not meant to be run interactively. 

Completely agree. But, this mysql docker image is used by a lot of developers to run a database on their laptop for development.  This will inevitably involve running `ctrl+c` when attached at some point.  IMO, this is an important use case and these "gotchas" with shutdowns should be documented somehow.  

I'm wondering if instead it makes sense to document the different shutdown behaviors in the docs for the image.  Something like this:

# Graceful Database Shutdowns

[mysqld responds to different termination commands](https://dev.mysql.com/doc/refman/8.0/en/unix-signal-response.html) differently.  It's important to remember that when Docker containers are stopped, different termination signals are sent to the running process, which therefore will affect how MySQL is shutdown or not.  If MySQL is not shut down gracefully, data could potentially be corrupted.

The default behavior of `docker stop` (and `docker-compose stop` is to [send a `SIGTERM` signal](https://docs.docker.com/engine/reference/commandline/stop/).  This should allow MySQL to shut down gracefully.  However, in some instances the shutdown process may take longer than the [default `docker stop` grace period](https://docs.docker.com/engine/reference/commandline/stop/) before sending a `SIGKILL` signal.  Therefore, you may want to increase the shutdown grace period timeout when running the container to allow for a graceful shutdown.

It is also worth noting, that if this image is used in attached mode (for instance while using the image for local development) that sending a `ctrl+c` will send a `SIGINT` signal which is not handled by mysqld and may result in a non-graceful shutdown.  It's also worth noting, if using this image with docker-compose, docker-compose will translate an attached `ctrl+c` command into a `docker stop` command and properly send a `SIGTERM` signal.

tianon:
IMO, the most I'd want to document here is that if you run interactively, you'll have to use `Ctrl+\` (or another terminal) to stop the container, since that's a really common gotcha that's kind of unique to Docker (since it's otherwise really uncommon to run `mysqld` interactively) -- anything more and we're just repeating official Docker documentation and/or official MySQL documentation.

tianon:
Thanks for the contribution and discussion!

I'm going to close this for now, but I'm still open to a note about `Ctrl+\` likely being a way to send `SIGQUIT` to an interactive container on Linux (I say "likely" because it probably depends on the terminal, terminal emulator, etc - Linux is so fun :joy:). :+1:

PR #1791, opened 2020-09-07
Friendica - Add Docker Secrets

nupplaphil:
Part of https://github.com/docker-library/official-images/pull/8678


friendica/content.md
```diff
@@ -140,6 +140,66 @@ The Friendica image supports auto configuration via environment variables. You c
...
+version: '3.2'
+
+services:
+  db:
+    image: mariadb
+    restart: always
+    volumes:
+      - db:/var/lib/mysql
+    environment:
+       - MYSQL_DATABASE_FILE=/run/secrets/mysql_db
```
yosifkit, line 157: This and the one in `app` ought to be `/run/secrets/mysql_database` to match the secret name.

nupplaphil: nice catch

PR #1792, opened 2020-09-08
Add Sentry deprecation notice

BYK:
See docker-library/official-images#8673

### DIFF
--- a/sentry/deprecated.md
+++ b/sentry/deprecated.md
@@ -1,0 +1,1 @@
+This image is deprecated in favor of the full installation explained over at [Self-Hosted Sentry Docs](https://develop.sentry.dev/self-hosted/) (last updated July 2019; [getsentry/docker-sentry#189](https://github.com/getsentry/docker-sentry/pull/189).


BYK:
Ping @tianon

BYK:
@tianon I don't see the deprecation notice over at https://hub.docker.com/_/sentry/

Is there anything else we need to do?

yosifkit:
It's not showing up because that is the "store" page that is masking the official-images page as mentioned in this [comment](https://github.com/getsentry/docker-sentry/issues/211#issuecomment-628285526). I'll try to hit some Docker folks to see if they can drop the store page part.

mattrobenolt:
Hey @yosifkit I submit a support ticket a bunch of hours ago into Docker folks to see what needs to be done to get rid of this and haven't heard anything back yet. So maybe if you can nudge them, that'd be good. I obviously give consent. :)

BYK:
@yosifkit @tianon is there any other channel that we should be contacting? It's been more than a week now without any updates and its been over 2 weeks since we submitted this pull request.

mattrobenolt:
I guess also worth adding, the support ticket I submit did not get a response back.

yosifkit:
I've poked some Docker folks again.

yosifkit:
Docker Hub page is fixed now! :tada:

PR #1793, opened 2020-09-09
Add note about stack immediately after the example code

igrep:
A Haskell newbie loving Docker got confused by the error.
I think we should guide him/her to the section about the error near from the example code of stack.


igrep:
Sorry, I'll go again after fixing content.md

PR #1794, opened 2020-09-09
Add note about stack immediately after the example code

igrep:
A Haskell newbie loving Docker got confused by the error.
I think we should guide him/her to the section about the error near from the example code of stack.

(Retry of https://github.com/docker-library/docs/pull/1793)

### DIFF
--- a/haskell/content.md
+++ b/haskell/content.md
@@ -36,6 +36,8 @@ RUN stack install pandoc pandoc-citeproc
 ENTRYPOINT ["pandoc"]
 ```
 
+**NOTE**: You might get a "Compiler version mismatched" error in some cases. See [Considerations for Stack](#considerations-for-stack) for details.
+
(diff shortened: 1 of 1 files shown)


yosifkit:
The link will work fine here on GitHub, but will go nowhere on Docker Hub. Maybe add a "section below" so they know where to look when the link fails?

```diff
-See [Considerations for Stack](#considerations-for-stack) for details.
+See the [Considerations for Stack](#considerations-for-stack) section below for details.
```

tianon:
(also cc `haskell` image maintainers @psftw @hvr)

tianon:
friendly ping @psftw @hvr

igrep:
ping

psftw:
Sorry, I don't see the value of this change, so I'm against it.

PR #1795, opened 2020-09-12
Update URL to the doc

obax:
URL is has changes without a redirection, this PR updates it to the current link

### DIFF
--- a/consul/content.md
+++ b/consul/content.md
@@ -9,7 +9,7 @@ Consul is a distributed, highly-available, and multi-datacenter aware tool for s
 
 # Consul and Docker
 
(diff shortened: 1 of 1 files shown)


tianon:
This file is auto-generated from `consul/content.md` -- can you move the change to that file instead?

also cc `consul` image maintainers; @mikemorris @s-christoff @crhino @i0rek @mkeeler
(no humans listed in `library/consul`'s `Maintainers:` and GitHub doesn't let me tag the team, so I've tagged the most recent humans to touch the file :sweat_smile:)

mkeeler (approved):
The new link location is the correct place. We recently refactored our website but missed the change here.

tianon:
I've pushed an amended commit to move the change from `README.md` to `content.md`.

Thanks for confirming, @mkeeler, and thanks for the contribution, @obax!

PR #1802, opened 2020-09-25
caddy: Updating docs for builder image changes

hairyhenderson:
Changes relevant to https://github.com/docker-library/official-images/pull/8735

Many thanks to @francislavoie for writing this 😉

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -50,7 +50,7 @@ The default `Caddyfile` only listens to port `80`, and does not set up automatic
 
 ```console
 $ docker run -d -p 80:80 -p 443:443 \
-    -v /site:/usr/share/caddy \
+    -v /site:/srv \
     -v caddy_data:/data \
     -v caddy_config:/config \
     %%IMAGE%% caddy file-server --domain example.com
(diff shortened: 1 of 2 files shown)


caddy/content.md
```diff
@@ -66,33 +66,33 @@ Most users deploying production sites will not want to rely on mounting files in
...
 #### Adding custom Caddy modules
 
-Caddy is extendable through the use of "modules". See https://caddyserver.com/docs/extending-caddy for full details.
+Caddy is extendable through the use of "modules". See https://caddyserver.com/docs/extending-caddy for full details. You can find a list of available modules on [the Caddy website's download page](https://caddyserver.com/download).
 
 You can use the `:builder` image as a short-cut to building a new Caddy binary:
 
 ```Dockerfile
-FROM %%IMAGE%%:2.0.0-builder AS builder
+FROM %%IMAGE%%:<version>-builder AS builder
```
francislavoie, line 82: @yosifkit I hope you don't mind me asking like this on a merged PR. 

Is there any way to dynamically replace this `<version>` with a placeholder? I dug through the scripts in this repo and I didn't find any relevant placeholders.

Ideally, I'd like a `%%LATEST_VERSION%%` placeholder that would basically be the version number that the `latest` tag shares.

I've already gotten a question from a user recently who is new to Docker asking "do I need to replace `<version>`?" so I think this should be improved.

tianon: Unfortunately, that's a somewhat complicated problem for us to solve generally, especially since in many cases using the "most specific" versioned tag isn't actually what we recommend users should use (PHP for example - most users should use X.Y not X.Y.Z so they continue to get patch updates and because that's the level at which compatibility is supposed to be kept between versions).

PR #1805, opened 2020-09-29
Note that PyPi does not provide wheels for alpine images

debuglevel:
PyPi does only provide prcompiled "wheels" for glibc based distributions. On Alpine, compilation will be done during "pip install" which slows down an image build. Of course feel free to change anything, this is just a suggestion; but users should know about this issue before they decide for the Alpine variant.


debuglevel:
I guess that's not the way then. 😐

PR #1813, opened 2020-10-09
Remove contradictory sentence in Environment Vars

ocularrhythm:
The sentence stated that no environment variables are required to use this image. However, the very first variable - POSTGRES_PASSWORD is stated as required. This is a direct contradiction. This change removes that contradiction and clarifies that the password is required.


tianon:
Good catch -- this is a case where the actual implementation has drifted from the initial. :smile:

Thanks for the contribution!

PR #1814, opened 2020-10-09
Aerospike Dockerfile Default Volume info

volmarl:


aerospike/content.md
```diff
@@ -62,6 +62,8 @@ A full example:
 $ docker run -d -v <DIRECTORY>:/opt/aerospike/data --name aerospike -p 3000:3000 -p 3001:3001 -p 3002:3002 -p 3003:3003 %%IMAGE%%
 ```
 
+__Note:__ As of version 5.1.0.10 the Dockerfile no longer has /opt/aerospike/data as a Docker Volume by default.
```
tianon, line 65: Per the CI failure, this format needs to be adjusted slightly:

```suggestion
**Note:** As of version 5.1.0.10 the Dockerfile no longer has /opt/aerospike/data as a Docker Volume by default.
```

PR #1821, opened 2020-10-23
Add "Recommended Runtime Settings" to Docker description

tianon:
This is intended as an alternative approach to https://github.com/docker-library/docker/pull/258.

cc @thaJeztah @cpuguy83 @AkihiroSuda

### DIFF
--- a/docker/content.md
+++ b/docker/content.md
@@ -188,6 +188,23 @@ $ docker run --privileged --name some-docker -d \
 	%%IMAGE%%:dind --storage-driver overlay2
 ```
 
+## Runtime Settings Considerations
+
+Inspired by the [official systemd `docker.service` configuration](https://github.com/docker/docker-ce-packaging/blob/57ae892b13de399171fc33f878b70e72855747e6/systemd/docker.service#L30-L45), you may want to consider different values for the following runtime configuration options, especially for production Docker instances:
+
+```console
+$ docker run --privileged --name some-docker -d \
+	... \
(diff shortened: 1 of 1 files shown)


cpuguy83 (commented):
I would maybe preface the recommendation with some caveats:

1. By default, it will indeed inherit these values from the daemon
2. ... except in the case of ulimits *if* there are default values set on the daemon config (`--default-ulimit nofile:999999`)
3. The recommendations are how we recommend to run docker in production, and may not be needed for short-lived or non-critical instances.

thaJeztah:
Thanks Tianon! I Forgot about opening a follow up PR 😓

Also a bit on the fence on calling it "recommended", as it really would depend on the use case 🤔

Setting OOM adjust to the same value as the daemon itself would likely not be great as it would then increase the chance that dockerd is killed before the dind container is killed.

I guess some more context to allow the user to make the right decision for their use would be good, but I know there's limited space for documentation on Hub 😔

tianon:
Fair, thanks for the suggestions!  I did a little wordsmithing in https://github.com/docker-library/docs/compare/5dcd7c1aba4e1a5c0e5a05509bcf18a77ee58fe9..1aa9c4e79a6370d84ea88241f8b775c01a626b02 -- is this any better?

cpuguy83 (approved):
LGTM!

thaJeztah (approved):
LGTM, thanks!!

PR #1822, opened 2020-10-28
perl: Add note about coexisting with Debian's /usr/bin/perl

zakame:
A few users of docker-perl have stumbled upon (and getting confused
with) the default /usr/bin/perl shipped by Debian, so put a short note
here explaining why it remains and also better describe how docker-perl
and its tooling is installed/configured.

References:

- https://github.com/Perl/docker-perl/issues/26
- https://github.com/Perl/docker-perl/issues/93
- https://github.com/Perl/docker-perl/issues/94


tianon (approved):

zakame:
Wow, that was fast, thanks @tianon ! :heart: :tada:

PR #1827, opened 2020-11-03
Set explicit value for INFLUXDB_HTTP_AUTH_ENABLED

matletix:
Setting the `INFLUXDB_HTTP_AUTH_ENABLED` variable to something other than `true` (eg. `INFLUXDB_HTTP_AUTH_ENABLED=1`) can result in errors in the init script. More details here : https://github.com/influxdata/influxdata-docker/issues/224

### DIFF
--- a/influxdb/README.md
+++ b/influxdb/README.md
@@ -197,7 +197,7 @@ Automatically initializes a database with the name of this environment variable.
 
 ##### INFLUXDB_HTTP_AUTH_ENABLED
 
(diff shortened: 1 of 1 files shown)


tianon:
cc @jsternberg (`influxdb` image maintainer)

Also, this change will need to be moved to `content.md` (as `README.md` is a generated file).

matletix:
> cc @jsternberg (`influxdb` image maintainer)
> 
> Also, this change will need to be moved to `content.md` (as `README.md` is a generated file).

Ok, closing this and I created #1828 with the changes in `content.md`

kidkidup:
#1857

PR #1830, opened 2020-11-07
Update Rakudo Star

sorairolake:
- Fix missing rename to Raku
- Change the scheme of external links to `https`
- Add a link to the [Language Specification](https://github.com/Raku/roast)
- Add a link to a [Wikipedia article](https://en.wikipedia.org/wiki/Rakudo)
- Use [GFM's extended Autolinks](https://github.github.com/gfm/#autolinks-extension-)

### DIFF
--- a/rakudo-star/content.md
+++ b/rakudo-star/content.md
@@ -2,13 +2,11 @@
 
 Rakudo Star is a Raku (formerly known as Perl 6) distribution designed for use by early adopters of the language. It includes a virtual machine (the JVM or MoarVM), the Rakudo compiler, and a suite of modules that users may find useful. This image includes the MoarVM backend for the compiler.
 
-Project homepage: [http://rakudo.org](http://rakudo.org)
+-	Project homepage: https://rakudo.org/
+-	Raku Language Specification: https://github.com/Raku/roast
+-	Raku Language Documentation: https://docs.raku.org/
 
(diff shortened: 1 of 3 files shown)


rakudo-star/content.md
```diff
@@ -2,13 +2,13 @@
 
 Rakudo Star is a Raku (formerly known as Perl 6) distribution designed for use by early adopters of the language. It includes a virtual machine (the JVM or MoarVM), the Rakudo compiler, and a suite of modules that users may find useful. This image includes the MoarVM backend for the compiler.
 
-Project homepage: [http://rakudo.org](http://rakudo.org)
+-	Project homepage: https://rakudo.org/
+-	GitHub repository: https://github.com/rakudo/star
+-	The Dockerfile responsible: %%GITHUB-REPO%%/blob/master/Dockerfile
```
tianon, line 7: All the `Dockerfile`s are linked above this in the generated description (pinned to the specific commit in `library/rakudo-star`, even), so I'd suggest this is probably worth just dropping completely instead.

tianon:
FYI image `Maintainer:` @moritz :+1:

(Going to merge without explicit approval since it looks like a good change to me, and seems to correspond to the latest image update -- feel free to let us know if we should revert/update. :+1:)

moritz:
All good, no need to revert anything :-)

PR #1844, opened 2020-12-07
Environment variables in docker-compose section

tetebueno:
Docker-compose file in the example was missing database configuration environment variables of the app container, making the stack to use SQLite by default instead of the DB container.


tetebueno:
Oops, I'll update `content.md`.

PR #1845, opened 2020-12-07
Environment variables in docker-compose section

tetebueno:
Docker-compose file in the example was missing database configuration environment variables of the app container, making the stack to use SQLite by default instead of the DB container.

### DIFF
--- a/nextcloud/content.md
+++ b/nextcloud/content.md
@@ -216,9 +216,14 @@ services:
       - 8080:80
     links:
       - db
+    restart: always
     volumes:
       - nextcloud:/var/www/html
-    restart: always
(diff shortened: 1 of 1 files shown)


tianon:
cc @tilosp @J0WI

J0WI:
This has no effect unless `NEXTCLOUD_ADMIN_*` are also set.

@tetebueno could you make a PR against https://github.com/nextcloud/docker/blob/master/README.md? These docs are just a copy of it.

tetebueno:
Sure.

Closing in favor of https://github.com/nextcloud/docker/pull/1316.

PR #1847, opened 2020-12-13
Redis: update configuration bindmount example.

yossigo:
The provided example may be misleading and will not work in certain
scenarios, where Redis attempts to create additional configuration files
or rewrite existing ones.

See redis/redis#8172.


tianon:
Oh interesting, thanks for the correction/update! :+1: :heart:

PR #1850, opened 2020-12-15
optimization in git clone using --depth

Rajpratik71:
optimize the git clone using --depth flag in term of size of clone
and also in term's of time taken to fetch the files and commit history
of whole repository .

More detail can be found at blog
https://www.atlassian.com/git/tutorials/big-repositories

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,10 +32,10 @@ ENV LANG C.UTF-8
 
 RUN mkdir ~/.vim ~/.vim/bundle ~/.vim/autoload
 RUN set -x \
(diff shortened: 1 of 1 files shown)


tianon:
Thanks for the contribution!

However, I'm confused - none of these repositories are particularly big (the final `/root/.vim` directory is ~1.4MB), and they all clone very fast.

Additionally, it's been so long since we used the functionality these are being cloned for that I have to wonder if we should remove the `vim`-related functionality from `push.pl` entirely (especially with the new "official" `hub-tool` coming soon).

tianon:
Thank you for the contribution nevertheless!! :heart:

PR #1855, opened 2020-12-25
Add note about read only possibility

NicolasCARPi:
Running containers as read only filesystems is considered best practices. Add a note to mention this possibility with this image.

### DIFF
--- a/haproxy/content.md
+++ b/haproxy/content.md
@@ -41,6 +41,8 @@ $ docker run -d --name my-running-haproxy my-haproxy
 
 You may need to publish the ports your HAProxy is listening on to the host by specifying the -p option, for example -p 8080:80 to publish port 8080 from the container host to port 80 in the container. Make sure the port you're using is free.
 
+Note: for added security, it is possible to run the container in read-only mode, as HAProxy doesn't need to write to the filesystem. You can pass the `--read-only` flag to the `docker run` command or add `read_only: true` to your `docker-compose` configuration file.
+
 ## Directly via bind mount
 
 ```console


tianon:
I agree that this is a good idea, but I'm not sure this is the best place to document it -- this is more generic Docker usage documentation? (We try not to reproduce Docker's documentation too much, unless it adds some specific value to the image for us to do so, but I don't think this one is particularly unique to the `haproxy` image.)

NicolasCARPi:
I don't see another place to document it. As the official doc doesn't speak about Docker at all. When I was looking for this information, I couldn't find it. The README of the Docker image seems to me like a good place for such information. Because that's the first document one reads before using this image.

tianon:
I meant
https://docs.docker.com/engine/reference/commandline/run/#mount-volume--v---read-only
(Docker's own official documentation).

NicolasCARPi:
You are linking the documentation about read-only mounts (which for this image would be the config folder for instance), I'm talking about having the whole container as read-only filesystem. Being able to have the whole container read-only is quite specific to an image, as it will break some images in unexpected ways, but for others it's fine. This is the information that is missing: "can I run this container as read-only?", and this is what I want to add to the doc :)

tianon:
It's both -- from that link:

> ```console
> $ docker run --read-only -v /icanwrite busybox touch /icanwrite/here
> ```
> 
> Volumes can be used in combination with `--read-only` to control where a container writes files. The `--read-only` flag mounts the container’s root filesystem as read only prohibiting writes to locations other than the specified volumes for the container.

NicolasCARPi:
I see. But then the information that the haproxy image **can** be run in read-only is nowhere to be found, isn't it? And I think for a webservice it's pretty neat to be able to run it on a read-only container, as it makes successful attacks much harder!

tianon:
Right, my point is that this isn't specific to HAProxy at all -- any
container can be run as read-only, and it's definitely a great way to
reduce the attack surface.  There's nothing particular about HAProxy
that makes it any more or less well-suited to this practice.

NicolasCARPi:
I understand your point but I disagree with the part that "there is nothing particular about HAProxy", as in fact there is because it doesn't write anything to the filesystem during its operation, unlike nginx for instance. So this is why I think it's worth mentionning that this particular image can be run in RO mode.

yosifkit:
While you say it may not be designed to run in a read-only container, even the nginx image can be run as read-only. You just have to run it a few times to figure out where the mutable state is. So it really is nothing special to `haproxy` to be able to run in a read-only container. With a little sleuthing, just about every image can be run in a read-only container.

```console
$ docker run -it --rm --read-only nginx
Unable to find image 'nginx:latest' locally
latest: Pulling from library/nginx
a076a628af6f: Pull complete 
0732ab25fa22: Pull complete 
d7f36f6fe38f: Pull complete 
f72584a26f32: Pull complete 
7125e4df9063: Pull complete 
Digest: sha256:10b8cc432d56da8b61b070f4c7d2543a9ed17c2b23010b43af434fd40e2ca4aa
Status: Downloaded newer image for nginx:latest
...
2021/01/14 00:08:42 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (30: Read-only file system)
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (30: Read-only file system)
$ docker run -it --rm --read-only --tmpfs /var/cache/nginx  nginx
...
2021/01/14 00:09:23 [emerg] 1#1: open() "/var/run/nginx.pid" failed (30: Read-only file system)
nginx: [emerg] open() "/var/run/nginx.pid" failed (30: Read-only file system)
$ # /var/run is just a symlink to /run
$ docker run -it --rm --read-only --tmpfs /var/cache/nginx/ --tmpfs /run/ nginx
...
/docker-entrypoint.sh: Configuration complete; ready for start up
(and then it logs to stdout when requests happen)
```

PR #1858, opened 2021-01-05
docs: fix typo in Caddy's readme

dunglas:


tianon:
As noted by the CI failure, the `README.md` file is auto-generated, so this change needs to be moved to `content.md` instead.

(cc/FYI `caddy` image maintainer: @hairyhenderson)

hairyhenderson:
Thanks for the fix @dunglas - like @tianon said, just flip it to `content.md` otherwise this LGTM 😉

PR #1859, opened 2021-01-09
[nats-streaming] Release v0.20.0

kozlovic:
Details can be found [here](https://github.com/nats-io/nats-streaming-server/releases/tag/v0.20.0)

### DIFF
--- a/nats-streaming/content.md
+++ b/nats-streaming/content.md
@@ -66,29 +66,29 @@ $ docker run -d -p 4222:4222 -p 8222:8222 %%IMAGE%%
 Output that you would get if you had started with `-ti` instead of `d` (for daemon):
 
 ```bash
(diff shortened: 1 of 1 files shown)


tianon:
(Next time, it would be helpful to include a link to/from https://github.com/docker-library/official-images/pull/9397 so we can connect the dots since we can't merge this one until that one is, so it ends up in a second pass and we need to go back and verify the other merged. :sweat_smile:)

kozlovic:
@tianon I usually do, but last time my two last doc PRs were closed: https://github.com/docker-library/docs/pull/1752 and https://github.com/docker-library/docs/pull/1826 without being merged. I know it says that I closed them, but I don't recall doing that. So I started to think that maybe GH was closing them because the official image that referenced them was merged, so it was auto-closing).
Next release I will get back to reference the doc in the official images PR and check if the doc PR gets closed before being merged again. Thanks!

PR #1872, opened 2021-02-09
Start Ghost in development mode

mmf-amarcos:
A pair off lines on the stack.yml to give some info on how to start Ghost in development mode if the image is to be used for development purposes, like theme edition.


mmf-amarcos:
Checks failed correctly. Editing the stack.yml file instead...

PR #1873, opened 2021-02-09
Start Ghost in development mode

mmf-amarcos:
A suggestion on how to start Ghost in development mode for theme development or similar tasks

### DIFF
--- a/ghost/stack.yml
+++ b/ghost/stack.yml
@@ -17,10 +17,10 @@ services:
       database__connection__user: root
       database__connection__password: example
       database__connection__database: ghost
-      # if you want to start ghost in development mode uncomment next line
-      # NODE_ENV: development
       # this url value is just an example, and is likely wrong for your environment!
       url: http://localhost:8080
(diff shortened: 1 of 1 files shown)


ghost/stack.yml
```diff
@@ -17,6 +17,8 @@ services:
       database__connection__user: root
       database__connection__password: example
       database__connection__database: ghost
+      # if you want to start ghost in development mode uncomment next line
+      # NODE_ENV: development
       # this url value is just an example, and is likely wrong for your environment!
       url: http://localhost:8080
```
tianon, line 23: Thanks for the suggestion!  However, I'd suggest we swap this to the end (given most users _shouldn't_ specify it), and to adjust the comment to be more clear about why it needs to be specified (since the documentation we link above claims `development` is the default):

```suggestion
      # this url value is just an example, and is likely wrong for your environment!
      url: http://localhost:8080
      # contrary to the default mentioned in the linked documentation, this image defaults to NODE_ENV=production (so development mode needs to be explicitly specified if desired)
      #NODE_ENV: development
```

mmf-amarcos: Perfect from my side.
Commited

tianon:
Thanks! :+1:

PR #1876, opened 2021-02-12
mediawiki: Update maintainer description

legoktm:
https://phabricator.wikimedia.org/T266186 raise some questions on who is
actually maintaining this image. I think it makes more sense to describe this
as something done by MediaWiki community members rather than as by
the Wikimedia Foundation.

cc @davidbarratt and @addshore

### DIFF
--- a/mediawiki/maintainer.md
+++ b/mediawiki/maintainer.md
@@ -1,1 +1,1 @@
-[Wikimedia Foundation & Docker Community](%%GITHUB-REPO%%)
+[MediaWiki community & Docker Community](%%GITHUB-REPO%%)


addshore:
Looks good to me

addshore (approved):

tianon:
That's two out of three image maintainers saying this is acceptable -- if @davidbarratt disagrees, maybe there'll be a competing PR in our future? :smile: :pray:

(I'm guessing it's fine, which is why I'm merging, but happy to revert if there's dissent while y'all work it out further. :+1:)

davidbarratt:
Seems fine.

legoktm:
Thanks everyone :)

PR #1880, opened 2021-02-24
MariaDB: docs update + repo move

grooverdan:
Note: I have been working with MariaDB Corporation on this. Expect an official notice from Julien.


yosifkit (approved):

tianon:
The other half of this is going to be updating https://github.com/MariaDB/mariadb-docker/blob/f0e04a7e0e4b7d83691efc889865da8d6ae5f00e/generate-stackbrew-library.sh#L46-L50 and/or https://github.com/docker-library/official-images/blob/cfbf7190dfafeb9b43301d4f9dfb17ff1d48f256/library/mariadb#L1-L5 :+1:

Congrats and welcome! :partying_face:

PR #1882, opened 2021-02-26
Rearrange InfluxDB docs so upgrade info is immediately visible.

danxmoran:
docker-library/official-images#9594 has caused a bit of a splash among community members who were using the `latest` tag 😓 mitigate the situation and prevent repeat performances in two ways:
1. Pull the info about upgrading from 1.x to the top of the docs, so anyone who upgrades unintentionally and wants to stick it out can immediately see what they need to do
2. Add version tags to all `%%IMAGE%%` references, so we aren't encouraging anybody to copy-paste a usage of `latest`


danxmoran:
@tiangolo @yosifkit would one of you mind looking at this, since you reviewed the original PR on the library side?

PR #1888, opened 2021-03-10
Fixup InfluxDB 2.x docs.

danxmoran:
* Remove use of `:ro`, which doesn't work w/ entrypoint's `chmod`
* Remove reference to "above" from section that was moved to top

Fixes influxdata/influxdata-docker#468


tianon:
Sorry for the delay!

PR #1898, opened 2021-03-20
Update MariaDB docs run command

iwschris:
The current docs for the MariaDB image incorrectly use the `--port` long form flag which is no longer supported:

image


tianon:
Ah, good catch!  cc @grooverdan

PR #1903, opened 2021-03-29
Change Caddy reload command to not return an error

LukasDoesDev:


LukasDoesDev:
The error returned is:
```
OCI runtime exec failed: exec failed: container_linux.go:367: starting container process caused: exec: "-w": executable file not found in $PATH: unknown
```

PR #1904, opened 2021-03-31
Update the CrateDB Docker image homepage

norosa:
Specifically:

- Update wording to reflect the `crate/crate` `README.rst` file
- Adds some small copyedits
- Adds a link to the screenshot we use


crate/README-short.txt
```diff
@@ -1 +1 @@
-CrateDB is a distributed SQL database handles massive amounts of machine data in real-time.
+CrateDB is a distributed SQL database that makes it simple to store and analyze massive amounts of machine data in real-time.
```
tianon, line 1: Unfortunately, this needs to be less than 100 characters. :sweat_smile:

PR #1911, opened 2021-04-08
Adjust WP-CLI docs for our customized `wp-config.php`

yosifkit:
Fixes https://github.com/docker-library/wordpress/issues/594

### DIFF
--- a/wordpress/variant-cli.md
+++ b/wordpress/variant-cli.md
@@ -8,7 +8,12 @@ The simplest way to use it with an existing WordPress container would be somethi
 $ docker run -it --rm \
 	--volumes-from some-wordpress \
 	--network container:some-wordpress \
-	wordpress:cli user list
+	-e WORDPRESS_DB_USER=... \
+	-e WORDPRESS_DB_PASSWORD=... \
+	# [and other used environment variables]
+	%%IMAGE%%:cli user list
 ```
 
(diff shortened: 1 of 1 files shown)


wordpress/variant-cli.md
```diff
@@ -8,7 +8,12 @@ The simplest way to use it with an existing WordPress container would be somethi
...
-	wordpress:cli user list
+	-e WORDPRESS_DB_USER=... \
+	-e WORDPRESS_DB_PASSWORD=... \
+	# [and other used environment variables]
+	%%IMAGE%%:cli user list
 ```
 
 Generally speaking, for WP-CLI to interact with a WordPress install, it needs access to the on-disk files of the WordPress install, and access to the database (and the easiest way to accomplish that such that `wp-config.php` does not require changes is to simply join the networking context of the existing and presumably working WordPress container, but there are many other ways to accomplish that which will be left as an exercise for the reader).
+
+**NOTE:** Since March 2021, WordPress images use a customized `wp-config.php` that pulls the values directly from the environment variables defined above (see `wp-config-docker.php` in [docker-library/wordpress#572](https://github.com/docker-library/wordpress/pull/572) and [docker-library/wordpress#577](https://github.com/docker-library/wordpress/pull/577)). As a result of reading environment variables directly, the cli container also needs the same set of environment variables to properly evaluate `wp-config.php`.
```
tianon, line 19: Was this second link intended to be docker-library/wordpress#577 or docker-library/wordpress#557? (maybe both?)

yosifkit: I figured that linking to 557 would confuse some users, since it is only in beta and so chose the ones that modify the release versions.

tianon: Sounds good :+1:

PR #1912, opened 2021-04-11
[mediawiki] Update logo

davidbarratt:
The MediaWiki logo has been updated! :) 
https://www.mediawiki.org/wiki/Manual:MediaWiki_logo_guidelines


tianon:
Nice, looks good! :+1:

This logo gets scaled down to 120x120 for the logo that Docker displays in several places (such as search), in case you weren't aware -- probably worth considering whether you want the version with or without the text; our scripts can also support a `logo.svg` instead of `logo.png` if you wanted to embed the original artifact from those guidelines. :sweat_smile:

tianon:
I'll merge for now (to get that new pretty one up!) and let you think about whether there's a different size/style you want instead. :+1:

PR #1914, opened 2021-04-13
[adminer] Improve docs for ADMINER_PLUGINS interaction with configurable plugins

kerhong:
Currently the documentation suggests that for configurable plugins you need to both add the plugin to `ADMINER_PLUGINS` and add config to `/var/www/html/plugins-enabled/`. This is incorrect and causes errors. Instead users should only add config to `/var/www/html/plugins-enabled/`


tianon:
cc @TimWolla (`adminer` image maintainer)

TimWolla (approved):
This reads much better, thanks for this!

PR #1918, opened 2021-04-17
Reduce Arch Linux rootfs retention time to 2 months

hashworks:
https://gitlab.archlinux.org/archlinux/archlinux-docker/-/merge_requests/57


tianon:
Aha, I knew we had written down the current time somewhere but couldn't remember where 😅

LGTM

PR #1928, opened 2021-04-26
adding the "run" command to example

danieldggomes:
just adding the "run" command to the example that explains build and execution. It was noticed the lack of the docker run ...


yosifkit:
It seems that your editor changed a lot of unrelated blank space.

PR #1932, opened 2021-04-27
Smallest secured golang image

jbleduigou:
There is a great Medium post online about how to build a smaller and more secured docker image for a golang app.
This PR is a proposal to add some explanations and a link to the post itself.


tianon:
Thanks for the contribution! :+1:

This is an interesting article, but I'm having a hard time understanding what sets it apart from other similar articles such that we should give it "top billing" here in our documentation, so to speak.

On the content side, one of the _biggest_ concerns I have is https://github.com/golang/go/issues/19938 (as noted in the `golang:<version>-alpine` section of our documentation) -- Alpine is not officially supported by the Go project, so we definitely don't want to be suggesting it in the official documentation quite so strongly unless/until the Go project officially supports it.

jbleduigou:
Hi,

Thanks for the review!
I do agree with your analysis.

PR #1933, opened 2021-04-29
Update Aerospike  license.md

volmarl:


aerospike/license.md
```diff
@@ -1,9 +1,5 @@
 Copyright 2020-2021 Aerospike, Inc.
 
-Aerospike Community Edition is licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at [http://www.apache.org/licenses/LICENSE-2.0](http://www.apache.org/licenses/LICENSE-2.0).
-
```
yosifkit, line 4: :confused: :question: The community edition is still part of the current image tags. Does this just need to move down to be tightly coupled with the "AS IS" section?

volmarl: Let me remove the CE tags from the official image tags.

PR #1938, opened 2021-05-02
Arch Linux: Add note on lsign-key

hashworks:
Related Issue:
https://gitlab.archlinux.org/archlinux/archlinux-docker/-/issues/18

### DIFF
--- a/archlinux/content.md
+++ b/archlinux/content.md
@@ -21,6 +21,8 @@ This image is intended to serve the following goals:
 -	`pacman` needs to work out of the box
 -	All installed packages have to be kept unmodified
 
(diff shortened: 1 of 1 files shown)


hashworks:
@tianon Hm, markdownfmt output changes the appearence:
Before:
before

After:
after

Just so you know, might be something to fix in markdownfmt. I'll adjust the markdown for now.

tianon:
Can you elaborate on what you were trying to accomplish with the `>>>` before?

Looking at https://spec.commonmark.org/0.29/, the only thing that could possibly mean is a triply-nested blockquote, which would have to precede each line of text in the quote.

If I pull up https://github.com/docker-library/docs/commit/f77b59e64837353afa0dab8be8bcb289fd7f1eee and click on the "render" button, even GitHub doesn't seem sure what to do with it (and appears to ignore the `>>>` entirely):

image

Compared to:

image

If you were going for a horizontal line, perhaps you meant `---` instead?

(In our experience, the Docker Hub markdown rendering that this ends up in is usually even less forgiving than GitHub's :grimacing: :sweat_smile:)

SantiagoTorres:
No, I think this is just a copy-over from Gitlab, that uses those `>>>`'s  for multiline blockquotes... I think it's fixed now :)

PR #1947, opened 2021-05-11
Added -p 80:80 in Apache example

leenwim36:
It is already there in the "Apache without a Dockerfile" section,
added it in the "Apache with Dockerfile" too.
It won't work without the "-p 80:80"

### DIFF
--- a/php/variant-apache.md
+++ b/php/variant-apache.md
@@ -13,7 +13,7 @@ Where `src/` is the directory containing all your PHP code. Then, run the comman
 
 ```console
 $ docker build -t my-php-app .
-$ docker run -d --name my-running-app my-php-app
+$ docker run -d -p 80:80 --name my-running-app my-php-app
 ```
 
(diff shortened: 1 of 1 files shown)


yosifkit:
It works just fine without it and can still be accessed from the Linux host via container IP.  I don't think we need to publish a port on every `docker run` line; users should have a familiarity with Docker generally, container networking, and `docker` cli flags when trying to run and access their containers.

leenwim36:
Sorry, I disagree.
It *doesn't* work just fine without.
Without the "-p 80:80" you can't access it via "http://<containerIP>/" (and off course also not via "http://localhost/").
With the "-p 80:80" you can access it via both "http://<containerIP>/" and "http://localhost/" (and many more).
I agree that this is obvious for someone (like you and me) who is familiar with Docker, networking, etc., 
but I think that *if* we give a "docker build" and "docker run" example, that we should try to be 100% correct.
(my students, and all newbies, will be grateful :-) )

PR #1954, opened 2021-05-17
Update dart content

subfuzion:
Add safeguard for `pub get` when there is no `.dockerignore` and recommend use of a `.dockerignore` file.

### DIFF
--- a/dart/content.md
+++ b/dart/content.md
@@ -22,22 +22,37 @@ FROM dart:stable AS build
 
 # Resolve app dependencies.
 WORKDIR /app
-COPY pubspec.* .
+COPY pubspec.* ./
 RUN dart pub get
 
 # Copy app source code and AOT compile it.
 COPY . .
+# Ensure packages are still up-to-date if anything has changed
+RUN dart pub get --offline
 RUN dart compile exe bin/server.dart -o /server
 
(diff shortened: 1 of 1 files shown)


dart/content.md
```diff
@@ -22,22 +22,37 @@ FROM dart:stable AS build
...
+# Ensure packages are still up-to-date if anything has changed
+RUN dart pub get --offline
 RUN dart compile exe bin/server.dart -o /server
 
 # Build minimal serving image from AOT-compiled `/server` and required system
 # libraries and configuration files stored in `/runtime/` from the build stage.
 FROM scratch
 COPY --from=build /runtime/ /
-COPY --from=build /server /bin/
+COPY --from=build /app/bin/server /app/bin/
```
yosifkit, line 38: Doesn't the `-o /server` on line 32 mean that this was already correct?

I'd say it also makes sense to stay with `/bin` as the target since that is part of the default `path` of Docker images and so could make the `CMD` shorter.

subfuzion: Updated `-o /server` to `-o bin/server` in #1955. Using `/bin` would be okay, but using `/app` convention since other than this trivial example, there's a good chance there will be other static assets that need to get copied to the target directory.

PR #1962, opened 2021-05-21
Update to Neo4j logo

jennyowen:
We recently updated the company logo at Neo4j, so I'd like to update the Dockerhub documentation to reflect that. I also took the opportunity to delete some documentation for a super old, no longer supported, version.
Thanks!

PR #1964, opened 2021-05-25
Updated Neo4j logo

jennyowen:
We recently updated the company logo at Neo4j, so I'd like to update the Dockerhub documentation to reflect that. I also took the opportunity to delete some old documentation.

this is a copy of pull request https://github.com/docker-library/docs/pull/1962 but should be editable by the team at docker.

thank you!


jennyowen:
thanks @tianon ! :star:

tianon:
Thanks for your persistence and patience :smile:

> Does updating the logo in this repo update the logo at the top of this page also?

Yes, although there's usually a slight delay (for both) -- however, it appears both are updated now. :+1:

PR #1975, opened 2021-06-02
Update haskell to make the example dockerfiles more stable

AlistairB:
Closes the following issues:

- https://github.com/haskell/docker-haskell/issues/26
- https://github.com/haskell/docker-haskell/issues/27

cc @psftw


tianon:
Thanks for the :+1:, @psftw :smile:

PR #1986, opened 2021-07-07
Always persist caddy data directory/volume

rugk:
This defines the volume as external in order to avoid the problem that `docker-compose down` removes the volume.
This makes sure the volume is not accidentally removed.

This is important, because of the reasons the Readme describes further above. That volume must not be treated as temporary.

See https://docs.docker.com/compose/compose-file/compose-file-v3/#external
Ref https://forums.docker.com/t/why-docker-compose-down-deletes-my-volume-how-to-define-volume-as-external/67433?u=rugk

### DIFF
--- a/caddy/README.md
+++ b/caddy/README.md
@@ -201,8 +201,11 @@ services:
 
 volumes:
   caddy_data:
+    external: true
   caddy_config:
 ```
+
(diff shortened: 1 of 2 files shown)


tianon:
cc @hairyhenderson

tianon:
(Also, `README.md` is generated, per the comment at the top of it, so this change will need to be moved to the `stack.yml` + `content.md` files)

rugk:
Oh sorry, did not see that… now changed.

hairyhenderson (approved):
Thanks @rugk! LGTM

PR #1991, opened 2021-07-13
Refresh SonarQube Docs

colin-mueller-sonarsource:
Hello! SonarSourcer here proposing some changes (alongside the Product Team) to the Docs:

* **What is SonarQube?**
  * We have Boilerplate guidelines when describing SonarQube we could use, beyond the current 12 word description. It has been added.
* **How to use this image**
  * Using “Data Center” Edition rather than DataCenter Edition
* **Docker Host Requirements**
  * Due for an update based on [current requirements](https://docs.sonarqube.org/latest/requirements/requirements/#header-5)
* **Get Started in Two Minutes Guide**
  * We’ve moved towards wording like **Try Out SonarQube**
* **First Installation**
  * The "first" installation is usually when they've tried out SonarQube above? This section is where they get serious about installation (and then configuration), so it's been renamed.
* **Configuration > Database**
  * “Data Center” Edition rather than Datacenter Edition
* **Configuration > Use volumes**
  * Now that v7.9 LTS is EOL, can we remove reference to the /opt/sonarqube/conf persistent volume
  * “Data Center” Edition rather than DataCenter Edition
* **Advanced configuration > Customized Image**
  * Time to bump this example to sonarqube:8.9-community which means removing the /opt/sonarqube/conf/ example and replacing it with another (adding a custom plugin)
* **Advanced configuration > Avoid hard termination of SonarQube**
  * SonarQube v7.8 was a while ago (and EOL). We now avoid mentioning it and just stick to discussing the hard termination of SonarQube.

### DIFF
--- a/sonarqube/content.md
+++ b/sonarqube/content.md
@@ -1,12 +1,12 @@
 # What is SonarQube?
 
-[SonarQube](https://www.sonarqube.org/) is an open source product for continuous inspection of code quality.
+[SonarQube](https://www.sonarqube.org/) is the leading tool for continuously inspecting the Code Quality and Security of your codebases, and guiding development teams during Code Reviews. Covering 27 programming languages, while pairing-up with your existing software pipeline, SonarQube provides clear remediation guidance for developers to understand and fix issues, and for teams overall to deliver better and safer software. With over 225,000 deployments helping small development teams as well as global organizations, SonarQube provides the means for all teams and companies around the world to own and impact their Code Quality and Security.
 
 %%LOGO%%
 
 # How to use this image
 
-Here you'll find the Docker images for the Community Edition, Developer Edition, Enterprise Edition, and DataCenter Edition of SonarQube.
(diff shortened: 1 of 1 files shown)


sonarqube/content.md
```diff
@@ -15,42 +15,42 @@ Because SonarQube uses an embedded Elasticsearch, make sure that your Docker hos
...
 
 > **Warning:** You cannot use the same volumes on multiple instances of SonarQube.
 
-## First Installation
-
-For installation instructions, see Installing the Server from the Docker Image on the [Install the Server](https://docs.sonarqube.org/latest/setup/install-server/) page.
-
-To run a cluster with the DataCenter Edition, please refer to Installing SonarQube from the Docker Image on the [Install the Server as a Cluster](https://docs.sonarqube.org/latest/setup/install-cluster/) page.
 
 ## Upgrading
```
tianon, line 55:
```suggestion
## Upgrading
```

:eyes: (this should fix the CI failure)

PR #1992, opened 2021-07-13
Fix OpenJDK info about base OracleLinux image version

dolfinus:
OpenJDK images based on OracleLinux 8 instead of OracleLinux 7. This is important because OracleLinux 8 does is using not `yum` package manager but a `microdnf` instead, which could lead to misunderstanding.

### DIFF
--- a/openjdk/content.md
+++ b/openjdk/content.md
@@ -15,7 +15,7 @@ Java is a registered trademark of Oracle and/or its affiliates.
 The most straightforward way to use this image is to use a Java container as both the build and runtime environment. In your `Dockerfile`, writing something along the lines of the following will compile and run your project:
 
 ```dockerfile
-FROM %%IMAGE%%:7
+FROM %%IMAGE%%:11
 COPY . /usr/src/myapp
 WORKDIR /usr/src/myapp
 RUN javac Main.java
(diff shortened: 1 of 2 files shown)


openjdk/content.md
```diff
@@ -34,7 +34,7 @@ $ docker run -it --rm --name my-running-app my-java-app
 There may be occasions where it is not appropriate to run your app inside a container. To compile, but not run your app inside the Docker instance, you can write something like:
 
 ```console
-$ docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%IMAGE%%:7 javac Main.java
+$ docker run --rm -v "$PWD":/usr/src/myapp -w /usr/src/myapp %%IMAGE%%:12 javac Main.java
```
yosifkit, line 37: `12` is already EOL, so this should probably `11` until the next LTS, `17`, is released.

openjdk/variant-oracle.md
```diff
@@ -1,5 +1,5 @@
-## `%%IMAGE%%:<version>` (from 12 onwards), `%%IMAGE%%:<version>-oracle` and `%%IMAGE%%:<version>-oraclelinux7`
+## `%%IMAGE%%:<version>` (from 12 onwards), `%%IMAGE%%:<version>-oracle` and `%%IMAGE%%:<version>-oraclelinux8`
```
yosifkit, line 1: There are still `oraclelinux7` based images. :thinking: Perhaps we drop the mentioning of `oraclelinux7` and `oraclelinux8` in this heading and provide links to both 7 and 8 in the first paragraph below?

openjdk/content.md
dolfinus: Ok, fixed.

openjdk/variant-oracle.md
dolfinus: Ok, fixed.

tianon (approved):
Thanks!

PR #1994, opened 2021-07-15
Add MYSQL_ROOT_HOST env var description to readme.

likemusic:


yosifkit:
This appears to be a copy from [upstream MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/docker-mysql-more-topics.html#docker_var_mysql-root-host). But that documentation is for the `mysql/mysql-server` images and not necessarily correct for the `mysql` images maintained here in the official-images.  This is one part that they differ since the default `MYSQL_ROOT_HOST` for `mysql/mysql-serrver` is to [use `localhost`](https://github.com/mysql/mysql-docker/blob/main/mysql-server/8.0/docker-entrypoint.sh#L121-L123), but the default for the images documented here is `%`


https://github.com/docker-library/mysql/blob/d60655b1b42f677c315d5fe2ca0e87126ca921f5/template/docker-entrypoint.sh#L217: 
```bash
	file_env 'MYSQL_ROOT_HOST' '%'
```

mysql/README.md
```diff
@@ -163,6 +163,9 @@ See also https://dev.mysql.com/doc/refman/5.7/en/environment-variables.html for
 
 This variable is mandatory and specifies the password that will be set for the MySQL `root` superuser account. In the above example, it was set to `my-secret-pw`.
 
+### `MYSQL_ROOT_HOST`
+This variable is optional. It allows you to use root connections from outside the container (for example from your host or remote machine). To allow root connections from your host machine set this value to `host.docker.internal`. The variable accepts only one entry, but wildcards are allowed (for example, `MYSQL_ROOT_HOST=172.*.*.*` or `MYSQL_ROOT_HOST=%`). If the variable is set a second root account will be created for the value you used, so you'll have `'root'@'localhost'` and `'root'@'${MYSQL_ROOT_HOST}'`.
```
yosifkit, line 167: `host.docker.internal` will not work in the default configuration since we [set `skip-name-resolve`](https://github.com/docker-library/mysql/blob/d60655b1b42f677c315d5fe2ca0e87126ca921f5/8.0/config/conf.d/docker.cnf#L2):

>```console
>[Warning] [MY-010315] [Server] 'user' entry 'root@host.docker.internal' ignored in --skip-name-resolve mode.
>```

PR #1995, opened 2021-07-15
Add MYSQL_LOG_CONSOLE env var description.

likemusic:


yosifkit:
This appears to be a copy from [upstream MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/docker-mysql-more-topics.html#docker_var_mysql-log-console). This variable does not appear in the [`mysql/mysql-server` source](https://github.com/mysql/mysql-docker), so it must be a variable like `MYSQL_HOST` that is handled by `mysqld` itself (or scripts in the deb/yum packaging).

Since this is not a variable specific to our Dockerization of MySQL in the offical-images (https://github.com/docker-library/mysql) or to the upstream `mysql/mysql-server`, so it should remain only in upstream documentation.

PR #1996, opened 2021-07-16
Remove rabbitmq environment variable usage since it is removed in 3.9

yosifkit:
Environment variables are removed with the addition of 3.9: https://github.com/docker-library/rabbitmq/pull/467. Users must provide an ini-style `.conf` file (or an erlang-based `.config`) instead of the environment variables that they were using in `3.8`.  A full `.conf` file is also valid to use with `3.8` and earlier (though old enough versions only support a `.config` file in erlang syntax).

Deprecation notice added to the 3.8 entrypoint script: https://github.com/docker-library/rabbitmq/pull/424 which will also print out the config file it generated. Users can also `docker cp` their current config from their running RabbitMQ container.

Reading https://www.rabbitmq.com/versions.html, I am not certain when the "end of general support" for `3.8.x` will be, so users should likely update their usage of `rabbitmq` as soon as possible

This may need changed depending on the outcome of https://github.com/docker-library/rabbitmq/pull/501

Since we only have a `3.9-rc`, should this wait for the 3.9.0 final release?

### DIFF
--- a/rabbitmq/content.md
+++ b/rabbitmq/content.md
@@ -29,50 +29,46 @@ This will start a RabbitMQ container listening on the default port of 5672. If y
 
 Note the `database dir` there, especially that it has my "Node Name" appended to the end for the file storage. This image makes all of `/var/lib/rabbitmq` a volume by default.
 
+### Environment Variables
+
+For a list of environment variables supported by RabbitMQ itself, see the [Environment Variables section of rabbitmq.com/configure](https://www.rabbitmq.com/configure.html#supported-environment-variables)
+
+**WARNING:** As of RabbitMQ 3.9, all of the docker-specific variables listed below are deprecated and no longer used. Please use a configuration file instead; visit [rabbitmq.com/configure](https://www.rabbitmq.com/configure.html) to learn more about the configuration file. For a starting point, the 3.8 images will print out the config file it generated from supplied environment variables.
+
+```bash
+# Unavailable in 3.9 and up
+RABBITMQ_DEFAULT_PASS
+RABBITMQ_DEFAULT_PASS_FILE
+RABBITMQ_DEFAULT_USER
+RABBITMQ_DEFAULT_USER_FILE
+RABBITMQ_DEFAULT_VHOST
+RABBITMQ_ERLANG_COOKIE
+RABBITMQ_MANAGEMENT_SSL_CACERTFILE
+RABBITMQ_MANAGEMENT_SSL_CERTFILE
+RABBITMQ_MANAGEMENT_SSL_DEPTH
+RABBITMQ_MANAGEMENT_SSL_FAIL_IF_NO_PEER_CERT
+RABBITMQ_MANAGEMENT_SSL_KEYFILE
+RABBITMQ_MANAGEMENT_SSL_VERIFY
+RABBITMQ_SSL_CACERTFILE
+RABBITMQ_SSL_CERTFILE
+RABBITMQ_SSL_DEPTH
+RABBITMQ_SSL_FAIL_IF_NO_PEER_CERT
+RABBITMQ_SSL_KEYFILE
+RABBITMQ_SSL_VERIFY
+RABBITMQ_VM_MEMORY_HIGH_WATERMARK
+```
+
 ### Memory Limits
 
-RabbitMQ contains functionality which explicitly tracks and manages memory usage, and thus needs to be made aware of cgroup-imposed limits.
+RabbitMQ contains functionality which explicitly tracks and manages memory usage, and thus needs to be made aware of cgroup-imposed limits (e.g. [`docker run --memory=..`](https://docs.docker.com/config/containers/resource_constraints/#limit-a-containers-access-to-memory)).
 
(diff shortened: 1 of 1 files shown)


tianon:
@michaelklishin @gerhard this PR is the image docs update that goes with the 3.8 / 3.9 deprecations -- it would be really useful if we could convince y'all to give it some review :innocent: :pray:

tianon:
Regarding whether to update the docs now or wait until 3.9 is GA, my opinion is that given 3.8 is already warning about these things being deprecated, it would be useful to have the current documentation already match that (especially since the conversion to a configuration file can be done in 3.8 and then for users that have pre-converted, 3.9 will not be nearly as disruptive).

gerhard:
> @yosifkit Reading https://www.rabbitmq.com/versions.html, I am not certain when the "end of general support" for 3.8.x will be, so users should likely update their usage of rabbitmq as soon as possible

In the past we used to give all users 6 months to migrate to the new release series. I have just updated the versions page to capture this for 3.9.0.

Reviewing these changes now. Thanks for the heads up @tianon !

rabbitmq/content.md
```diff
@@ -29,50 +29,46 @@ This will start a RabbitMQ container listening on the default port of 5672. If y
...
-
-### Erlang Cookie
+### Environment Variables
 
-See the [RabbitMQ "Clustering Guide"](https://www.rabbitmq.com/clustering.html#erlang-cookie) for more information about cookies and why they're necessary.
+For a list of environment variables supported by RabbitMQ itself, see the [Environment Variables section of rabbitmq.com/configure](https://www.rabbitmq.com/configure.html#supported-environment-variables)
 
-For setting a consistent cookie (especially useful for clustering but also for remote/cross-container administration via `rabbitmqctl`), use `RABBITMQ_ERLANG_COOKIE`:
+**WARNING:** As of RabbitMQ 3.9, all of the docker-specifc variables listed below are deprecated and no longer used. Please use a configuration file instead; visit [rabbitmq.com/configure](https://www.rabbitmq.com/configure.html) to learn more about the configuration file. For a starting point, the 3.8 images will print out the config file it generated from supplied environment variables.
```
gerhard, line 36:
```diff
- docker-specifc
+ docker-specific
```

gerhard: I think that we should encourage users to use configuration fragments, as mentioned here https://github.com/docker-library/rabbitmq/pull/501#issuecomment-882506021.

Also, do we mean **deprecated** or **removed**? I think it's the latter.

rabbitmq/content.md
```diff
@@ -29,50 +29,46 @@ This will start a RabbitMQ container listening on the default port of 5672. If y
...
+RABBITMQ_SSL_CERTFILE
+RABBITMQ_SSL_DEPTH
+RABBITMQ_SSL_FAIL_IF_NO_PEER_CERT
+RABBITMQ_SSL_KEYFILE
+RABBITMQ_SSL_VERIFY
+RABBITMQ_VM_MEMORY_HIGH_WATERMARK
 ```
 
-This can then be used from a separate instance to connect:
+### Memory Limits
```
gerhard, line 61: This reminded me about adding support in RabbitMQ for cgroup memory limits. This is the last conversation that I remember (almost 1 year old now): https://github.com/docker-library/rabbitmq/pull/424#issuecomment-663491388

Is this still relevant @tianon?

gerhard (changes requested):
There are a few minor changes that I think we should do before merging this.

In a new PR, would it be worth adding a section about the first-class & free support for managing RabbitMQ via [cluster-operator](https://github.com/rabbitmq/cluster-operator)? I know that this is K8s-specific, but there is a lot of RabbitMQ operations know-how, especially in the context of a cluster, that this utility handles out of the box. If nothing else, it can be used as a reference for all the things that users can (and might want to) implement in their container orchestration tool of choice.

rabbitmq/content.md
tianon: Yes, this is definitely still relevant.  I summarized it much shorter in https://github.com/docker-library/rabbitmq/pull/424#issuecomment-752272295

rabbitmq/content.md
tianon: I think we're splitting hairs a little bit on the distinction between those two -- we usually use the word "deprecated" for things like that because that's pretty clearly intended to mean "don't use this anymore and it might break in the future" (in this case, the "deprecation period" was just really short and even happened retroactively within the 3.8 series).

tianon:
> In a new PR, would it be worth adding a section about the first-class & free support for managing RabbitMQ via [cluster-operator](https://github.com/rabbitmq/cluster-operator)? I know that this is K8s-specific, but there is a lot of RabbitMQ operations know-how, especially in the context of a cluster, that this operator handles out of the box. If nothing else, it can be used as a reference for all the things that users can (and might want to) implement in their container orchestration tool of choice.

Yeah, I think that's fine -- want to work on something with the wording/positioning you'd like to see?

tianon (approved):

tianon:
(Getting this merged since the 3.9 changes are up now and we want the docs to match -- easy enough to update via more PRs. :+1:)

PR #1997, opened 2021-07-22
Added an example of starting pgAdmin4.

mariugul:
There was no description of how to set up pgAdmin4, which is perhaps the most used postgres database tool. I added this file to make it easier for people to get setup with this tool if they do not wish to use adminer.


tianon:
The example with `adminer` is intended to be a basic example of PostgreSQL usage -- users are not expected to use it as-is.  Adminer was chosen specifically because it has a maintained image within the Official Images program (unlike pgAdmin).

PR #2000, opened 2021-07-27
Create deprecation notice for fsharp

stevedesmond-ca:
Per https://github.com/docker-library/official-images/pull/10612#issuecomment-886992900 by @tianon, this PR adds a deprecation notice to the fsharp image's documentation, pointing people to the newer, more maintained images.


fsharp/deprecated.md
```diff
@@ -0,0 +1 @@
+These images were removed in favor of the [official .NET images](https://hub.docker.com/_/microsoft-dotnet) maintained by Microsoft.
```
tianon, line 1: Is there a specific image over there that current https://hub.docker.com/_/fsharp users should transition to?
(Having trouble understanding the mapping myself, and I imagine existing users might too, but perhaps not.)

stevedesmond-ca: Indeed, I feel like this type of clarification is something Microsoft and the .NET Foundation are always working on, what with .NET trying to be everything for everyone. Let me know if the amended commit helps!

tianon:
FYI/cc @ninjarobot -- this seems fine to me :+1:

PR #2004, opened 2021-07-30
Adds section for subdirectory set-up to README.md

fhackenberger:
So far it was unclear how a set-up with docker-wordpress would work reliably when hosting wordpress in the subdirectory of a domain. This change adds a working solution.


fhackenberger:
Sorry, I had to re-do the pull request, I was unaware that editing the README.md directly is not possible.

PR #2008, opened 2021-08-04
add eclipse-temurin and deprecate adoptopenjdk

gdams:
see https://github.com/docker-library/official-images/pull/10662 for context


adoptopenjdk/deprecated.md
```diff
@@ -0,0 +1 @@
+This image is officially deprecated in favor of [the `temurin-compliance` image](https://hub.docker.com/_/temurin-compliance/), and will receive no further updates after 2021-08-01 (Aug 01, 2021). Please adjust your usage accordingly.
```
tianon, line 1: `temurin-compliance` ? :eyes: :thinking:

tianon: It also seems kind of strange to actually deprecate this older image before the new one is actually a full, complete replacement?

gdams: fixed the name, thanks!

> It also seems kind of strange to actually deprecate this older image before the new one is actually a full, complete replacement?

I think deprecating now is the right thing to do, this repo will no longer be updated, the OpenJDK versions in that repo are now out of date and not recommended in production use. We will be adding the other architectures in the coming weeks so there won't be a big overlap

tianon (approved):

tianon:
FYI affected image maintainers:

- `cassandra`: @tianon @yosifkit
- `gradle`: @keeganwitt
- `groovy`: @keeganwitt
- ~~`maven`: @carlossg~~ (https://github.com/docker-library/official-images/pull/10728)
- `open-liberty`: @crpotter @wraschke @arthurdm
- `tomcat`: @tianon @yosifkit
- `websphere-liberty`: @crpotter @wraschke @arthurdm

(This also helped catch https://github.com/docker-library/bashbrew/pull/32 :sweat_smile:)

PR #2014, opened 2021-08-18
add eclipse-temurin logo

gdams:
Please can somebody also set the actual image logo as this one below: CC @tianon 

Logo_Temurin_2021_08_17_JRR_RGB-V1B


tianon:
I'm a little confused here -- your PR description includes a vertical/square image, but the PR itself includes a wide image. :confused:

I think there's a delay on Docker's backend (I'm honestly not sure why it's not updated yet -- it really should be), but this same logo is also used in the following place:

image

gdams:
@tianon I didn't realise that they used the same logo. If that's the case then let's stick with the wide image

PR #2017, opened 2021-08-25
Updating the Semeru Runtimes docker documentation.

narkedi:
IBM Semeru Runtimes is responsible for delivering the official Docker Images with OpenJDK + Eclipse OpenJ9


ibm-semeru-runtimes/content.md
```diff
@@ -0,0 +1,37 @@
+## Overview
+
+The files in this repository are used to build the images that contain IBM Semeru Runtime Open Edition OpenJDK binaries.
```
yosifkit, line 3: While the generated `README` will *link* to the Dockerfiles used for building, this statement is incorrect since the files here are just the documentation that will appear on the Docker Hub.

PR #2037, opened 2021-09-23
Update fedora doc and logo

cverna:


cverna:
@yosifkit would you know how we actually change the logo next to the image name here https://hub.docker.com/_/fedora ?

yosifkit:
It should be changed when the docs are pushed by our jenkins jobs (https://doi-janky.infosiftr.net/job/docs/), but Docker Hub has caching or something so it takes a while to finally show up.

PR #2042, opened 2021-09-28
update: use the new README

noraj:
- use the new readme from https://github.com/traefik/traefik-library-image/blob/master/README.md instead of the old one from the archived repo https://github.com/containous/traefik-library-image/blob/master/README.md
- fixes https://github.com/traefik/traefik-library-image/issues/13
- completes https://github.com/docker-library/docs/pull/2040

### DIFF
--- a/traefik/content.md
+++ b/traefik/content.md
@@ -1,110 +1,52 @@
 %%LOGO%%
 
-[Traefik](https://traefik.io) is a modern HTTP reverse proxy and load balancer that makes deploying microservices easy.
+[Træfɪk](https://github.com/traefik/traefik) is a modern HTTP reverse proxy and load balancer that makes deploying microservices easy.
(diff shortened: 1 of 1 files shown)


ldez:
Sorry, but we don't agree with these changes.

ldez:
Please close this PR: my PR #2043 do the expected changes.

noraj:
> Please close this PR: my PR #2043 do the expected changes.

🤔 

1. Why having Traefik v2 + Traefik v1 examples ?
2. Please change `traefik:v2.0` tags to `traefik:v2.5` tags
3. For the documentation rather than giving a 1.7 + 2.5 link, why not just giving https://doc.traefik.io/traefik/ without a version then people can choose the version they want 
image
Else it we will need to update the documentation link at each minor version of each branch.

ldez:
1. because we want to have both
2. fixed in my PR
3. fixed in my PR
4. fixed in my PR

PR #2044, opened 2021-09-29
Add more Ubuntu details

tianon:
@mwhudson @toabctl this is a "straw man" to try to put more of this down in text form because _I_ can never seem to find it when I need it, so IMO it's useful data, but I'm very open to suggestions, alternative wordings, disgust, etc. :smile:

### DIFF
--- a/ubuntu/content.md
+++ b/ubuntu/content.md
@@ -10,7 +10,7 @@ Development of Ubuntu is led by Canonical Ltd. Canonical generates revenue throu
 
 # What's in this image?
 
-This image is built from official rootfs tarballs provided by Canonical (specifically, https://partner-images.canonical.com/core/).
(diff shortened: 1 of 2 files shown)


mwhudson:
LGTM

ubuntu/content.md
```diff
@@ -10,7 +10,7 @@ Development of Ubuntu is led by Canonical Ltd. Canonical generates revenue throu
 
 # What's in this image?
 
-This image is built from official rootfs tarballs provided by Canonical (specifically, https://partner-images.canonical.com/core/).
+This image is built from official rootfs tarballs provided by Canonical (specifically, https://partner-images.canonical.com/oci/ and https://partner-images.canonical.com/core/).
```
toabctl, line 13: `/oci` for >= bionic and `/core` for < bionic could be mentioned, but this is already an improvement. Thanks!

toabctl (approved):

ubuntu/content.md
tianon: Sure, updated :smile:

PR #2046, opened 2021-09-29
redis: prefer "--save" over "--appendonly"

newhouse:
The `--appendonly` options has some cons that, IMO, lend the `--save` option to be a more appropriate suggestion in these docs.

I recently found out that in a local `docker-compose`, the `appendonly.aof` file had ballooned to over 1.3GB in my `redis` service, thanks to following the default suggestion here.

I think steering people towards `--save` in this simple example is probably better.

### DIFF
--- a/redis/content.md
+++ b/redis/content.md
@@ -25,10 +25,10 @@ $ docker run --name some-redis -d %%IMAGE%%
 ## start with persistent storage
 
 ```console
-$ docker run --name some-redis -d %%IMAGE%% redis-server --appendonly yes
+$ docker run --name some-redis -d %%IMAGE%% redis-server --save 60 1 --loglevel warning
 ```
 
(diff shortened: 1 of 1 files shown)


tianon:
@itamarhaber would you be willing to share your thoughts or pull in someone else who might be able to?  This makes sense to me, but I'm definitely not an authority. :innocent: :pray:

itamarhaber:
Thanks for pulling me in @tianon.

When you spin up a Redis server (also one from the library's image) it defaults to no AOF and periodic snapshotting:

```
127.0.0.1:6379> config get save
1) "save"
2) "3600 1 300 100 60 10000"
127.0.0.1:6379> config get appendonly
1) "appendonly"
2) "no"
```

Defaults are just that, and as such need to be examined carefully before serious use. This applies to all the defaults in Redis, which are primarily designed to "work out of the box". Also note, that:

```
docker run --name some-redis -d %%IMAGE%% redis-server --appendonly yes
```

does **not** disable the default snapshotting policy, so basically it means "enable AOF and rely on the default snapshot policy".

That said, I agree with @newhouse's suggestion. It would be good to refer the reader to the [Redis documentation page about persistence](https://redis.io/topics/persistence) for reference and self-help.

itamarhaber (approved):

PR #2050, opened 2021-10-04
Update README.md

gmpassos:
Add extra information for authentication and user issues that are common for initial users. Without that the docker image can be useless in some development environments.

### DIFF
--- a/mysql/README.md
+++ b/mysql/README.md
@@ -141,10 +141,10 @@ This will start a new container `some-mysql` where the MySQL instance uses the c
 
 ### Configuration without a `cnf` file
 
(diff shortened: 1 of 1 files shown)


tianon:
This definitely has some overlap with https://github.com/docker-library/docs/pull/1203 (even if it's not completely a duplicate).

I'm very hesitant to change/recommend a different value for this setting, because I don't think MySQL upstream would've changed it "willy nilly" (especially given how disruptive it is) -- they likely have very good reasons for changing it, so I'm not comfortable actively recommending our users change it back.

tianon:
Sorry for the delay -- unfortunately, we're not comfortable making this change.  Thank you for the contribution, though!

PR #2052, opened 2021-10-07
Update examples to supported release

dbready:
The referenced Perl tag in the examples is ancient and EOL. Update to a far more modern release.

Not a Perl guy, but needed to run a quick script. Mindlessly copied the tag from the examples which failed due to its age.


tianon:
Nice, thanks!

FYI @PeterMartini @zakame

PR #2062, opened 2021-10-25
Clarify Dart getting started steps

mit-mit:
Based on feedback in https://github.com/dart-lang/dart-docker/issues/28

### DIFF
--- a/dart/content.md
+++ b/dart/content.md
@@ -18,7 +18,7 @@ After installing the Dart SDK, use the `dart` command to create a new server app
 $ dart create -t server-shelf myserver
 ```
 
-### Running server with Docker Desktop
+### Running the server with Docker Desktop
 
 If you have [Docker Desktop](https://www.docker.com/get-started) installed, you can build and run on your machine with the `docker` command:
 


yilmazdurmaz:
@athomas @mit-mit 

thanks for this addition.

a little problem though: the command seems pretty new. I use dart 2.13.3 which is an installation no older than 6 months, in it the command template is "server-simple" not "server-shelf". 

can you include dart version info too? a minimum maybe? or maybe give both commands if both are applicable in docker.

mit-mit:
Would this help?
https://github.com/docker-library/docs/pull/2087/files

yilmazdurmaz:
@mit-mit  about dart versions, you have more knowledge than I may possible cant have for another year :)

if `server-shelf` came out when 2.14 did (so specific minor version) then that edit is fine.

as for `server-simple`. I saw this `stagehand` replaced by `dart create` somewhere in near history. I don't know of those dates. I just happened to have version 2.13 and there the template name is different and it does not have `dockerfile` in it.

and in that case, this docker document has a dockerfile content documented down below that same page.

so maybe adding something like this helps:

"for older versions of dart, create your dart server project, add a `dockerfile` and refer to the dockerfile image documentation down below (eg. template name is `server-simple` in version 2.13 )"

mit-mit:
> if server-shelf came out when 2.14 did (so specific minor version) then that edit is fine.

It did.

> As for server-simple. I saw this stagehand replaced by dart create somewhere in near history.

Yes, that also happened in 2.14: https://github.com/dart-lang/sdk/issues/46100

> "for older versions of dart"

We don't generally document past stable releases as we're still evolving quite a bit, so this quickly get's unsustainable.


Thanks for the feedback! I'll get that new PR landed.

PR #2067, opened 2021-10-28
Remove large Vault logo from readme

mochanghc:
Removed a full width Vault logo that was making the page difficult to read

### DIFF
--- a/vault/README.md
+++ b/vault/README.md
@@ -48,14 +48,11 @@ WARNING:
 -	**Source of this description**:  
 	[docs repo's `vault/` directory](https://github.com/docker-library/docs/tree/master/vault) ([history](https://github.com/docker-library/docs/commits/master/vault))
 
-# Vault
-
+# Vault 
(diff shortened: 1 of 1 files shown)


mochanghc:
Hi I like the idea of adding the width and height. What would be the best
way to do that?

tianon:
I think the easiest way would be updating `logo.svg` to change:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 250 250">
```

To something like:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 250 250" width="250" height="250">
```

... but I'd like to hear from the `vault` image maintainers for their preferences first. :sweat_smile:

re-cc @chrishoffman @jefferai @meirish :pray:

jefferai:
@tianon None of the @'d people are maintainers on this any longer -- @kalafut @mladlow @briankassouf

tianon:
> @tianon None of the @'d people are maintainers on this any longer -- @kalafut @mladlow @briankassouf

Doh, thanks @jefferai :facepalm:

For @kalafut @mladlow @briankassouf: it'd be great to update https://github.com/docker-library/official-images/blob/2d3e617baa553d021932a122762af7731e836d99/library/vault#L1-L3 when you get a chance :sweat_smile: :grimacing: :heart:

kalafut:
Thanks for flagging this. We'll take a look.

PR #2068, opened 2021-10-29
Nextcloud: Update to work with MariaDB >= 10.6

agowa338:
Nextcloud is incompatible with the new innodb-read-only-compressed feature introduced with MariaDB 10.6

See:
* https://mariadb.com/docs/reference/mdb/cli/mariadbd/innodb-read-only-compressed/
* https://techoverflow.net/2021/08/17/how-to-fix-nextcloud-4047-innodb-refuses-to-write-tables-with-row_formatcompressed-or-key_block_size/
* https://help.nextcloud.com/t/update-to-next-cloud-21-0-2-has-get-an-error/117028

### DIFF
--- a/nextcloud/content.md
+++ b/nextcloud/content.md
@@ -230,7 +230,7 @@ services:
   db:
     image: mariadb
     restart: always
-    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW
+    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW --skip-innodb-read-only-compressed
     volumes:
       - db:/var/lib/mysql
     environment:
@@ -276,7 +276,7 @@ services:
   db:
     image: mariadb
     restart: always
-    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW
+    command: --transaction-isolation=READ-COMMITTED --binlog-format=ROW --skip-innodb-read-only-compressed
     volumes:
       - db:/var/lib/mysql
     environment:


grooverdan:
Maybe tag the mariadb image with 10.5 (per https://github.com/nextcloud/server/issues/25436#issuecomment-936588457) to give nextcloud devs more time to implement their migration path away from ROW_FORMAT=compressed (which has started https://github.com/nextcloud/server/pull/29051).

agowa338:
But wouldn't the current patch do the same by just disabling the conflicting feature?

grooverdan:
Given the migration path away from ROW_FORMAT=compressed may be non-trivial, capping nextcloud users to 10.5 until the exact migration path of [MDEV-22367](https://jira.mariadb.org/browse/MDEV-22367) has been decided.

Given some users are going to choose `:latest` anyway, and passing a `mariadb:10.5` as the image along with `--skip-innodb-read-only-compressed` will fail, how about both, set `mariadb:10.5` as the image name, and pass `--loose-skip-innodb-read-only-compressed` such that setting won't cause and error on the 10.5 image?

tianon:
(I'm also going to cc the `nextcloud` folks for their thoughts/visibility/review: @tilosp @J0WI)

J0WI:
Nextcloud [recommends 10.5](https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html) and the recommendation for Docker should be the same: https://github.com/nextcloud/docker/commit/1a432d98fa98563182a56bdeab2b00d446a33e75 https://github.com/nextcloud/docker/commit/c91097a08e4d9fcb7ce5058e121760dc17909ae5

Unfortunately the docs here are not synced automatically the the readme in the repo.

grooverdan:
no longer needed since https://jira.mariadb.org/browse/MDEV-27736

tianon:
Thanks!  (and thanks for the contribution anyhow :heart:)

PR #2070, opened 2021-10-31
Add info on the -perl tag variants to the nginx README

prasadkatti:


prasadkatti:
The `no-readme` ci check is failing. I updated a readme file, but looks like those are autogenerated and I need to update content.md instead. The issue is I can't find the updated section about image variants in the content.md file.

yosifkit:
Thanks for figuring out `variant-perl.md` :+1:

Seems fine to me, but I'll let the `nginx` image maintainer decide on content.  CC @thresheek
(this will show up in the ["Image Variants" section](https://github.com/docker-library/docs/blob/4b4653d9ec7ad9c38314cd57a86ee8224ef7baac/nginx/README.md#image-variants)).

thresheek:
Looks good to me!  Thank you very much!

PR #2076, opened 2021-11-07
Fixed typo in doc

sowdowdow:


yosifkit:
Thanks for the contribution; this change will need to be in the respective `content.md` since each image `README.md` is generated (the failing build test). You'll need to replace the current commit, since the build tests will still fail. If you'd like, I can move the change and force push the commit on your branch.

PR #2078, opened 2021-11-12
mention how to change internal service port

gcbw:
This is very useful when running in host mode network.

Is often requested and this is the "official documentation" so far https://github.com/MariaDB/mariadb-docker/issues/334

### DIFF
--- a/mariadb/README.md
+++ b/mariadb/README.md
@@ -212,7 +212,7 @@ By default, the entrypoint script automatically loads the timezone data needed f
 
 ### `MYSQL_TCP_PORT`
 
-Changes the TCP port used inside the container. Defaults to 3306. Setting name doesn't follow the pattern as it is used directly by MariaDB.
+Changes the TCP port used inside the container. Defaults to 3306. Variable name doesn't follow the pattern as it is used directly by MariaDB.
 
 Using the default (i.e. not defining this variable) and exposing the desired external port via `ports` composer setting is recommended instead. Use this setting only for `network_mode: host` cases. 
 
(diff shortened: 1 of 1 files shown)


tianon:
As the CI's eventually going to point out, this change will need to move from `README.md` into `content.md`.

I'm also going to CC @grooverdan and @dbart (the MariaDB image maintainers).

grooverdan:
While this env obviously does work, it does look a bit out of place compared to other env variables. It also doesn't really have a 3306 default

Any reason we just couldn't recommend `--port 3909` or similar in the `Connect to MariaDB` section which is far too docker generic at the moment? It would prevent reintroducing a MYSQL only naming and hint towards other MariaDB system server command line options that could be used by in the same way.

gcbw:
This was not intended as an improvement (which i agree with @grooverdan would be to add the explicit default and use --port receiving a proper docker composer variable). This was instead intended as simply documenting the currently accepted work around when using host-only network driver.

yosifkit:
Yeah, it might make sense to expand the "configuration without a cnf file section" to mention `--port` in the same way you can customize all the other MariaDB variables, via flags to `mariadbd` (https://github.com/docker-library/docs/tree/ee77758ebb4ab10e7165c533d887649b55510eb8/mariadb#configuration-without-a-cnf-file).

And the `yaml` equivalent is `command`:

```yaml
version: '3.1'

services:

  db:
    image: mariadb
    environment:
      MYSQL_ROOT_PASSWORD: example
    command: --port 3309
```

PR #2084, opened 2021-11-19
[nats-streaming] Release v0.23.2

kozlovic:
Details can be found [here](https://github.com/nats-io/nats-streaming-server/releases/tag/v0.23.2)

### DIFF
--- a/nats-streaming/README-short.txt
+++ b/nats-streaming/README-short.txt
@@ -1,1 +1,1 @@
-NATS Streaming is an open-source, high-performance, cloud native messaging streaming system.
+DEPRECATED; An open-source, high-performance, cloud native messaging streaming system.
--- a/nats-streaming/content.md
+++ b/nats-streaming/content.md
@@ -1,9 +1,5 @@
 # [NATS Streaming](https://nats.io): A high-performance cloud native messaging streaming system.
 
-# **WARNING:** Deprecation Notice
-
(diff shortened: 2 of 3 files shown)


tianon:
I would suggest two things:

- if you put this in a `deprecated.md`, it'll show up at the very very top of the description (increasing visibility): https://github.com/docker-library/docs/blob/8221c8012d002395cc4ac527643ecfd181a9ee44/update.sh#L107-L112
- I would also recommend adding `DEPRECATED;` to the beginning of `readme-short.txt` so that the status is clear to users directly in `docker search` and the Hub search interface/views

kozlovic:
@tianon Hum... trying to add `DEPRECATED;` as you suggested, but get a failure.. I tried first on the same line, got a failure, so made a new line, still failure..

nats-streaming/README-short.txt
```diff
@@ -1 +1,2 @@
+DEPRECATED;
 NATS Streaming is an open-source, high-performance, cloud native messaging streaming system.
```
tianon, line 2: Sorry, Docker Hub has a limit of 100 characters; how about this?

```suggestion
DEPRECATED; NATS Streaming is a high-performance, cloud native messaging streaming system.
```

kozlovic: Sorry, just saw that in your README, so removed "NATS Streaming" from the sentence, since it is redundant because it is already in the NATS Streaming page. Should be good now.

PR #2091, opened 2021-12-02
Fix typo in the varnish docs

jdreesen:
I assume this is a typo.


tianon:
Nice catch! :joy: (fyi @gquintard)

Thank you!

PR #2098, opened 2021-12-28
[README] Add comment about database host name

akashin:
This will make it easier to set up the mediawiki for users not intimately familiar with Docker Stacks conventions for service host naming. I was lucky to find the answer to this problem on https://stackoverflow.com/questions/57312109/mediawiki-docker-official-image-connection-refused-by-mysql, but having it in the docs would make the setup much more smooth.

### DIFF
--- a/mediawiki/README.md
+++ b/mediawiki/README.md
@@ -125,6 +125,7 @@ services:
       # this yaml and uncomment the following line and use compose to restart
       # the mediawiki service
       # - ./LocalSettings.php:/var/www/html/LocalSettings.php
+  # This key also defines the name of the database host used during setup instead of the default "localhost"
   database:
     image: mariadb
     restart: always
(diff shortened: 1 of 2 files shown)


tianon:
cc `mediawiki` image maintainers: @davidbarratt, @legoktm, @addshore

(for this change to be acceptable, it'll need to move to the non-generated file in the same directory since `README.md` is a fully generated artifact, but if the image maintainers are on board with it and you don't get to moving it, I can assist with that after this week, since I'm currently off :see_no_evil:)

davidbarratt:
@akashin Feel free to make this change to https://github.com/akashin/docs/blob/patch-1/mediawiki/content.md but I'd prefer to not abandon the automattion.

akashin:
Ah, did realize that this file was auto-generated, thanks for pointing out! I've moved the changes to the `stack.yml`.
Don't mind the additional commits, I will squash them into a single one before submitting.

tianon:
Looks fine to me -- want to squash the commits or would you rather I squash-and-merge? :sweat_smile:

PR #2101, opened 2022-01-06
Note that Haskell ARM64 support does not include the stack tool

AlistairB:
As it does not [currently support ARM64](https://github.com/commercialhaskell/stack/issues/2103).

And clarify old dockerfiles are no longer included in the repo.

Should be merged after https://github.com/docker-library/official-images/pull/11620


AlistairB:
This is ready to go.

PR #2105, opened 2022-01-15
update monica branch name

asbiin:


monica/license.md
```diff
@@ -1 +1 @@
-View [license information](https://github.com/monicahq/monica/blob/master/LICENSE) for the software contained in this image.
+View [license information](https://github.com/monicahq/monica/blob/main/LICENSE) for the software contained in this image.
```
tianon, line 1: Looks like this one also needs a `.md` on it :sweat_smile: 

```suggestion
View [license information](https://github.com/monicahq/monica/blob/main/LICENSE.md) for the software contained in this image.
```

asbiin: Good catch! Thanks

PR #2111, opened 2022-02-05
Update README.md adding docker-compose template for InfluxDB

RobsOnWaves:
Adding docker-compose template

### DIFF
--- a/influxdb/README.md
+++ b/influxdb/README.md
@@ -422,6 +422,39 @@ $ docker run -d -p 8086:8086 \
       influxdb:2.0
 ```
 
+#### Using docker-compose approach
+To add you automated setted up InfluxDB database to your docker-compose file, you can add the folling lines:
+
+``` code
+version: "3.9"
+
+services:
+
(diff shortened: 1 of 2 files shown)


tianon:
I would suggest putting this in a `stack.yml` like several of the other examples (then the templating can add a "Play With Docker" button for users to try it directly), but that's going to be the image maintainers' call.  No matter what, it'll need to move out of `README.md` (which is a generated file) and into the templating files around it.

cc @bnpfeife (`influxdb` image maintainer)

RobsOnWaves:
> I would suggest putting this in a `stack.yml` like several of the other examples (then the templating can add a "Play With Docker" button for users to try it directly), but that's going to be the image maintainers' call. No matter what, it'll need to move out of `README.md` (which is a generated file) and into the templating files around it.
> 
> cc @bnpfeife (`influxdb` image maintainer)

@tianon Thanks for the suggestion, I proceeded accordingly for the readme part and pushed the correction in the fork

PR #2113, opened 2022-02-07
Update docs for registry v2.8.0 release

milosgajdos:


registry/get-help.md
```diff
@@ -1,3 +1 @@
-[the Docker Community Forums](https://forums.docker.com/), [the Docker Community Slack](https://dockr.ly/slack), or [Stack Overflow](https://stackoverflow.com/search?tab=newest&q=docker)
-
--	**Need a free, secure hosted registry?** Try [Docker Hub](https://www.docker.com/pricing).
```
tianon, line 3: This was added over in #1798 in conversation with @j0hnst0n - I guess now that Distribution is CNCF, this is a bit more complicated. :grimacing:

(Apologies if a conversation around this already happened and the removal was intentional/pre-discussed!)

tianon (commented):
Nice!  Great to see a logo. :+1:

Overall, this looks good - just a couple minor comments / discussion points.

FWIW, `maintainer.md` is intended to be the _image_ maintainers, not the software maintainers, so I would suggest we wait to change that until we work out where the final home of the image repository is :see_no_evil: (for some reason GitHub's having a glitch and won't let me comment on the line from that file directly)

registry/get-help.md
milosgajdos: There was a discussion about all removing references to Docker (Hub) from all docs now that distribution has been donated to CNCF, yeah. Hence this change.

milosgajdos:
> FWIW, `maintainer.md` is intended to be the _image_ maintainers, not the software maintainers

We basically followed what `fluentd` have done, but I see what you mean. Yeah, we need to decide where the official image repo lives before we make this change.

registry/content.md
```diff
@@ -1,6 +1,8 @@
 # Docker Registry
```
thaJeztah, line 1: This probably should be changed to something else

I _think_ this was part of the reason for adding https://github.com/docker-library/docs/pull/2113/files#r801121627 (https://github.com/docker-library/docs/pull/1798), as users tended to confuse `Docker Registry` and `Docker Hub (Registry)`.

milosgajdos: Should it be simply just `Distribution`?

registry/content.md
```diff
@@ -1,6 +1,8 @@
...
 
-This image contains an implementation of the Docker Registry HTTP API V2 for use with Docker 1.6+. See [github.com/docker/distribution](https://github.com/docker/distribution) for more details about what it is.
+%%LOGO%%
+
+This image contains an implementation of the Docker Registry HTTP API V2 for use with Docker 1.6+. See [github.com/distribution/distribution](https://github.com/distribution/distribution) for more details about what it is.
```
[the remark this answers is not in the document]
milosgajdos: I think mentioning V2 is still correct -- especially for this release! But we can certainly add the mention of OCI, though I'm not sure how compliant v2.8.0 with OCI is...`main` branch is fully compliant, though.

registry/get-help.md
milosgajdos: https://github.com/docker-library/docs/pull/2113/commits/ffb8ea0a91cd22a64143cd75cb151de3c8a0a1db

milosgajdos:
> FWIW, maintainer.md is intended to be the image maintainers, not the software maintainers, so I would suggest we wait to change that until we work out where the final home of the image repository is 🙈

Makes sense, I've put the original ref back: https://github.com/docker-library/docs/pull/2113/commits/3846c5cb1d9d9ad1bb2455c07e1ea2c085fe836d

registry/content.md
tianon: Should we merge for now with this as-is, or do you want to resolve this before merge?

milosgajdos: let me rename it to `Distribution`, @tianon, since we now have this shiny new logo :)

PR #2118, opened 2022-02-20
Haskell - add notes on slim variant

AlistairB:
Based on https://github.com/docker-library/official-images/pull/11899 changes


tianon:
Now that https://github.com/docker-library/official-images/pull/11899 is merged, it's more clear that you already get something similar to this automatically: https://github.com/docker-library/docs/tree/master/haskell#haskellversion-slim

If you'd like to customize it, I'd suggest copying `.template-helpers/variant-slim.md` into the `haskell` folder and adjusting it. :smile:

PR #2154, opened 2022-05-17
Warn about Docker host charset and newlines with httpd custom config

fterrani:
Greetings,

`httpd` Docker image's `README.md` is quite concise and useful. However I found a small problem in the *Configuration* section, more particularly with this shell command:
```console
$ docker run --rm httpd:2.4 cat /usr/local/apache2/conf/httpd.conf > my-httpd.conf
```
If this command is run in Windows, `httpd.conf`'s charset is converted to UTF-16 LE BOM and newlines go from `LF` to `CRLF`. This makes the configuration parsing fail when the resulting file is copied back into a Docker container based on `httpd`.

Since this command transfers data from a Docker container to the host system, I think it would be worth at least mentioning that stdout redirection can alter charset on non-Linux operating systems. I also added a Windows version of the command since this might help many.

### DIFF
--- a/httpd/content.md
+++ b/httpd/content.md
@@ -42,6 +42,12 @@ To customize the configuration of the httpd server, first obtain the upstream de
 $ docker run --rm %%IMAGE%%:2.4 cat /usr/local/apache2/conf/httpd.conf > my-httpd.conf
 ```
 
+Note that stdout redirection will be handled by the Docker host. Some systems (notably Windows) will alter character encoding and newlines, which might prevent your custom configuration from being parsed properly and httpd from starting. You might need to use an host-specific binary-safe command to prevent this. E.g. on Windows PowerShell:
+
+```powershell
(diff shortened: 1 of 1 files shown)


tianon:
Hmm, not sure how I feel about including powershell-specific information here (it would be the first image to do so).

fterrani:
Considering that, I can understand the reluctance. It's probably better to remove the PowerShell example.

Should we keep the mention about the Docker host's influence on the stdout redirection? Or should we simply close this PR?

tianon:
IMO probably close, but I don't want to discourage you from contributing -- it's appreciated :smile: :heart:

fterrani:
Sorry for the late reply, life got me busy!

I understand your position. By the way, I did some quick and dirty tests using `cmd` instead of PowerShell. To my surprise and in my case, `cmd` doesn't seem to alter encoding like PowerShell does (I guess that's Microsoft's definition of progress? ;) ).

More seriously, this shows it actually depends on the *shell* used rather than on the operating system (which is completely logical now that I think of it). I still think it is worth a short warning to the user regarding the potential influence of the shell used to launch that command. But I might not have an accurate overview of the situation.

I suggest to let this PR open for 2 or 3 months and if people don't manifest their interest here, I will close it.

fterrani:
I modified the suggested changes to make the warning as general and concise as I could.

PR #2156, opened 2022-05-24
chore(mediawiki): add explicit volume definitions

### LINKED ISSUE #2146: [MediaWiki] Request to include volume specification or otherwise a notice for compose examples dealing with databases
I'll admit this is partly my fault but I can see other people making this mistake too, I setup Mediawiki a good while back and recently went to pull everything down, what I didn't realise is that I didn't have a volume setup for my database. Yes, I shouldn't have rushed the setup and just copy pasted the example, but I'd rather save anyone else from making the same mistake. Needless to say, I lost all my data there.

The relevant section I'm talking about is [in this README](https://github.com/docker-library/docs/tree/master/mediawiki#-via-docker-stack-deploy-or-docker-compose).

I propose either:
- Adding an example volume
- Adding a notice below the example explicitly stating that there is no storage configured for the database

I also quickly glanced and saw that this also applies to the teamspeak image and I assume others too.
If this is a change you're willing to make, I'd be happy to go through and make a PR using one of the two solutions.

insertish:
Going by example of [Wordpress](https://github.com/docker-library/docs/blob/master/wordpress/stack.yml), I added explicit volume definitions to the example `stack.yml` so it's sufficient to copy and paste the configuration and have persistence.

Fixes #2146.

### DIFF
--- a/mediawiki/stack.yml
+++ b/mediawiki/stack.yml
@@ -12,7 +12,7 @@ services:
     links:
       - database
     volumes:
-      - /var/www/html/images
+      - images:/var/www/html/images
       # After initial setup, download LocalSettings.php to the same directory as
       # this yaml and uncomment the following line and use compose to restart
       # the mediawiki service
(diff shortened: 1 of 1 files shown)


tianon:
cc `mediawiki` image maintainers: @davidbarratt @legoktm @addshore

legoktm:
Thanks for the ping, just a note that I won't have time to review this properly until early next week (if someone else can do it sooner, please don't wait for me).

addshore:
This looks pretty good to me!

PR #2160, opened 2022-06-09
Update Eggdrop docs

vanosg:


eggdrop/content.md
```diff
@@ -26,13 +26,13 @@ This variable sets the IRC server Eggdrop will connect to. Examples are:
 
 ```console
   -e SERVER=just.a.normal.server
-  -e SERVER=you.need.to.change.this:6667
-  -e SERVER=another.example.com:7000:password
-  -e SERVER=[2001:db8:618:5c0:263::]:6669:password
-  -e SERVER=ssl.example.net:+6697
+  -e SERVER="you.need.to.change.this 6667"
+  -e SERVER="another.example.com:7000 password"
+  -e SERVER="[2001:db8:618:5c0:263::] 6669 password"
```
tianon, line 31: You've got three different syntaxes here - intentional or typo?
(`server port`, `server:port password`, `server port password`)

Essentially, is it supposed to be space or colon? :smile:

PR #2166, opened 2022-06-21
[oraclelinux] update docs to improve the first use experience

Djelibeybi:
I'm trying to make that fact more obvious in the docs that the Docker Hub autogenerated pull command at the top of the page is invalid for this image.

Related: is there any way we can reduce or remove the Quick Start/Tags content at the top of the page? Or move it lower down somewhere? Or fix Docker Hub so it knows not to autogenerate invalid pull commands?


tianon:
This change seems fine to me -- I've raised a thread internally discussing the invalid pull commands being generated in the UI. :see_no_evil:

Djelibeybi:
Thanks!

Djelibeybi:
> We'd rather keep this consistent across images 

Sure, but does it need to be so large? 😉

tianon:
If you've got alternative suggestions for formatting it, I'd love to see them :eyes:

PR #2174, opened 2022-08-07
Create settings.json

sss6672:


tianon:
Are you sure this is filed in the right place, or was this opened accidentally?  (There's a lot of commits here and some unrelated `.vscode` content we definitely won't want.)

If this was not opened in error, it would probably help to describe in a bit more detail what it is you're trying to accomplish/solve.

yosifkit:
UGH @github; this was not a merged PR. This was a spammy PR that force pushed to be equal to master.

PR #2177, opened 2022-08-11
Adjust ghost docs to use "development" mode to use sqlite

yosifkit:
MySQL is the default db for production mode now; to avoid all the envs for connecting to a `mysql` instance on every `docker run`, we'll just use dev mode; stack file still has an example of production mode with a MySQL server.

Related to https://github.com/docker-library/ghost/pull/323

Perhaps we need a specific call out directing to the stack for a production example? Other notes that should be added?

### DIFF
--- a/ghost/content.md
+++ b/ghost/content.md
@@ -8,7 +8,7 @@ Ghost is a free and open source blogging platform written in JavaScript and dist
 
 # How to use this image
 
-This will start a Ghost instance listening on the default Ghost port of 2368.
+This will start a Ghost development instance listening on the default Ghost port of 2368.
 
 ```console
 $ docker run -d --name some-ghost -e NODE_ENV=development %%IMAGE%%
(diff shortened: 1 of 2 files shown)


tianon:
@ErisDS I realize containers aren't your thing and you don't have the bandwidth for _too_ much thought here, so I'll try to be brief and targeted :innocent: :heart:

For context, we try to keep our documentation as light as we can, focused on just the basics.  Here, we've updated most of the "simple one-liner" examples to use `NODE_ENV=development`, but our long-form Compose ("full stack") example is still using the full set of MySQL environment variables (as before).  Do you feel like that's sufficient?

IMO this clearly communicates that this is a development configuration but leaves "production configuration" somewhat as an exercise for the reader, which is kind of where we'd rather be than trying to squeeze a full Ghost tutorial in our one-pager on Hub. :sweat_smile:

ghost/content.md
```diff
@@ -11,15 +11,15 @@ Ghost is a free and open source blogging platform written in JavaScript and dist
 This will start a Ghost instance listening on the default Ghost port of 2368.
```
ErisDS, line 11: > This will start a Ghost development instance

maybe?

tianon:
Thank you so much - it really is very appreciated. :bow:

Here's what the commit I just pushed looks like when the templates are applied/rendered:

---

## Production mode

To run Ghost for production you'll also need to be running with MySQL 8, https, and a reverse proxy configured with appropriate `X-Forwarded-For`, `X-Forwared-Host`, and `X-Forwarded-Proto` (`https`) headers.

The following example demonstrates some of the necessary configuration for running with MySQL. For more detail, see [Ghost's "Configuration options" documentation](https://ghost.org/docs/config/#configuration-options).

Example `stack.yml` for `ghost`:

```yaml
version: '3.1'

services:

  ghost:
    image: ghost:4-alpine
    restart: always
    ports:
      - 8080:2368
    environment:
      # see https://ghost.org/docs/config/#configuration-options
      database__client: mysql
      url: http://localhost:8080
      #NODE_ENV: development

  db:
    image: mysql:8.0
    restart: always
    environment:
      MYSQL_ROOT_PASSWORD: example
```

[Try in PWD](http://play-with-docker.com?stack=https://raw.githubusercontent.com/docker-library/docs/57f9907ee5bbeaede8c97f439b9c11bc1081dd75/ghost/stack.yml)

tianon:
I'm going to merge this for now, especially so we can merge https://github.com/docker-library/ghost/pull/323 and get 5.9 released :sweat_smile:

If you have more suggestions, we're very eager to implement them, and appreciate the help you've provided to us as we navigate this. :heart:

tianon:
For users who find this because their deployments broke, see https://github.com/docker-library/ghost/pull/323#issuecomment-1212554428

PR #2178, opened 2022-08-15
Add Azul Zulu to list of alternative JDK's

HanSolo:


tianon:
Thank you for the contribution! :heart:

Unfortunately, our intent with this list is to highlight other images within the program which are suitable alternatives, not to maintain a comprehensive list of all possible alternatives on Docker Hub (or even "some" - we'd have to draw the line somewhere, and it'd be hard to define that line objectively otherwise). :see_no_evil:

HanSolo:
Hmm…understand even if I think Azul Zulu is not some unknown distribution given the fact that you also mention alternatives like Corretto or SAP Machine… 🫤

tianon:
Right, I agree!  Zulu is definitely not an _unknown_ distribution, but it is not part of the Docker Official Images program (yet? :eyes:), for which this is the documentation, where these other implementations _are_. :sweat_smile:

HanSolo:
Ah ok…so how could it become part of it then? 😁

tianon:
:grinning: I'd suggest starting with a read-through of https://github.com/docker-library/official-images#contributing-to-the-standard-library -- feel free to reach out if you read through that and end up with more questions. :+1:

HanSolo:
Will do, thx 👍🏻😁

tianon:
(Closing for now :heart:)

PR #2203, opened 2022-09-22
added MariaDB _HASH environment variables

TheAlgorythm:
This is due to the idea in https://github.com/MariaDB/mariadb-docker/issues/458 and the merge of https://github.com/docker-library/official-images/pull/13184.


grooverdan:
Looks good to me.

PR #2205, opened 2022-09-22
Add deprecation notice for CentOS

tianon:
Closes https://github.com/docker-library/official-images/issues/11679

### DIFF
--- a/centos/deprecated.md
+++ b/centos/deprecated.md
@@ -1,0 +1,1 @@
+This image is no longer supported/maintained (non-EOL tags last updated November 16, 2020, [docker-library/official-images#9102](https://github.com/docker-library/official-images/pull/9102); see also https://www.centos.org/centos-linux-eol/ and [docker-library/docs#2205](https://github.com/docker-library/docs/pull/2205)). Please adjust your usage accordingly.
--- a/centos/README-short.txt
+++ b/centos/README-short.txt
@@ -1,1 +1,1 @@
-The official build of CentOS.
+DEPRECATED; The official build of CentOS.


centos/deprecated.md
```diff
@@ -0,0 +1 @@
+This image is no longer supported/maintained (last updated September 15, 2021, [docker-library/official-images#10304](https://github.com/docker-library/official-images/pull/10304); see also https://www.centos.org/centos-linux-eol/). Please adjust your usage accordingly.
```
yosifkit, line 1: Should this point to the earlier PRs that actually changed `7.x` (https://github.com/docker-library/official-images/pull/9082 and https://github.com/docker-library/official-images/pull/9102 for `ppc64le` and https://github.com/docker-library/official-images/pull/5146 for `arm32v7` :see_no_evil:)?

tianon: Yeah, it's a bit complicated!  This PR is the last time the image maintainer updated any tag of the image, but there's one slightly newer where you removed the EOL versions (which is what the PR I linked to was updating) where these are the most recent PRs to update an ostensibly "supported" version. :weary:

Maybe we should change the wording to make it clear this is the last time the maintainer updated any part of the image, not necessarily the supported tags?  Maybe we should instead drop this bit of the message entirely?  Not sure.

tianon: Ok, updated with some alternative wording for your consideration. :bow:

tianon:
For future readers of this PR, I'd like to be absolutely clear that I'd love to revert this! :heart:

What's necessary for that to happen is a proper update to the supported versions of CentOS, and a commitment to keep them appropriately updated in the future (on a timely schedule -- we recommend once a month at the very most). :bow:

yosifkit (approved):

PR #2207, opened 2022-09-27
mariadb: MARIADB_ROOT_PASSWORD_HASH to set of manditory vars

grooverdan:
Order the manditory MARIADB_ROOT_PASSWORD* by decreasing strength.

Add ref MARIADB_PASSWORD_HASH to MARIADB_ROOT_PASSWORD_HASH for setting password.


grooverdan:
Got inspired to document more (even though I'm still over the size limit).

`skip-host-cache` was effectively `host-cache-size=0` but I've documented the newer form in advance of https://github.com/MariaDB/mariadb-docker/pull/462 coming in next release.

tianon (approved):

PR #2210, opened 2022-10-01
Fix mysql import in nextcloud readme

TheDevPanda:
[README] In `docker-compose exec db sh -c "mysql -u USER -p PASSWORD nextcloud < /dmp"`:
Replaced `-p` with `--password` because of: _'If you use the short option form (-p), you cannot have a space between the option and the password.'_ - mysql man page.
Also replaced `-u` with `--user` for consistency.

This fix is necessary since the command otherwise failes because of a wrong password.


tianon:
cc @tilosp @J0WI

PR #2212, opened 2022-10-07
Readme Typo when referring to pre-existing db folder

rouralberto:
In the current version, reading this may cause confusion.


yosifkit:
Seems fine to me.

cc @grooverdan

grooverdan:
me too. Got over "not"ed

PR #2215, opened 2022-10-07
Link to more Stack Exchange sites

tianon:
Also, link directly to their "What topics can I ask about here?" pages instead of to a search.


tianon:
This also drops the link to the Docker Community Forums (https://forums.docker.com/) which I understand to no longer be very active/recommended.

PR #2234, opened 2022-11-22
Fix builder docker image example reference

pofallon:
The docker build example had the version and `builder` reversed.

### DIFF
--- a/caddy/README.md
+++ b/caddy/README.md
@@ -149,7 +149,7 @@ Caddy is extendable through the use of "modules". See https://caddyserver.com/do
 You can use the `:builder` image as a short-cut to building a new Caddy binary:
 
 ```Dockerfile
-FROM caddy:<version>-builder AS builder
+FROM caddy:builder-<version> AS builder
 
 RUN xcaddy build \
     --with github.com/caddyserver/nginx-adapter \


caddy/README.md
```diff
@@ -149,7 +149,7 @@ Caddy is extendable through the use of "modules". See https://caddyserver.com/do
 You can use the `:builder` image as a short-cut to building a new Caddy binary:
 
 ```Dockerfile
-FROM caddy:<version>-builder AS builder
+FROM caddy:builder-<version> AS builder
```
tianon, line 152: Hmm, looking at the `builder`-related tags of `caddy`, I think what's here is already correct?
(perhaps you're confusing the `alpine` and `windowsservercore` variants for the versions?)

```console
$ bashbrew list caddy | grep builder
caddy:2.6.2-builder-alpine
caddy:2.6-builder-alpine
caddy:2-builder-alpine
caddy:builder-alpine
caddy:2.6.2-builder
caddy:2.6-builder
caddy:2-builder
caddy:builder
caddy:2.6.2-builder-windowsservercore-1809
caddy:2.6-builder-windowsservercore-1809
caddy:2-builder-windowsservercore-1809
caddy:builder-windowsservercore-1809
caddy:2.6.2-builder
caddy:2.6-builder
caddy:2-builder
caddy:builder
caddy:2.6.2-builder-windowsservercore-ltsc2022
caddy:2.6-builder-windowsservercore-ltsc2022
caddy:2-builder-windowsservercore-ltsc2022
caddy:builder-windowsservercore-ltsc2022
caddy:2.6.2-builder
caddy:2.6-builder
caddy:2-builder
caddy:builder
```

(cc @hairyhenderson)

hairyhenderson: _oh my_... I can see how this is confusing. But indeed, by `<version>` the docs mean something like `2.6.2`.

So this is already correct.

hairyhenderson (commented):
The current text is correct - `<version>-builder` is the right pattern, like `2.6.2-builder` for example.

IMO this should be closed

caddy/README.md
pofallon: Ah I see.  Yes, you're right -- I interpreted "version" as `[alpine | windowsservercore...]` etc. which gave me `alpine-builder` which didn't work :smile:

PR #2246, opened 2022-12-07
Improve Hashicorp Vault docker run example for server mode

theodiefenthal:
… to actually be directly usable and fix backend to current wording of storage.

Current version leaves new user of HashiCorp image questioning on how to progress. Vault starts without any listener, doesn't open any port and is hence not usable for any quickstart or test. With this change, a user can directly call localhost:8200 in his browser and initialize and navigate in Vault UI. 

Further, its important to change wording from `backend` to `storage` as there aren't many examples of Vault JSON config out there and with the example naming the storage  `backend` it might be misleading for new users that HCL configs can't be translated easily to JSON configs.

### DIFF
--- a/vault/content.md
+++ b/vault/content.md
@@ -45,16 +45,24 @@ As an example:
 $ docker run --cap-add=IPC_LOCK -e 'VAULT_DEV_ROOT_TOKEN_ID=myroot' -e 'VAULT_DEV_LISTEN_ADDRESS=0.0.0.0:1234' %%IMAGE%%
 ```
 
-## Running Vault in Server Mode
+## Running Vault in Server Mode for Development
 
 ```console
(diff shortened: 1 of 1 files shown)


tianon:
cc `vault` image maintainers: @mladlow @calvn @ncabatoff

mladlow (commented):
Thanks @theodiefenthal! I think this will make this readme easier to use.

We wanted to clarify that we wouldn't recommend this configuration in production though - would you mind making the suggested changes?

I couldn't make this a suggested change, and if you prefer I can do this myself in a follow-up PR, but we would also like to add between lines 58 & 60:

We suggest volume mounting a directory into the Docker image in order to give both the configuration and TLS certificates to Vault. You can accomplish this with:

`$ docker run --volume config/:/vault/config.d ...`

For more scalability and reliability, we suggest running containerized Vault in an orchestration environment like k8s or OpenShift.

theodiefenthal:
Thanks for the further suggestions @mladlow , I agree that its much better to give an explicit note that the second option is also not meant for production. I hope I made all changes at your like.

mladlow:
@theodiefenthal yes, looks good. Thank you for the improvements! @tianon I cannot approve but if you could merge this we'd appreciate it.

PR #2257, opened 2023-01-06
Change from WEBrick to puma

petiepooo:
Ever since https://github.com/docker-library/redmine/pull/261 the default uses puma instead of WEBrick.  Puma is considered production-ready whereas WEBrick was not recommended for production.


petiepooo:
I saw content updated more recently than readme..  Please feel free to take it from here.  Thanks!

PR #2261, opened 2023-01-13
Update README.md

GlozingNeuter:
Correction to a small typo


tianon:
:sweat_smile: great catch!

However, this change needs to be made in `caddy/content.md` instead (as `README.md` is generated) -- would you like to update and move it, or would you rather I take it from here? :bow:

GlozingNeuter:
Oh sorry, I'll try that again with the right file. Thank you very much for pointing it out.

PR #2264, opened 2023-01-17
[ArangoDB] Update logo, claim, key features, managed service

Simran-B:


tianon:
FYI for `arangodb` image maintainers: @fceller @dothebart @KVS85

(Nothing here looks very objectionable and I think these folks are your colleagues, but just in case / for visibility :smile: :+1:)

Simran-B:
Approved by our CTO himself 🙂
But it's good that you tag the official maintainers.
Thanks!

PR #2265, opened 2023-01-17
Add usage examples for PowerShell and Windows CMD

keeganwitt:


tianon:
It's not _exactly_ the same thing, but https://github.com/docker-library/docs/pull/2154#issuecomment-1129461333 is pretty related. :sweat_smile:

(However, you're the maintainer and these are your docs for your images, so if you feel strongly that this extra verbosity should be included, I won't block it. :bow:)

PR #2268, opened 2023-01-18
Update README.md

sktomer:
Update the section about AL2 LTS with the latest EOL timelines.


sktomer:
check failed. will update content.md file instead.

PR #2275, opened 2023-01-27
Update Rust versions in examples

lfrancke:
I originally opened this here: https://github.com/rust-lang/docker-rust/issues/129

Old Rust versions are referenced, when you just update them without touching anything else, there's a good chance your resulting image will not work as later builders are using bullseye but the examples still use buster which results in a mismatch in Glibc version.


yosifkit:
Thanks! LGTM (cc rust image maintainer @sfackler just for visibility)

Issue #2277, opened 2023-02-05
Inaccurate instruction: Accessing a service of a container by container IP from the host


candleindark:
https://github.com/docker-library/docs/blob/bb6b29b9a31fe8fe9e083c989b3091914e301fd6/rabbitmq/README.md?plain=1#L159

The above instruction doesn't seem to be correct. By default, I don't seem to be able to access a service provided by a container through the container's IP address from the host. This [post](https://stackoverflow.com/questions/68310353/i-cant-access-to-docker-container-directly-from-his-ip) confirms my experience as well.

yosifkit:
I am unable to reproduce a problem on a regular Linux system using the bridge network; it works without using `--publish/-p`. The `-p` is required to enable access from outside the Docker host. For example, running Linux containers using Docker Desktop, then it is required to access it from the host (like Windows or OSX) since Docker is inside a VM.

```console
$ docker run -d --hostname my-rabbit --name some-rabbit rabbitmq:3-management
Unable to find image 'rabbitmq:3-management' locally
3-management: Pulling from library/rabbitmq
7608715873ec: Pull complete 
5063b4e8e152: Pull complete 
89b7cb89a912: Pull complete 
27c7cee8f48c: Pull complete 
7204ec11924c: Pull complete 
33794a61c918: Pull complete 
c99fd6551816: Pull complete 
e8a147f91d59: Pull complete 
0c314e21ffac: Pull complete 
1d8181beff16: Pull complete 
e30eae262367: Pull complete 
Digest: sha256:ee66395719d01e6122ab4868dda83cc54e787a02118c6bcadc86aea0203b5669
Status: Downloaded newer image for rabbitmq:3-management
49005e12e1be8541b08f70167d4b3b904c27f005a4c24de08626a6c4cef62bb1
 $ docker ps
CONTAINER ID   IMAGE                       COMMAND                  CREATED         STATUS         PORTS                                                                                        NAMES
49005e12e1be   rabbitmq:3-management       "docker-entrypoint.s…"   3 seconds ago   Up 2 seconds   4369/tcp, 5671-5672/tcp, 15671-15672/tcp, 15691-15692/tcp, 25672/tcp                         some-rabbit
$ docker container inspect --format '{{ .NetworkSettings.Networks.bridge.IPAddress }}' some-rabbit
172.17.0.4
$ ping 172.17.0.4
PING 172.17.0.4 (172.17.0.4) 56(84) bytes of data.
[...]

PR #2278, opened 2023-02-05
Publish port of management console as intended

candleindark:
The command in the containing example need to publish the port of the management console to make this example work.


yosifkit:
https://github.com/docker-library/docs/issues/2277#issuecomment-1419675273

candleindark:
Even if `-p` is not needed for publish the port on Linux platforms, don't you still need it to map the default management console port of 15672 to 8080?

PR #2280, opened 2023-02-08
Update readme with Apache2 setup

nicoandresr:
Adds the information about how to set up apache2 as a reverse proxy with the appropriate headers

### DIFF
--- a/ghost/README.md
+++ b/ghost/README.md
@@ -149,7 +149,7 @@ version: '3.1'
 services:
 
   ghost:
-    image: ghost:4-alpine
+    image: ghost:5
     restart: always
     ports:
       - 8080:2368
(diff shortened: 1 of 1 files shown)


ghost/README.md
```diff
@@ -164,18 +164,52 @@ services:
...
+volumes:
+  ghost-data:
+  db-data:
 ```
 
 [![Try in PWD](https://github.com/play-with-docker/stacks/raw/cff22438cb4195ace27f9b15784bbb497047afa7/assets/images/button.png)](http://play-with-docker.com?stack=https://raw.githubusercontent.com/docker-library/docs/57f9907ee5bbeaede8c97f439b9c11bc1081dd75/ghost/stack.yml)
 
+### Production mode with Apache2 as reverse proxy
```
yosifkit, line 187: This is just a general reverse proxy setup, so not something we want to focus our very limited documentation on.

ghost/README.md
```diff
@@ -149,7 +149,7 @@ version: '3.1'
 services:
 
   ghost:
-    image: ghost:4-alpine
+    image: ghost:5
```
yosifkit, line 152: I'm not opposed to adding these updates to the docker-compose yaml file, but they do need to be in [`ghost/stack.yml`](https://github.com/docker-library/docs/blob/d9a9f142121491b0c43968548bb76eba9e6b15b7/ghost/stack.yml) since the `ghost/README.md` file is generated.

Technically, the compose file will work as-is on docker-compose while still preserving the image-defined volumes when creating new containers (thus preserving state), they just won't have "pretty" names to keep track of them.

yosifkit (changes requested):
Any changes need to be made in the `stack.yml` or `content.md` since `README.md` is generated from them.

PR #2283, opened 2023-02-23
[hashicorp/consul] Add deprecation notice

mdeggies:
Adds an official image deprecation notice/warning for `consul`, and points end users to the verified publisher images at hashicorp/consul.

If there is another way to add a deprecation notice, or other files that should be updated in this PR, please let me know! :)


yosifkit:
We have code for a `deprecated.md` that will ensure it gets put at the top of the Docker Hub page content. Like [openjdk `deprecated.md`](https://github.com/docker-library/docs/blob/c5d210501aae2bababfb72a3403f822765cd15a6/openjdk/deprecated.md) as seen in the generated [`README.md`](https://github.com/docker-library/docs/blob/c5d210501aae2bababfb72a3403f822765cd15a6/openjdk/README.md) and on [Docker Hub](https://hub.docker.com/_/openjdk).

Is there a timeline for the 1.16 release?  As I understand this wording, `1.15.*` will be added to the official images, but `1.16.*` will not. Will the `1.14` and `1.15` release series still be updated in official images after the `1.16` transition? Do you want us to delete the `latest` tag once `1.16` is released (and the `library/consul` updated to not reference a `latest` tag)?

david-yu:
Hi @yosifkit our plan is to stop producing any patches for any releases (including n1.15.x, 1.14.x) at the time 1.16.0 will GA. This will be roughly 4 months from now in June 2023.

mdeggies:
Thanks @yosifkit! I just updated the PR. 

At the time 1.16 is released, it would be great to make those additional updates you mentioned (deleting the `latest` tag and updating references to it) to nudge users to adopt the verified publisher images instead. Let me know where we can put in a request for this!

yosifkit:
To have a us delete specific tags from Docker Hub, just include a github comment in the  PR that updates the `library/consul` file that removes those tags from official images. (https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fconsul)

mdeggies:
Got it, we will be sure to do this in June at the time of the 1.16 release. Thanks for all your help!

mdeggies:
Let me know if there are any other todo's prior to merging, and thanks again!

tianon:
We're sorry to see you go, and our door is definitely open for helping however we can to bring these back or continue to keep them in DOI as well, but we totally understand that from a publisher perspective this is difficult to maintain in multiple places/pipelines. :bow: :heart:

PR #2289, opened 2023-03-07
Add permission detail influxdb readme

Beanow:
Just spent the better part of two hours finding this little detail out the hard way :]


tianon:
cc `influxdb` image maintainers: @bnpfeife @jeffreyssmith2nd

(This would also need to move to `content.md`, but I can help with that if you need :bow:)

jeffreyssmith2nd (commented):
Looks good to me, thanks for the fix! :shipit:

PR #2290, opened 2023-03-07
Update Caddy logo

mholt:
Caddy [has a new look](https://caddy.community/t/the-new-caddy-logo/18315), so this change is to update the logo for the Docker image docs.


tianon:
Nice, I like it! :+1:

PR #2291, opened 2023-03-07
Add deprecation notice to Vault

mladlow:
I'd like to update our docker image docs to warn about an impending switch to publishing only the Verified Publisher images per a deprecation notice here: https://developer.hashicorp.com/vault/docs/deprecation.

I've also gotten the Verified Publisher documentation (https://hub.docker.com/r/hashicorp/vault) updated though it doesn't have all the fancy updates from the update script.

### DIFF
--- a/vault/deprecated.md
+++ b/vault/deprecated.md
@@ -1,0 +1,2 @@
+Upcoming in Vault 1.14, we will stop publishing official Dockerhub images and publish only our Verified Publisher images. Users of Docker images should pull from [hashicorp/vault](https://hub.docker.com/r/hashicorp/vault) instead of [vault](https://hub.docker.com/_/vault). Verified Publisher images can be found at https://hub.docker.com/r/hashicorp/vault.
+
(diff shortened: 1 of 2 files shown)


tianon:
This should probably match https://github.com/docker-library/docs/pull/2283, right? :sweat_smile:
(esp. the bits about using our `deprecated.md` functionality to make this more prominent :eyes:)

I'd also very explicitly echo my comment from https://github.com/docker-library/docs/pull/2283#issuecomment-1454060662: :heart:

> We're sorry to see you go, and our door is definitely open for helping however we can to bring these back or continue to keep them in DOI as well, but we totally understand that from a publisher perspective this is difficult to maintain in multiple places/pipelines. :bow: :heart:

mladlow:
@tianon I will update this. It will take me longer than I want. Thank you for the thoughts and kind words! The multiple pipelines has become fairly difficult to maintain.

vault/deprecated.md
```diff
@@ -0,0 +1,2 @@
+Upcoming in Vault 1.14, we will stop publishing official Dockerhub images and publish only our Verified Publisher images. Users of Docker images should pull from [hashicorp/vault](https://hub.docker.com/r/hashicorp/vault) instead of [vault](https://hub.docker.com/_/vault). Verified Publisher images can be found at https://hub.docker.com/r/hashicorp/vault.
+
```
tianon, line 2:
```suggestion
```

I think this is all the CI is complaining about :sweat_smile:

(I'm happy to take care of this change if you'd rather or that would be helpful :bow:)

PR #2299, opened 2023-03-15
Update docs to reflect AL 2023 GA release

sktomer:
Hi,

The updates now mention AL2023 as the latest and recommended version.
We also removed AL2 specific information and replaced it with FAQs link for each supported OS.

Thanks,
Sumit


sktomer:
Will fix the reported errors.

PR #2315, opened 2023-04-12
Fixed grammar

eltronix:


yosifkit:
This was just a quote pulled from the Wikipedia article, but it has changed since we added this so we should probably just resync what is on Wikipedia instead.

PR #2321, opened 2023-05-06
Fix typo in "Production mode" docs

sidmclaughlin:
Fix typo, updating _X-Forwared-Host_ to _X-Forwarded-Host_.


yosifkit:
Thanks!

PR #2340, opened 2023-06-21
fix: rephrase the Postgres PGDATA documentation and add a warning

xvzf:
The content of this PR should be self-explaining.

Signed-off-by: Matthias Riegler <matthias.riegler@ankorstore.com>

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -100,7 +100,6 @@ See the PostgreSQL documentation on [`pg_hba.conf`](https://www.postgresql.org/d
 
 This optional variable can be used to define another location - like a subdirectory - for the database files. The default is `/var/lib/postgresql/data`. If the data volume you're using is a filesystem mountpoint (like with GCE persistent disks), or remote folder that cannot be chowned to the `postgres` user (like some NFS mounts), or contains folders/files (e.g. `lost+found`), Postgres `initdb` requires a subdirectory to be created within the mountpoint to contain the data.
 
-
 For example:
 
 ```console
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -96,7 +96,10 @@ See the PostgreSQL documentation on [`pg_hba.conf`](https://www.postgresql.org/d
 
 ### `PGDATA`
 
-This optional variable can be used to define another location - like a subdirectory - for the database files. The default is `/var/lib/postgresql/data`. If the data volume you're using is a filesystem mountpoint (like with GCE persistent disks) or remote folder that cannot be chowned to the `postgres` user (like some NFS mounts), Postgres `initdb` recommends a subdirectory be created to contain the data.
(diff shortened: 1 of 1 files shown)


postgres/README.md
```diff
@@ -189,7 +189,9 @@ See the PostgreSQL documentation on [`pg_hba.conf`](https://www.postgresql.org/d
 
 ### `PGDATA`
 
-This optional variable can be used to define another location - like a subdirectory - for the database files. The default is `/var/lib/postgresql/data`. If the data volume you're using is a filesystem mountpoint (like with GCE persistent disks) or remote folder that cannot be chowned to the `postgres` user (like some NFS mounts), Postgres `initdb` recommends a subdirectory be created to contain the data.
+> :warning: when mounting a volume to `/var/lib/posgresql`, the `/var/lib/postgresql/data` path is a local volume from the container runtime, thus data is not persistet on the mounted volume.
```
tianon, line 192: Unfortunately, I don't think Docker Hub supports GitHub's emoji, so this will need to be an embedded literal emoji: (also a minor typo)

```suggestion
> ⚠ when mounting a volume to `/var/lib/posgresql`, the `/var/lib/postgresql/data` path is a local volume from the container runtime, thus data is not persisted on the mounted volume.
```

Additionally, `README.md` is generated (per the comment at the top of the file :sweat_smile:), so this change will need to be applied in `content.md` instead.

(I'm happy to assist from here if you'd prefer.)

xvzf: This is fixed now @tianon! Didn't. see the `contents.md` file 👍

postgres/content.md
```diff
@@ -96,7 +96,10 @@ See the PostgreSQL documentation on [`pg_hba.conf`](https://www.postgresql.org/d
 
 ### `PGDATA`
 
-This optional variable can be used to define another location - like a subdirectory - for the database files. The default is `/var/lib/postgresql/data`. If the data volume you're using is a filesystem mountpoint (like with GCE persistent disks) or remote folder that cannot be chowned to the `postgres` user (like some NFS mounts), Postgres `initdb` recommends a subdirectory be created to contain the data.
+> **Important Note:** when mounting a volume to `/var/lib/posgresql`, the `/var/lib/postgresql/data` path is a local volume from the container runtime, thus data is not persisted on the mounted volume.
+
+This optional variable can be used to define another location - like a subdirectory - for the database files. The default is `/var/lib/postgresql/data`. If the data volume you're using is a filesystem mountpoint (like with GCE persistent disks), or remote folder that cannot be chowned to the `postgres` user (like some NFS mounts), or contains folders/files (e.g. `lost+found`), Postgres `initdb` requires a subdirectory to be created within the mountpoint to contain the data.
+
```
tianon, line 102:
```suggestion
```

:eyes: (this will appease `markdownfmt`)

PR #2343, opened 2023-06-23
Update mentioned php versions that are EOL

yosifkit:
Also update any mentioned module versions so that they build with the latest PHP.

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -11,7 +11,7 @@ PHP is a server-side scripting language designed for web development, but which 
 ### Create a `Dockerfile` in your PHP project
 
 ```dockerfile
-FROM %%IMAGE%%:7.4-cli
+FROM %%IMAGE%%:8.2-cli
 COPY . /usr/src/myapp
 WORKDIR /usr/src/myapp
 CMD [ "php", "./your-script.php" ]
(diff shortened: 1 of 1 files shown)


php/content.md
```diff
@@ -76,59 +76,57 @@ Some extensions are compiled by default. This depends on the PHP version you are
...
+	&& pecl install memcached-3.2.0 \
 	&& docker-php-ext-enable memcached
 ```
 
-It is *strongly* recommended that users use an explicit version number in their `pecl install` invocations to ensure proper PHP version compatibility (PECL does not check the PHP version compatiblity when choosing a version of the extension to install, but does when trying to install it).
+It is *strongly* recommended that users use an explicit version number in their `pecl install` invocations to ensure proper PHP version compatibility (PECL does not check the PHP version compatibility when choosing a version of the extension to install, but does when trying to install it). Beyond the compatibility issue, it's also a good practice to ensure you know when your dependencies receive updates and can control those updates directly.
 
-For example, `memcached-2.2.0` has no PHP version constraints (https://pecl.php.net/package/memcached/2.2.0), but `memcached-3.1.4` requires PHP 7.0.0 or newer (https://pecl.php.net/package/memcached/3.1.4). When doing `pecl install memcached` (no specific version) on PHP 5.6, PECL will try to install the latest release and fail.
```
yosifkit, line 94: I did not think it was worth finding a new failing example.

php/content.md
```diff
@@ -76,59 +76,57 @@ Some extensions are compiled by default. This depends on the PHP version you are
...
 Some extensions are not provided via either Core or PECL; these can be installed too, although the process is less automated:
 
 ```dockerfile
-FROM %%IMAGE%%:5.6-cli
-RUN curl -fsSL 'https://xcache.lighttpd.net/pub/Releases/3.2.0/xcache-3.2.0.tar.gz' -o xcache.tar.gz \
-	&& mkdir -p xcache \
-	&& tar -xf xcache.tar.gz -C xcache --strip-components=1 \
-	&& rm xcache.tar.gz \
+FROM %%IMAGE%%:8.2-cli
+RUN curl -fsSL '[url-to-custom-php-module]' -o module-name.tar.gz \
```
yosifkit, line 102: Happy to put this back to a real example, but `xcache.lighttpd.net` is 404 and their [github mirror](https://github.com/lighttpd/xcache) hasn't updated for years.

tianon (approved):

Issue #2304, opened 2023-03-20
Automake anchor links


danepowell:
I frequently have to link to Docker library docs. On [long pages like for the PHP image](https://hub.docker.com/_/php), it would be really nice to be able to link to a specific heading.

For instance, GitHub automakes anchor links, such as for the PHP README: https://github.com/docker-library/docs/blob/master/php/README.md#how-to-install-more-php-extensions

But the docs on hub.docker.com are missing such anchors.

tianon:
The rendering engine used for the markdown on Docker Hub has changed several times, and I don't think it's always consistently supported anchors.  When I need an anchor, I usually link to the canonical version of the docs on GitHub instead :sweat_smile:

tianon:
See also https://github.com/docker/hub-feedback/issues/956

PR #2348, opened 2023-06-28
Add NET_ADMIN documentation

otbutz:
@francislavoie

### DIFF
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -20,6 +20,14 @@ In the examples below, a [named volume](https://docs.docker.com/storage/volumes/
 
 Note that named volumes are persisted across container restarts and terminations, so if you move to a new image version, the same data and config directories can be re-used.
 
+### Linux capabilities
+
+Caddy ships with HTTP/3 support enabled by default. To improve the performance of this UDP based protocol, the underlying quic-go library tries to increase the buffer sizes for its socket. The `NET_ADMIN` capability allows it to override the low default limits of the operating system without having to change kernel parameters via sysctl.
+
+Giving the container this capability is optional and has potential, though unlikely, [security implications](https://unix.stackexchange.com/a/508816).
+
+see https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes
+
 ### Basic Usage
 
 The default config file simply serves files from `/usr/share/caddy`, so if you want to serve `index.html` from the current working directory:
@@ -49,7 +57,7 @@ $ docker run -d -p 80:80 \
 The default `Caddyfile` only listens to port `80`, and does not set up automatic TLS. However, if you have a domain name for your site, and its A/AAAA DNS records are properly pointed to this machine's public IP, then you can use this command to simply serve a site over HTTPS:
 
 ```console
-$ docker run -d -p 80:80 -p 443:443 -p 443:443/udp \
+$ docker run --cap-add=NET_ADMIN -d -p 80:80 -p 443:443 -p 443:443/udp \
     -v /site:/srv \
     -v caddy_data:/data \
     -v caddy_config:/config \
@@ -120,6 +128,8 @@ services:
   caddy:
     image: %%IMAGE%%:<version>
     restart: unless-stopped
+    cap_add:
+      - NET_ADMIN
     ports:
       - "80:80"
       - "443:443"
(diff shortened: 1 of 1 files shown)


caddy/content.md
```diff
@@ -20,6 +20,14 @@ In the examples below, a [named volume](https://docs.docker.com/storage/volumes/
 
 Note that named volumes are persisted across container restarts and terminations, so if you move to a new image version, the same data and config directories can be re-used.
 
+### Linux capabilities
```
francislavoie, line 23: I don't think `Linux capabilities` should be _above_ `Basic Usage`. I think it should go later in the document, because it's not absolutely necessary to run Caddy, it's just a nice-to-have for HTTP/3 performance.

otbutz: I wasn't sure where to put it, to be honest. Some security-conscious people may want to see this notice in the top part, but I agree that Basic Usage should go first here.

Since the automatic TLS example would already grant NET_ADMIN, should we put the section in front of it?

francislavoie: I think the point of this document is to explain _how to use the image_ and details like this are less important and should go later in the document.

I wish Docker Hub had proper markdown anchor tags like every other modern markdown renderer because then we could just add a link to jump to that section from the TLS section which mentions it. https://github.com/docker-library/docs/issues/2304 and https://github.com/docker/hub-feedback/issues/956

caddy/content.md
```diff
@@ -20,6 +20,14 @@ In the examples below, a [named volume](https://docs.docker.com/storage/volumes/
...
+
+Caddy ships with HTTP/3 support enabled by default. To improve the performance of this UDP based protocol, the underlying quic-go library tries to increase the buffer sizes for its socket. The `NET_ADMIN` capability allows it to override the low default limits of the operating system without having to change kernel parameters via sysctl.
```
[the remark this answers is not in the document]
francislavoie: :man_shrugging: you'd certainly know better than I, Docker+kernel+sys interactions are not an area I'm knowledgeable with.

But I think `NET_ADMIN` would allow the software to automatically fix the limits automatically if it finds it needs to, rather than trying to give a one-size-fits-all number. But this isn't our code doing it, it's code from quic-go, a dependency which provides QUIC and HTTP/3 support.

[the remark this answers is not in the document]
otbutz: I'll keep this as a draft and test with the final 2.7.

tianon:
https://github.com/quic-go/quic-go/issues/3418 has some more very relevant discussion, but no good conclusion :disappointed: :heart:

caddy/content.md
```diff
@@ -109,6 +109,14 @@ $ caddy_container_id=$(docker ps | grep caddy | awk '{print $1;}')
 $ docker exec -w /etc/caddy $caddy_container_id caddy reload
 ```
 
+### Linux capabilities
+
+Caddy ships with HTTP/3 support enabled by default. To improve the performance of this UDP based protocol, the underlying quic-go library tries to increase the buffer sizes for its socket. The `NET_ADMIN` capability allows it to override the low default limits of the operating system without having to change kernel parameters via sysctl.
+
+Giving the container this capability is optional and has potential, though unlikely, [security implications](https://unix.stackexchange.com/a/508816).
+
+see https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes
```
francislavoie, line 118:
```suggestion
Giving the container this capability is optional and has potential, though unlikely, to have [security implications](https://unix.stackexchange.com/a/508816).

See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for more details.
```

caddy/content.md
```diff
@@ -49,7 +49,7 @@ $ docker run -d -p 80:80 \
 The default `Caddyfile` only listens to port `80`, and does not set up automatic TLS. However, if you have a domain name for your site, and its A/AAAA DNS records are properly pointed to this machine's public IP, then you can use this command to simply serve a site over HTTPS:
 
 ```console
-$ docker run -d -p 80:80 -p 443:443 -p 443:443/udp \
+$ docker run --cap-add=NET_ADMIN -d -p 80:80 -p 443:443 -p 443:443/udp \
```
francislavoie, line 52:
```suggestion
$ docker run -d --cap-add=NET_ADMIN -p 80:80 -p 443:443 -p 443:443/udp \
```

francislavoie (commented):
This seems fine to me. If @tianon has no objections I'm okay with it.

Suggesting some minor adjustments.

otbutz:
@francislavoie Anything missing?

francislavoie:
I don't maintain this repo, I can't merge this.

otbutz:
@yosifkit

yosifkit:
Just a quick close/open to get GitHub actions to run.

PR #2351, opened 2023-07-12
Fix a typo with /usr/lib/postgres/data

AlexHayton:
Seems like a mistake in the README


yosifkit:
Fixed by https://github.com/docker-library/docs/pull/2364

PR #2358, opened 2023-08-09
Improve docker-compose install instructions

lectrician1:
Add more-specific instructions as to what to make the database host if using docker-compose. I was having trouble connecting using `localhost` until I realized you need to use `database://localhost`. This change specifies this more-directly.

### DIFF
--- a/mediawiki/README.md
+++ b/mediawiki/README.md
@@ -147,6 +147,8 @@ volumes:
 
 Run `docker stack deploy -c stack.yml mediawiki` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
 
+When specifying the database host as part of the wiki setup process, make sure to use `database://localhost`.
+
 ## Adding additional libraries / extensions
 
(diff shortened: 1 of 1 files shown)


mediawiki/README.md
```diff
@@ -147,6 +147,8 @@ volumes:
 
 Run `docker stack deploy -c stack.yml mediawiki` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
 
+When specifying the database host as part of the wiki setup process, make sure to use `database://localhost`.
```
tianon, line 150: Hmm, `localhost` wouldn't be correct though, because that would imply the container `mediawiki` is running in -- rather, that should be something like `database://db`, right?  (This would also need to be part of `content.md` as `README.md` is generated, so it might be even better to include it as a code comment inside `stack.yml`)

Also cc `mediawiki` image maintainers: @legoktm @addshore @christian-heusel

lectrician1: `localhost` would be correct though because there's are separate `mediawiki` and `database` services. For the `mediawiki` container to connect to the database it needs to use the `localhost` of the `database` container which the MySQL database is exposing itself to.

LaurentGoderre: `localhost` might work but it's not ideal to leave the compose environment and the name of the service should be used instead (`database` in this case).

PR #2359, opened 2023-08-10
Update arangodb docs

nerpaula:
Added note about macOS/ARM and how to override the default chosen processor architecture.

### DIFF
--- a/arangodb/content.md
+++ b/arangodb/content.md
@@ -31,7 +31,14 @@ In order to start an ArangoDB instance, run:
 unix> docker run -e ARANGO_RANDOM_ROOT_PASSWORD=1 -d --name arangodb-instance %%IMAGE%%
 ```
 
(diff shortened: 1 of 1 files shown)


dothebart (commented):
Thanks, Paula.

@tianon why is it that these macs rather choose the incomplete Rosetta x86 emulation than native ARM docker containers?

tianon:
Yeah, that sounds to me like a solid bug elsewhere :thinking:

@nerpaula can you give more details about what your environment looked like when you ran into this?

nerpaula:
@tianon this issue was reported here: https://github.com/arangodb/arangodb-docker/issues/114

yosifkit:
Considering that the problem is caused by configuration on the host (`DOCKER_DEFAULT_PLATFORM=linux/amd64`), I don't think this should be called out on a specific image's docs. The users' config/environment is overriding the best choice. It is unfortunate that docker doesn't warn the user when `DOCKER_DEFAULT_PLATFORM` is used.

PR #2363, opened 2023-08-25
Update README.md

timeroute:
Fixed word 'posgresql' to 'postgresql'


yosifkit:
Thank you!

This change would need to go in `content.md`, but it looks like there is a second PR with the same fix: https://github.com/docker-library/docs/pull/2364, so I'll just merge that one.

PR #2366, opened 2023-09-07
[Draft] Add sonar-scanner-cli as official docker image

jCOTINEAU:
Hello everyone, we would like to add a new official image for the tool called sonar-scanner-cli.

This is the tool used to interact with SonarQube product (which is already an official image).

Will add the link to the related [official image](https://github.com/docker-library/official-images) PR.

Those PR will stay in draft while gathering your feedback.

Thanks a lot in advance


yosifkit:
Closing since the related PR was closed: https://github.com/docker-library/official-images/pull/15326

PR #2369, opened 2023-09-19
Improve explanation about how to customize Nginx configuration

DQNEO:
First of all, thank you for the great effort to maintain this marvelous project. 

When I first saw this document, I didn't understand the meaning of `This can also be accomplished`.
I actually misinterpreted it as "Copying the default nginx configuration file from a running container to the host system can also be accomplished" , but the docker file looked doing the opposite.

So I tried to improve some points:

* "copying the default conf" relates to both mounting and building , so I moved it to the head
* It is not  necessarily "complex", so I omitted it.

### DIFF
--- a/nginx/content.md
+++ b/nginx/content.md
@@ -35,15 +35,11 @@ $ docker run --name some-nginx -d -p 8080:80 some-content-nginx
 
 Then you can hit `http://localhost:8080` or `http://host-ip:8080` in your browser.
 
-## Complex configuration
+## Customize configuration
 
-```console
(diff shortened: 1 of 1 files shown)


yosifkit:
Seems ok to me (minus the minor fix to make the markdown formatter happy).

cc the `nginx` image maintainer: @thresheek to make sure they are ok with content.

thresheek:
Thanks @yosifkit - I think the change is good.

DQNEO:
I've fix the markdown format.

LaurentGoderre (approved):

PR #2370, opened 2023-09-25
Update Ghost stack.yml to have persistent data

AnomalRoil:
This is adding named volumes to the example docker-compose file, since otherwise the uploaded data is easily lost, same for the DB.

### DIFF
--- a/ghost/stack.yml
+++ b/ghost/stack.yml
@@ -18,9 +18,17 @@ services:
       url: http://localhost:8080
       # contrary to the default mentioned in the linked documentation, this image defaults to NODE_ENV=production (so development mode needs to be explicitly specified if desired)
       #NODE_ENV: development
+    volumes:
+     - some-ghost-data:/var/lib/ghost/content
 
   db:
     image: mysql:8.0
     restart: always
     environment:
       MYSQL_ROOT_PASSWORD: example
+    volumes:
+      - some-ghost-mysql:/var/lib/mysql
+
+volumes:
+  - some-ghost-data:
+  - some-ghost-mysql:
--- a/ghost/stack.yml
+++ b/ghost/stack.yml
@@ -5,12 +5,14 @@ services:
   ghost:
     image: ghost:4-alpine
     restart: always
+    depends_on:
+      - ghost-db
     ports:
       - 8080:2368
     environment:
       # see https://ghost.org/docs/config/#configuration-options
       database__client: mysql
-      database__connection__host: db
+      database__connection__host: ghost-db
       database__connection__user: root
       database__connection__password: example
       database__connection__database: ghost
@@ -21,7 +23,7 @@ services:
     volumes:
       - some-ghost-data:/var/lib/ghost/content
 
-  db:
+  ghost-db:
     image: mysql:8.0
     restart: always
     environment:
--- a/ghost/stack.yml
+++ b/ghost/stack.yml
@@ -5,14 +5,12 @@ services:
   ghost:
     image: ghost:4-alpine
     restart: always
-    depends_on:
-      - ghost-db
     ports:
       - 8080:2368
     environment:
       # see https://ghost.org/docs/config/#configuration-options
       database__client: mysql
-      database__connection__host: ghost-db
+      database__connection__host: db
       database__connection__user: root
       database__connection__password: example
       database__connection__database: ghost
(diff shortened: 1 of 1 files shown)


LaurentGoderre (approved):

ghost/stack.yml
```diff
@@ -18,9 +18,17 @@ services:
...
     image: mysql:8.0
     restart: always
     environment:
       MYSQL_ROOT_PASSWORD: example
+    volumes:
+      - some-ghost-mysql:/var/lib/mysql
+
+volumes:
+  - some-ghost-data:
+  - some-ghost-mysql:
```
yosifkit, line 34: `docker compose` requires `volumes` to be a map, not a list of maps.
```suggestion
  some-ghost-data:
  some-ghost-mysql:
```

ghost/stack.yml
```diff
@@ -18,9 +18,17 @@ services:
       url: http://localhost:8080
       # contrary to the default mentioned in the linked documentation, this image defaults to NODE_ENV=production (so development mode needs to be explicitly specified if desired)
       #NODE_ENV: development
+    volumes:
+     - some-ghost-data:/var/lib/ghost/content
```
yosifkit, line 22: Oh, it seems the `ymlfmt` test wants one extra space indent to keep things consistent.
```suggestion
      - some-ghost-data:/var/lib/ghost/content
```

yosifkit:
Thank you for your contribution! Just a couple minor suggestions.

You are welcome to update the PR, or I am happy to take over from here.

ghost/stack.yml
AnomalRoil: Ah good catch, I tweaked my own to become "minimal" and messed that up

ghost/stack.yml
```diff
@@ -5,22 +5,32 @@ services:
   ghost:
     image: ghost:4-alpine
     restart: always
+    depends_on:
+      - ghost-db
```
tianon, line 9: `depends_on:` doesn't actually do anything meaningful :see_no_evil:

```suggestion
```

ghost/stack.yml
```diff
@@ -5,22 +5,32 @@ services:
...
       database__connection__database: ghost
       # this url value is just an example, and is likely wrong for your environment!
       url: http://localhost:8080
       # contrary to the default mentioned in the linked documentation, this image defaults to NODE_ENV=production (so development mode needs to be explicitly specified if desired)
       #NODE_ENV: development
+    volumes:
+      - some-ghost-data:/var/lib/ghost/content
 
-  db:
+  ghost-db:
```
tianon, line 26: I don't think we really need to change this, do we?  It's already scoped to this stack (in both compose and swarm).

```suggestion
  db:
```

ghost/stack.yml
```diff
@@ -5,22 +5,32 @@ services:
...
     restart: always
+    depends_on:
+      - ghost-db
     ports:
       - 8080:2368
     environment:
       # see https://ghost.org/docs/config/#configuration-options
       database__client: mysql
-      database__connection__host: db
+      database__connection__host: ghost-db
```
tianon, line 15:
```suggestion
      database__connection__host: db
```

ghost/stack.yml
```diff
@@ -5,22 +5,32 @@ services:
...
     image: mysql:8.0
     restart: always
     environment:
       MYSQL_ROOT_PASSWORD: example
+    volumes:
+      - some-ghost-mysql:/var/lib/mysql
+
+volumes:
+  some-ghost-data:
+  some-ghost-mysql:
```
tianon, line 36: We don't have any other `stack.yml` files that use this naming scheme (we use it in the prose so it's more obviously an example that should be modified, where here the entire file is that).

```suggestion
      - db:/var/lib/mysql

volumes:
  ghost:
  db:
```

ghost/stack.yml
```diff
@@ -5,22 +5,32 @@ services:
...
+      database__connection__host: ghost-db
       database__connection__user: root
       database__connection__password: example
       database__connection__database: ghost
       # this url value is just an example, and is likely wrong for your environment!
       url: http://localhost:8080
       # contrary to the default mentioned in the linked documentation, this image defaults to NODE_ENV=production (so development mode needs to be explicitly specified if desired)
       #NODE_ENV: development
+    volumes:
+      - some-ghost-data:/var/lib/ghost/content
```
tianon, line 24:
```suggestion
      - ghost:/var/lib/ghost/content
```

tianon (commented):
Sorry, I have a few more comments (just got back from some very long PTO) :see_no_evil:

tianon (approved):

PR #2379, opened 2023-10-13
Fix link in README.md

RDxR10:


LaurentGoderre (approved):

yosifkit:
The change seems fine, but it needs to be done in [`content.md`](https://github.com/docker-library/docs/blob/b28b0e94f22a4701cc40f8c629042b64c739613a/amazonlinux/content.md) because `README.md` is generated.

CC a few of the `amazonlinux` image maintainers to ensure that this is the link that they want here: @jhowell-burke @dadetayo @sktomer @mrthornazon

RDxR10:
Hello @yosifkit,
So I should wait for the verification, or go ahead with the task?

sktomer:
Please cancel this pull request and update the link here as correctly recommended by @yosifkit : https://github.com/docker-library/docs/blob/b28b0e94f22a4701cc40f8c629042b64c739613a/amazonlinux/content.md?plain=1

PR #2389, opened 2023-11-08
Update golang/content.md to explain when and how to install git-lfs

danielorbach:
The Go toolchain uses the local git installation to fetch archives of modules from Git servers directly. My company commits large files using Git LFS which happen to be part of our Go module source tree. Without _git-lfs_ installed, the local git installation does not automatically resolve those files which causes "_checksum mismatch_" against the committed `go.sum`.

This pull-request aims to ease the strain on developers that encounter that unfortunately under-documented scenario, and lower the barrier of entry for companies to the Golang world.

### DIFF
--- a/golang/content.md
+++ b/golang/content.md
@@ -69,3 +69,51 @@ $ for GOOS in darwin linux; do
 >   done
 > done
 ```
+
+## Enable `git lfs`
+
+If your application fails to `go mod verify` inside the Go container with a "_checksum mismatch_" error, like so:
+
+> ```console
+>  > [builder 4/7] RUN go mod download && go mod verify:
+> 5.888 verifying <module-import-path>: checksum mismatch
+> 5.888 	downloaded: h1:broICiQ+pcdcV/2qtCYrIIAnhQH65LuA9UvmaNjcy/w=
+> 5.888 	go.sum:     h1:maNCe3B0zkLaBfPUpJk3CBGavDPAI9ONaI4wKq2GUv0=
+> ```
+
+And the following condition apply:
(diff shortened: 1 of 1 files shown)


danielorbach:
@jbleduigou I saw you were the last to commit on this file, can you guide me towards improving this pull-request?

This is my first time contributing to this repository, I will be glad to make any adjustments necessary to fit the existing conventions and quality requirements.

danielorbach:
This pull-request follows my original suggestion to update the base `golang` image at https://github.com/docker-library/golang/pull/492

golang/content.md
```diff
@@ -69,3 +69,51 @@ $ for GOOS in darwin linux; do
...
+```
+FROM %%IMAGE%%:1.21-alpine
+...
+
+# install git-lfs for Go modules with LFS committed files, before go mod download
+RUN apk add --no-cache git-lfs
+
+RUN go mod download && go mod verify
+...
+```
```
tianon, line 119: This almost doubles the size of our maintained content for what's honestly a pretty extreme edge case (Go dependencies with static files that are large enough to warrant use of Git LFS). :sweat_smile:

I think we can still include something, but it should be a lot simpler, maybe something more like this:

---

## Git LFS

If downloading your dependencies results in an error like "checksum mismatch", you should check whether they are using [Git LFS](https://git-lfs.com/) (and thus need it installed for downloading them and calculating correct `go.sum` values).

---

<details>
<summary>GitHub "Suggested Change":</summary>

````suggestion
## Git LFS

If downloading your dependencies results in an error like "checksum mismatch", you should check whether they are using [Git LFS](https://git-lfs.com/) (and thus need it installed for downloading them and calculating correct `go.sum` values).
````

</details>

PR #2393, opened 2023-11-17
Postgres: Add healthcheck examples

LaurentGoderre:


postgres/content.md
```diff
@@ -40,6 +40,29 @@ postgres=# SELECT 1;
 
 Run `docker stack deploy -c stack.yml %%REPO%%` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
 
+## Healtcheck
+
+To add a health check to the image use the following command:
+
+```console
+docker run --name some-postgres -e POSTGRES_PASSWORD=mysecretpassword \
```
jamesbraza, line 48: Do you need `--name some-postgres`? I feel like for demonstration purposes, it's not necessary

postgres/content.md
```diff
@@ -40,6 +40,29 @@ postgres=# SELECT 1;
...
+```yaml
+services:
+  postgres:
+    image: postgres
+    restart: always
+    environment:
+      POSTGRES_PASSWORD: example
+    healthcheck:
+        test: pg_isready
+        interval: 30s
```
jamesbraza, line 63: What do you think of making this `interval` smaller?

LaurentGoderre: Certainly. I don't know what would be a good balance.

jamesbraza: I am rolling with 2 seconds, and think in general an example may be better served by 5-sec or 10-sec.  I defer to your judgement here beyond this, and thanks for doing this!

postgres/content.md
LaurentGoderre: This is to stay consistent with the previous example in the docs

LaurentGoderre: https://github.com/docker-library/docs/blob/master/postgres/content.md?plain=1#L16

postgres/content.md
LaurentGoderre: 5s seems reasonable.

postgres/content.md
jamesbraza: Okay gotchu!  Consistency is good.

Thanks for answering, you have my approval 👍 🚀

postgres/content.md
```diff
@@ -40,6 +40,29 @@ postgres=# SELECT 1;
 
 Run `docker stack deploy -c stack.yml %%REPO%%` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
 
+## Healtcheck
+
+To add a health check to the image use the following command:
+
+```console
+docker run --name some-postgres -e POSTGRES_PASSWORD=mysecretpassword \
+	--health-cmd='pg_isready' --health-interval=5s -d %%IMAGE%%
```
tianon, line 49: Won't just a raw `pg_isready` like this happily connect to the temporary unix-socket-only instance of the server that's used for initdb, and thus prematurely report that the database is up and healthy?

(see also https://github.com/docker-library/faq#healthcheck, many of which applies to trying to make "generic" documentation for health checks too, and https://github.com/docker-library/healthcheck/blob/40afbf64d69cf933af0da4df6383958a29113601/postgres/docker-healthcheck which is likely more than we'd want in the documentation but the main point there being explicitly using TCP to ensure we don't connect to the temporary server)

postgres/content.md
```diff
@@ -40,6 +40,29 @@ postgres=# SELECT 1;
 
 Run `docker stack deploy -c stack.yml %%REPO%%` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
 
+## Healtcheck
```
Chealer, line 43: s/Healt/Health/

PR #2392, opened 2023-11-15
Eggdrop: Add new ENV variables to docs

vanosg:

PR #2396, opened 2023-11-29
Fix tcl-lib typo

LaurentGoderre:
Replaces #2392

### DIFF
--- a/eggdrop/content.md
+++ b/eggdrop/content.md
@@ -88,10 +88,10 @@ to your docker run command line (and then edit your config file to load the scri
 
 ## Adding packages required for scripts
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
@vanosg

vanosg:
@LaurentGoderre You are amazing, thank you- One day my git-fu will be as good as yours. Thank you!

vanosg:
And sorry, no good deed goes unpunished- I noticed the environmental var addition (d093924) is missing from this PR

LaurentGoderre:
@vanosg those already are in the docs: https://hub.docker.com/_/eggdrop

tianon:
Thanks to you both! :smile:

vanosg:
Thanks to YOU both- I don't know how that one got so screwed up. We'll do better on the next one :)

PR #2397, opened 2023-12-06
Initial commitment of documentation for clickhouse official image

Felixoid:
Hello, dear official library team. Here's our attempt to create an official image `clickhouse` in the scope of https://github.com/docker-library/official-images/issues/14136 and https://github.com/ClickHouse/ClickHouse/issues/31473

Both the docs and official images PRs are created simultaneously.

https://github.com/docker-library/official-images/pull/15846


Felixoid:
The content.md + license.md are the same as docker/server/README.md in https://github.com/ClickHouse/ClickHouse/pull/63400

tianon:
Close/reopen to retrigger fresh GHA checks :+1:

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+
+For more information and documentation see https://clickhouse.com/.
+
+## Versions
+
+-	The `latest` tag points to the latest release of the latest stable branch.
+-	Branch tags like `22.2` point to the latest release of the corresponding branch.
+-	Full version tags like `22.2.3.5` point to the corresponding release.
+-	The tag `head` is built from the latest commit to the default branch.
+-	Each tag has optional `-alpine` suffix to reflect that it's built on top of `alpine`.
```
tianon, line 19: This was dropped from https://github.com/docker-library/official-images/pull/15846 :eyes:

```suggestion
```

Felixoid:
It looks like I missed some metadata.json there. I'll read what it is.

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+
+-	The amd64 image requires support for [SSE3 instructions](https://en.wikipedia.org/wiki/SSE3). Virtually all x86 CPUs after 2005 support SSE3.
+-	The arm64 image requires support for the [ARMv8.2-A architecture](https://en.wikipedia.org/wiki/AArch64#ARMv8.2-A) and additionally the Load-Acquire RCpc register. The register is optional in version ARMv8.2-A and mandatory in [ARMv8.3-A](https://en.wikipedia.org/wiki/AArch64#ARMv8.3-A). Supported in Graviton >=2, Azure and GCP instances. Examples for unsupported devices are Raspberry Pi 4 (ARMv8.0-A) and Jetson AGX Xavier/Orin (ARMv8.2-A).
+
+## How to use this image
+
+### start server instance
+
+```bash
+docker run -d --name some-clickhouse-server --ulimit nofile=262144:262144 clickhouse/clickhouse-server
```
tianon, line 31:
```suggestion
docker run -d --name some-clickhouse-server --ulimit nofile=262144:262144 %%IMAGE%%
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+By default, ClickHouse will be accessible only via the Docker network. See the [networking section below](#networking).
+
+By default, starting above server instance will be run as the `default` user without password.
+
+### connect to it from a native client
+
+```bash
+docker run -it --rm --link some-clickhouse-server:clickhouse-server --entrypoint clickhouse-client clickhouse/clickhouse-server --host clickhouse-server
```
tianon, line 41: I would not recommend adding _new_ documentation referencing `--link` (it is deprecated; see https://docs.docker.com/engine/network/links/ and https://github.com/docker-library/docs/pull/1441), but it's up to you -- the image reference needs to be correct though:

```suggestion
docker run -it --rm --link some-clickhouse-server:clickhouse-server --entrypoint clickhouse-client %%IMAGE%% --host clickhouse-server
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+-	The `latest` tag points to the latest release of the latest stable branch.
+-	Branch tags like `22.2` point to the latest release of the corresponding branch.
+-	Full version tags like `22.2.3.5` point to the corresponding release.
+-	The tag `head` is built from the latest commit to the default branch.
+-	Each tag has optional `-alpine` suffix to reflect that it's built on top of `alpine`.
+
+### Compatibility
+
+-	The amd64 image requires support for [SSE3 instructions](https://en.wikipedia.org/wiki/SSE3). Virtually all x86 CPUs after 2005 support SSE3.
+-	The arm64 image requires support for the [ARMv8.2-A architecture](https://en.wikipedia.org/wiki/AArch64#ARMv8.2-A) and additionally the Load-Acquire RCpc register. The register is optional in version ARMv8.2-A and mandatory in [ARMv8.3-A](https://en.wikipedia.org/wiki/AArch64#ARMv8.3-A). Supported in Graviton >=2, Azure and GCP instances. Examples for unsupported devices are Raspberry Pi 4 (ARMv8.0-A) and Jetson AGX Xavier/Orin (ARMv8.2-A).
```
tianon, line 24: As the author of https://github.com/opencontainers/image-spec/commit/2d95dde8615f9ed8a911873f8eb72a8582af577a, I love this -- I wish we had a way to specify that the `amd64` image variant should be `v2` (corresponding to the lowest level with SSE3) and the `arm64` variant could be `v8.2`, but we don't currently have a way to specify either in our system and the latter would break folks (until https://github.com/containerd/platforms/pull/8 and similar trickle down to popular runtimes), so for now I'm just happy to see it documented explicitly (hopefully ClickHouse also reports a clear error when these are missing). :smile: :heart:

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+
+## How to use this image
+
+### start server instance
+
+```bash
+docker run -d --name some-clickhouse-server --ulimit nofile=262144:262144 clickhouse/clickhouse-server
+```
+
+By default, ClickHouse will be accessible only via the Docker network. See the [networking section below](#networking).
```
tianon, line 34: Unfortunately, cross-header links like this won't work on Docker Hub's rendering. :disappointed:

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+# OR
+docker exec -it some-clickhouse-server clickhouse-client
+```
+
+More information about the [ClickHouse client](https://clickhouse.com/docs/en/interfaces/cli/).
+
+### connect to it using curl
+
+```bash
+echo "SELECT 'Hello, ClickHouse!'" | docker run -i --rm --link some-clickhouse-server:clickhouse-server curlimages/curl 'http://clickhouse-server:8123/?query=' -s --data-binary @-
```
tianon, line 51: Unfortunately, this image reference isn't acceptable here -- `buildpack-deps:curl` (or `buildpack-deps:bookworm-curl` if you want to be more explicit) might be an acceptable alternative for you?
(https://github.com/docker-library/buildpack-deps/blob/3e18c3af1f5dce6a48abf036857f9097b6bd79cc/debian/bookworm/curl/Dockerfile)

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+docker stop some-clickhouse-server
+docker rm some-clickhouse-server
+```
+
+### networking
+
+You can expose your ClickHouse running in docker by [mapping a particular port](https://docs.docker.com/config/containers/container-networking/) from inside the container using host ports:
+
+```bash
+docker run -d -p 18123:8123 -p19000:9000 --name some-clickhouse-server --ulimit nofile=262144:262144 clickhouse/clickhouse-server
```
tianon, line 68:
```suggestion
docker run -d -p 18123:8123 -p19000:9000 --name some-clickhouse-server --ulimit nofile=262144:262144 %%IMAGE%%
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+echo 'SELECT version()' | curl 'http://localhost:18123/' --data-binary @-
+```
+
+`22.6.3.35`
+
+or by allowing the container to use [host ports directly](https://docs.docker.com/network/host/) using `--network=host` (also allows achieving better network performance):
+
+```bash
+docker run -d --network=host --name some-clickhouse-server --ulimit nofile=262144:262144 clickhouse/clickhouse-server
```
tianon, line 77:
```suggestion
docker run -d --network=host --name some-clickhouse-server --ulimit nofile=262144:262144 %%IMAGE%%
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+Typically you may want to mount the following folders inside your container to achieve persistency:
+
+-	`/var/lib/clickhouse/` - main folder where ClickHouse stores the data
+-	`/var/log/clickhouse-server/` - logs
+
+```bash
+docker run -d \
+    -v $(realpath ./ch_data):/var/lib/clickhouse/ \
+    -v $(realpath ./ch_logs):/var/log/clickhouse-server/ \
+    --name some-clickhouse-server --ulimit nofile=262144:262144 clickhouse/clickhouse-server
```
tianon, line 94:
```suggestion
    --name some-clickhouse-server --ulimit nofile=262144:262144 %%IMAGE%%
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+
+Typically you may want to mount the following folders inside your container to achieve persistency:
+
+-	`/var/lib/clickhouse/` - main folder where ClickHouse stores the data
+-	`/var/log/clickhouse-server/` - logs
+
+```bash
+docker run -d \
+    -v $(realpath ./ch_data):/var/lib/clickhouse/ \
+    -v $(realpath ./ch_logs):/var/log/clickhouse-server/ \
```
tianon, line 93: It would be a good idea to double quote these, and IMO leaving `realpath` as an exercise to the reader would be fine/sane, but ultimately up-to-you:

```suggestion
    -v "$PWD/ch_data:/var/lib/clickhouse/" \
    -v "$PWD/ch_logs:/var/log/clickhouse-server/" \
```

or at least:

```suggestion
    -v "$(realpath ./ch_data):/var/lib/clickhouse/" \
    -v "$(realpath ./ch_logs):/var/log/clickhouse-server/" \
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+### Linux capabilities
+
+ClickHouse has some advanced functionality, which requires enabling several [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html).
+
+They are optional and can be enabled using the following [docker command-line arguments](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities):
+
+```bash
+docker run -d \
+    --cap-add=SYS_NICE --cap-add=NET_ADMIN --cap-add=IPC_LOCK \
+    --name some-clickhouse-server --ulimit nofile=262144:262144 clickhouse/clickhouse-server
```
tianon, line 112:
```suggestion
    --name some-clickhouse-server --ulimit nofile=262144:262144 %%IMAGE%%
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Linux capabilities
+
+ClickHouse has some advanced functionality, which requires enabling several [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html).
+
+They are optional and can be enabled using the following [docker command-line arguments](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities):
+
+```bash
+docker run -d \
+    --cap-add=SYS_NICE --cap-add=NET_ADMIN --cap-add=IPC_LOCK \
```
tianon, line 111: Given that they're optional, you don't happen to have any documentation about what each of these are used for and when/why users might want to set them (specific to ClickHouse behavior), do you?

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+## Configuration
+
+The container exposes port 8123 for the [HTTP interface](https://clickhouse.com/docs/en/interfaces/http_interface/) and port 9000 for the [native client](https://clickhouse.com/docs/en/interfaces/tcp/).
+
+ClickHouse configuration is represented with a file "config.xml" ([documentation](https://clickhouse.com/docs/en/operations/configuration_files/))
+
+### Start server instance with custom configuration
+
+```bash
+docker run -d --name some-clickhouse-server --ulimit nofile=262144:262144 -v /path/to/your/config.xml:/etc/clickhouse-server/config.xml clickhouse/clickhouse-server
```
tianon, line 124:
```suggestion
docker run -d --name some-clickhouse-server --ulimit nofile=262144:262144 -v /path/to/your/config.xml:/etc/clickhouse-server/config.xml %%IMAGE%%
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+### Start server as custom user
+
+```bash
+# $(pwd)/data/clickhouse should exist and be owned by current user
+docker run --rm --user ${UID}:${GID} --name some-clickhouse-server --ulimit nofile=262144:262144 -v "$(pwd)/logs/clickhouse:/var/log/clickhouse-server" -v "$(pwd)/data/clickhouse:/var/lib/clickhouse" clickhouse/clickhouse-server
```
tianon, line 131: The shell auto-maintains a `$PWD` variable: :eyes:

```suggestion
# $PWD/data/clickhouse should exist and be owned by current user
docker run --rm --user "${UID}:${GID}" --name some-clickhouse-server --ulimit nofile=262144:262144 -v "$PWD/logs/clickhouse:/var/log/clickhouse-server" -v "$PWD/data/clickhouse:/var/lib/clickhouse" %%IMAGE%%
```

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+When you use the image with local directories mounted, you probably want to specify the user to maintain the proper file ownership. Use the `--user` argument and mount `/var/lib/clickhouse` and `/var/log/clickhouse-server` inside the container. Otherwise, the image will complain and not start.
+
+### Start server from root (useful in case of enabled user namespace)
+
+```bash
+docker run --rm -e CLICKHOUSE_UID=0 -e CLICKHOUSE_GID=0 --name clickhouse-server-userns -v "$(pwd)/logs/clickhouse:/var/log/clickhouse-server" -v "$(pwd)/data/clickhouse:/var/lib/clickhouse" clickhouse/clickhouse-server
+```
+
```
tianon, line 141: Should this block be updated in some way, or removed?

clickhouse/content.md
```diff
@@ -0,0 +1,165 @@
...
+### How to create default database and user on starting
+
+Sometimes you may want to create a user (user named `default` is used by default) and database on a container start. You can do it using environment variables `CLICKHOUSE_DB`, `CLICKHOUSE_USER`, `CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT` and `CLICKHOUSE_PASSWORD`:
+
+```bash
+docker run --rm -e CLICKHOUSE_DB=my_database -e CLICKHOUSE_USER=username -e CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT=1 -e CLICKHOUSE_PASSWORD=password -p 9000:9000/tcp clickhouse/clickhouse-server
```
tianon, line 147:
```suggestion
docker run --rm -e CLICKHOUSE_DB=my_database -e CLICKHOUSE_USER=username -e CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT=1 -e CLICKHOUSE_PASSWORD=password -p 9000:9000/tcp %%IMAGE%%
```

tianon (commented):
CI is complaining about `metadata.json`, which is new since you opened this -- I think something like the following is probably the contents you'll want, but if you'd rather just leave that I'm happy to ignore the failing CI and take care of that myself out-of-band (your choice):

```json
{
  "hub": {
    "categories": [
      "databases-and-storage"
    ]
  }
}
```

clickhouse/content.md
Felixoid: Hmm, is this [URL](https://clickhouse.com/docs/knowledgebase/configure_cap_ipc_lock_and_cap_sys_nice_in_docker) worth adding here?

clickhouse/content.md
Felixoid: Updated, for sure.

Felixoid (commented):
For the rest, I'll try to build a limited version of your `update.sh` in our repo to have compatibility between repos. Will be back in a moment

clickhouse/content.md
Felixoid: Getting rid of `--link` would require some good replacement, and off the top of my head, the docker network juggling doesn't give an easy-to-use alternative.

Can you suggest something here?

clickhouse/content.md
Felixoid: The line is added below:

`Read more in [knowledge base](https://clickhouse.com/docs/knowledgebase/configure_cap_ipc_lock_and_cap_sys_nice_in_docker).`

Felixoid (commented):
The `--link` is only question, I think

clickhouse/content.md
tianon: https://github.com/docker-library/docs/pull/1441 was the best I could come up with (without having a common blurb for every image trying to explain Docker Networks), but it's very "go read the manual" so I'm not super happy about it either (I really like the `--link` feature, fwiw :sob:)

If you'd rather keep `--link` that is OK :+1:

clickhouse/content.md
tianon: Yeah, for sure!  I wonder if it could be improved to explain _why_, but that's not a blocker at all, just something I think would be useful. :smile: :heart:

clickhouse/content.md
```diff
@@ -0,0 +1,160 @@
...
+
+For more information and documentation see https://clickhouse.com/.
+
+### Compatibility
+
+-	The amd64 image requires support for [SSE3 instructions](https://en.wikipedia.org/wiki/SSE3). Virtually all x86 CPUs after 2005 support SSE3.
+-	The arm64 image requires support for the [ARMv8.2-A architecture](https://en.wikipedia.org/wiki/AArch64#ARMv8.2-A) and additionally the Load-Acquire RCpc register. The register is optional in version ARMv8.2-A and mandatory in [ARMv8.3-A](https://en.wikipedia.org/wiki/AArch64#ARMv8.3-A). Supported in Graviton >=2, Azure and GCP instances. Examples for unsupported devices are Raspberry Pi 4 (ARMv8.0-A) and Jetson AGX Xavier/Orin (ARMv8.2-A).
+-	Since the Clickhouse 24.11 Ubuntu images started using `ubuntu:22.04` as its base image. It requires docker version >= `20.10.10` containing [patch](https://github.com/moby/moby/commit/977283509f75303bc6612665a04abf76ff1d2468). As a workaround you could use `docker run [--privileged | --security-opt seccomp=unconfined]` instead, however that has security implications.
```
tianon, line 17: This note is fine to leave in, but suggesting `--privileged` is a no-go -- we have/maintain https://github.com/docker-library/official-images/issues/16829 as a place to link folks for these types of issues, if you wanted to include that, but that's not necessary (only removing `--privileged`, because it's a very large hammer with much larger security implications, and should really only be used for a very limited set of use cases :sweat_smile:).

clickhouse/content.md
```diff
@@ -0,0 +1,160 @@
...
+# OR
+docker exec -it some-clickhouse-server clickhouse-client
+```
+
+More information about the [ClickHouse client](https://clickhouse.com/docs/en/interfaces/cli/).
+
+### connect to it using curl
+
+```bash
+echo "SELECT 'Hello, ClickHouse!'" | docker run -i --rm --link some-clickhouse-server:clickhouse-server buildpack-deps:curl 'http://clickhouse-server:8123/?query=' -s --data-binary @-
```
tianon, line 44:
```suggestion
echo "SELECT 'Hello, ClickHouse!'" | docker run -i --rm --link some-clickhouse-server:clickhouse-server buildpack-deps:curl curl 'http://clickhouse-server:8123/?query=' -s --data-binary @-
```

(sorry for the confusion; that's on me -- this image doesn't default to curl, but does have it installed, so command is still needed :sweat_smile:)

clickhouse/github-repo
```diff
@@ -0,0 +1 @@
+https://github.com/ClickHouse/docker-library
```
tianon, line 1: Mostly an FYI, this will get used for things like links in the "Quick Reference" section to GitHub issues -- is this where you want users to file issues related to the image, or would you rather they go to ClickHouse/ClickHouse?  (do you plan to keep issues enabled on the docker-library repository?)

tianon (commented):
Only a couple really minor things :eyes:

clickhouse/content.md
Felixoid: I'm 100% fine getting rid of `--privileged`. I agree that it's a bad practice and an overhead here.

clickhouse/github-repo
Felixoid: I misinterpret [the readme](https://github.com/docker-library/docs?tab=readme-ov-file#github-repo), it should be the `ClickHouse/ClickHouse` for sure

clickhouse/content.md
Felixoid: Thanks, will change it in a moment

clickhouse/content.md
Felixoid: There's a part of the ClickHouse logs related to the capabilities on the KB page:

`2023.04.19 08:04:10.022720 [ 1 ] {} <Information> Application: It looks like the process has no CAP_IPC_LOCK capability, binary mlock will be disabled. It could happen due to incorrect ClickHouse package installation. You could resolve the problem manually with 'sudo setcap cap_ipc_lock=+ep /usr/bin/clickhouse'. Note that it will not work on 'nosuid' mounted filesystems.`

`2023.04.19 08:04:10.065860 [ 1 ] {} <Information> Application: It looks like the process has no CAP_SYS_NICE capability, the setting 'os_thread_priority' will have no effect. It could happen due to incorrect ClickHouse package installation. You could resolve the problem manually with 'sudo setcap cap_sys_nice=+ep /usr/bin/clickhouse'. Note that it will not work on 'nosuid' mounted filesystems.`

Should I add it as well explicitly?

tianon: Naw, I think what you've got now is fine. :+1:

tianon (approved):

Felixoid:
Oh, my chaos, my chaos, my chaos! Did we finish it? What's next?

I am so excited ^_^

tianon:
(as noted in https://github.com/docker-library/official-images/pull/15846#issuecomment-2465782436, just waiting for @yosifkit to give me another set of eyes :crossed_fingers:)

yosifkit (approved):

Felixoid:
@tianon "funny" aftermath. I haven't found that the 24.10 is built on top of jammy, AKA 22.04, so the https://github.com/docker-library/docs/pull/2397/files#diff-6a2321fc3b812fb166b7184bef1475db0c18919ad2612b02e06429f4cc986fc4R17 is not valid. I'll fix it in the follow-up PR to the library, where I will make this version `FROM ubuntu:20.04`.

PR #2399, opened 2023-12-20
mongo: add security section

whalelines:
Closes docker-library/mongo#656.

### DIFF
--- a/mongo/content.md
+++ b/mongo/content.md
@@ -8,6 +8,10 @@ First developed by the software company 10gen (now MongoDB Inc.) in October 2007
 
 %%LOGO%%
 
+# Security
+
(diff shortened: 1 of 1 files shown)


whalelines:
By my calculations, this does not cause the README without the tags to exceed the Hub size limit.

tianon (approved):

tianon:
Nice, thank you!

godmar:
While a good start, that section is insufficient because it does not address the behavior on `-p hostport:containerport`. I suggest looking at the Redis documentation for an example to follow.

The pitfall here is not just that you ship with authentication disabled: it's that the common defense mechanism of binding to the loopback interface does not work inside a container.  As soon as you create a network path to the container -- which you have to do to be able to use your software at all - you run the risk of exposing the database to the Internet.

PR #2406, opened 2024-01-09
Fix broken links for mysql

masonbrothers:
The link http://dev.mysql.com/doc/en/adding-users.html appears to be broken.

I think the correct link should be https://dev.mysql.com/doc/refman/en/creating-accounts.html

### DIFF
--- a/mysql/content.md
+++ b/mysql/content.md
@@ -92,7 +92,7 @@ This variable is mandatory and specifies the password that will be set for the M
 
 ### `MYSQL_DATABASE`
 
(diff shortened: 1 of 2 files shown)


tianon:
Nice catch, thanks!  I guess we could also pretty reasonably link to the `GRANT` docs or to https://dev.mysql.com/doc/refman/8.2/en/privileges-provided.html#priv_all for `ALL` specifically, but this doc links to both of those eventually if someone digs in (and matches what we linked to previously), so I think we're good. :+1:

PR #2411, opened 2024-01-28
nginx: Update command to generate host config file

JohnCMoon:
Currently, instructions to copy the default configuration from the nginx container to the host uses three commands. This can be done more efficiently with one command (assuming the presence of "cat" in the container which should be a safe).

Update the instructions to use the single command.

### DIFF
--- a/nginx/content.md
+++ b/nginx/content.md
@@ -42,9 +42,7 @@ You can mount your configuration file, or build a new image with it.
 If you wish to adapt the default configuration, use something like the following to get it from a running nginx container:
 
 ```console
-$ docker run --name tmp-nginx-container -d %%IMAGE%%
-$ docker cp tmp-nginx-container:/etc/nginx/nginx.conf /host/path/nginx.conf
-$ docker rm -f tmp-nginx-container
+$ docker run --rm --entrypoint=cat nginx /etc/nginx/nginx.conf > /host/path/nginx.conf
 ```
 
 And then edit `/host/path/nginx.conf` in your host file system.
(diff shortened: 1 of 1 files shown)


tianon:
cc `nginx` image maintainer @thresheek (and this would need to be moved over to `content.md` as `README.md` is generated)

thresheek:
Thanks @JohnCMoon and @tianon !

I think the change looks good, so please fix the issues mentioned and we'll be good to go.

JohnCMoon:
Moved change to `content.md` - sorry I missed the "do not edit" header at the top of `README.md`!

nginx/content.md
```diff
@@ -42,9 +42,7 @@ You can mount your configuration file, or build a new image with it.
 If you wish to adapt the default configuration, use something like the following to get it from a running nginx container:
 
 ```console
-$ docker run --name tmp-nginx-container -d %%IMAGE%%
-$ docker cp tmp-nginx-container:/etc/nginx/nginx.conf /host/path/nginx.conf
-$ docker rm -f tmp-nginx-container
+$ docker run --rm --entrypoint=cat nginx /etc/nginx/nginx.conf > /host/path/nginx.conf
```
tianon, line 45: Doh sorry, didn't notice sooner -- this should be `%%IMAGE%%` so the templating uses the correct version on the per-arch pages too:

```suggestion
$ docker run --rm --entrypoint=cat %%IMAGE%% /etc/nginx/nginx.conf > /host/path/nginx.conf
```

tianon: (I'm happy to make this change for you if you'd prefer)

JohnCMoon: Good catch!

tianon:
Thank you!

PR #2416, opened 2024-02-08
postgres: update README.md

dbulashev:
* added example for docker stack deploy (shared memory size)
  * fix incorrect link

### DIFF
--- a/postgres/stack.yml
+++ b/postgres/stack.yml
@@ -1,11 +1,19 @@
 # Use postgres/example user/password credentials
-version: '3.1'
+version: '3.9'
 
 services:
 
   db:
     image: postgres
     restart: always
+    # set shared memory limit when using docker-compose
+    shm_size: 128mb
+    # or set shared memory limit when deploy via swarm stack
+    #volumes:
+    #  - type: tmpfs
+    #    target: /dev/shm
+    #    tmpfs:
+    #      size: 134217728 # 128*2^20 bytes = 128Mb
     environment:
       POSTGRES_PASSWORD: example
 
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -243,7 +243,7 @@ The three easiest ways to get around this:
 
 If there is no database when `postgres` starts in a container, then `postgres` will create the default database for you. While this is the expected behavior of `postgres`, this means that it will not accept incoming connections during that time. This may cause issues when using automation tools, such as `docker-compose`, that start several containers simultaneously.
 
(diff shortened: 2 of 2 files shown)


postgres/stack.yml
```diff
@@ -1,11 +1,19 @@
...
     image: postgres
     restart: always
+    # set shared memory limit when deploy swarm stack
+    tmpfs:
+      - /tmp:size=128mb
+    volumes:
+      - type: tmpfs
+        target: /dev/shm
+    # set shared memory limit when deploy through docker-compose
+    # shm_size: 128mb
```
yosifkit, line 16: We should swap the default to `shm_size` and then comment for the swarm stack version since we plan to rename them all back to `docker-compose.yml` at some point.

```suggestion
    # set shared memory limit when using docker-compose
    shm_size: 128mb
    # or set shared memory limit when deploy via swarm stack
    #tmpfs:
    #  - /dev/shm:size=128m
```

postgres/stack.yml
```diff
@@ -1,11 +1,19 @@
...
     image: postgres
     restart: always
+    # set shared memory limit when using docker-compose
+    shm_size: 128mb
+    # or set shared memory limit when deploy via swarm stack
+    #tmpfs:
+    # - /tmp:size=128mb
+    #volumes:
+    # - type: tmpfs
+    #   target: /dev/shm
```
tianon, line 16: This should be a bit less: :eyes:

```suggestion
    #tmpfs:
    #  - /dev/shm:size=128m
```

(the `128mb` with a `b` suffix causes errors, and if you want that to apply to `/dev/shm`, it has to be on `/dev/shm`, not just `/tmp` :sweat_smile:)

dbulashev: @tianon, thanks for the good comments.  /tmp target was an unfortunate typo.
btw, limitations in short syntax (/dev/shm:size=128m) does not work ([compose-spec/docs#tmpfs-long-syntax](https://github.com/compose-spec/compose-spec/blob/master/05-services.md#long-syntax-5)).

I checked that the `b` suffix does not cause errors. Specifying units of measurement for megabytes can be `m` or `mb`
([compose-spec/docs#specifying-byte-values](https://github.com/compose-spec/compose-spec/blob/master/spec.md#specifying-byte-values)).

tianon: So glad we have a second spec that's confusingly incompatible/inconsistent with the `docker run` flags it's wrapping :sweat_smile: :sob: :heart:

tianon (approved):
Thank you!

PR #2418, opened 2024-02-11
Corrected info how WORDPRESS_CONFIG_EXTRA is put into effect

MorgusLethe:
The previous wording is incorrect. The value is not embedded verbatim, but is evaluated by php's eval() function:
```if ($configExtra = getenv_docker('WORDPRESS_CONFIG_EXTRA', '')) {
        eval($configExtra);
}```


tianon:
Nice, thank you!  Just for posterity, this is a result of the discussion in https://github.com/docker-library/wordpress/issues/771#issuecomment-1896002916 (behavior was changed in https://github.com/docker-library/wordpress/pull/557). :heart:

PR #2420, opened 2024-02-21
python: Add note on `pip install` in slim variants

fred-from-york:
Add an image-specific slim variant section for Python. This overrides the default description of a slim variant.

The image-specific description clarifies when `pip install` might fail in slim variants of the Python image.


tianon (approved):
Nice, thank you!

PR #2431, opened 2024-03-26
Reset mongo express basic auth

pothitos:
ME_CONFIG_BASICAUTH=false is no longer the default value following https://github.com/mongo-express/mongo-express-docker/commit/ec6ee123c1ecd59a7662680bec407e0913a0321e#diff-8d7a21b017921bb88eaf71656b7b5767203db16e8126fc1e5ad2a9ba0bc542f5R29

If we don't set it to `false`, mongo express will prompt for a basic auth username and password. But such credentials don't exist in `stack.yml`.


tianon:
cc @BlackthornYugen

yosifkit (approved):

PR #2433, opened 2024-03-29
Update Nextcloud

J0WI:


nextcloud/content.md
```diff
@@ -79,14 +81,23 @@ If you want to use named volumes for all of these, it would look like this:
...
 ```
 
+### Custom volumes
+
+If mounting additional volumes under `/var/www/html`, you should consider:
+
+-	Confirming that [upgrade.exclude](https://github.com/nextcloud/docker/blob/master/upgrade.exclude) contains the files and folders that should persist during installation and upgrades; or
+-	Mounting storage volumes to locations outside of `/var/www/html`.
+
+> [!WARNING] You should note that data inside the main folder (`/var/www/html`) will be overridden/removed during installation and upgrades, unless listed in [upgrade.exclude](https://github.com/nextcloud/docker/blob/master/upgrade.exclude). The additional volumes officially supported are already in that list, but custom volumes will need to be added by you. We suggest mounting custom storage volumes outside of `/var/www/html` and if possible read-only so that making this adjustment is unnecessary. If you must do so, however, you may build a custom image with a modified `/upgrade.exclude` file that incorporates your custom volume(s).
```
tianon, line 99: `> [!WARNING] ...` is a github-specific markdownism that probably will not translate well to Docker Hub

nextcloud/content.md
```diff
@@ -280,7 +326,7 @@ Then run `docker-compose up -d`, now you can access Nextcloud at http://localhos
...
+When using the FPM image, you need another container that acts as web server on port 80 and proxies the requests to the Nextcloud container. In this example a simple nginx container is combined with the Nextcloud-fpm image and a MariaDB database container. The data is stored in docker volumes. The nginx container also needs access to static files from your Nextcloud installation. It gets access to all the volumes mounted to Nextcloud via the `volumes_from` option.The configuration for nginx is stored in the configuration file `nginx.conf`, that is mounted into the container. An example can be found in the examples section [here](https://github.com/nextcloud/docker/tree/master/.examples).
```
tianon, line 329: I think this change was unintentional (`option.The configuration` probably wants a space between `.` and `The` :smile:)

nextcloud/content.md
```diff
@@ -499,99 +545,65 @@ The `--pull` option tells docker to look for new versions of the base image. The
...
+	-	To import from a MySQL dump use the following commands`console
+		docker cp ./database.dmp nextcloud_db_1:/dmp
+		docker-compose exec db sh -c "mysql --user USER --password PASSWORD nextcloud < /dmp"
+		docker-compose exec db rm /dmp
+		`
+	-	To import from a PostgreSQL dump use to following commands`console
+		docker cp ./database.dmp nextcloud_db_1:/dmp
+		docker-compose exec db sh -c "psql -U USER --set ON_ERROR_STOP=on nextcloud < /dmp"
+		docker-compose exec db rm /dmp
+		`
```
tianon, line 558: This definitely isn't right :sweat_smile:  I think maybe something like this?

```suggestion
	-	To import from a MySQL dump use the following commands

		```console
		docker cp ./database.dmp nextcloud_db_1:/dmp
		docker-compose exec db sh -c "mysql --user USER --password PASSWORD nextcloud < /dmp"
		docker-compose exec db rm /dmp
		```

	-	To import from a PostgreSQL dump use to following commands

		```console
		docker cp ./database.dmp nextcloud_db_1:/dmp
		docker-compose exec db sh -c "psql -U USER --set ON_ERROR_STOP=on nextcloud < /dmp"
		docker-compose exec db rm /dmp
		```

```

nextcloud/content.md
```diff
@@ -499,99 +545,65 @@ The `--pull` option tells docker to look for new versions of the base image. The
...
-		-	In case of PostgreSQL database
-
-		```php
-		'dbhost' => 'db:5432',
-		```
-
-	2.	Make sure you have no configuration for the `apps_paths`. Delete lines like these
-
-		```php
+		-	In case of MySQL database`php
```
tianon, line 561: Again, something seems to have mangled the formatting of these code blocks :sob:

nextcloud/content.md
```diff
@@ -123,7 +134,7 @@ The Nextcloud image supports auto configuration via environment variables. You c
 -	`POSTGRES_PASSWORD` Password for the database user using postgres.
 -	`POSTGRES_HOST` Hostname of the database server using postgres.
 
-As an alternative to passing sensitive information via environment variables, `_FILE` may be appended to the previously listed environment variables, causing the initialization script to load the values for those variables from files present in the container. See [Docker secrets](#docker=secrets) section below.
+As an alternative to passing sensitive information via environment variables, `_FILE` may be appended to the previously listed environment variables, causing the initialization script to load the values for those variables from files present in the container. See [Docker secrets](#docker-secrets) section below.
```
tianon, line 137: FWIW, I don't think this link will work correctly on Docker Hub either way :see_no_evil: (see https://github.com/docker/hub-feedback/issues/956)

tianon (approved):

PR #2434, opened 2024-03-29
Update Matomo

J0WI:


matomo/content.md
```diff
@@ -50,25 +50,24 @@ The following environment variables are also honored for configuring your Matomo
...
 -	`MATOMO_DATABASE_DBNAME`
 
+The PHP memory limit can be configured with the following environment variable:
+
+-	`PHP_MEMORY_LIMIT`
+
 ## Docker-compose examples and log import instructions
 
-A minimal set-up using docker-compose is available in the [.examples folder](https://github.com/matomo-org/docker/tree/master/.examples).
+A minimal set-up using docker-compose is available in the [.examples folder](https://github.com/matomo-org/docker/tree/master/.examples), a more complete [example can be found at IndieHosters/matomo](https://github.com/libresh/compose-matomo/blob/master/docker-compose.yml).
```
tianon, line 59: This new example seems to be 6 years old -- is it definitely still accurate and more complete than what's in the matomo org?

matomo/content.md
```diff
@@ -50,25 +50,24 @@ The following environment variables are also honored for configuring your Matomo
...
 
 If you want to use the import logs script, you can then run the following container as needed, in order to execute the python import logs script:
 
-```console
-docker run --rm --volumes-from="matomo_app_1" --link matomo_app_1 python:2-alpine python /var/www/html/misc/log-analytics/import_logs.py --url=http://ip.of.your.piwik --login=yourlogin --password=yourpassword --idsite=1 --recorders=4 /var/www/html/logs/access.log
-```
+	docker run --rm --volumes-from="matomo-app-1" --link matomo-app-1 python:3-alpine python /var/www/html/misc/log-analytics/import_logs.py --url=http://ip.of.your.matomo.example --login=yourlogin --password=yourpassword --idsite=1 --recorders=4 /var/www/html/logs/access.log
```
tianon, line 63: Shouldn't this just swap to `bash` instead of `console` like the one above? (instead of dropping highlighting entirely)

matomo/content.md
```diff
@@ -50,25 +50,24 @@ The following environment variables are also honored for configuring your Matomo
...
 
 ## Contribute
 
 Pull requests are very welcome!
 
-We'd love to hear your feedback and suggestions in the issue tracker: [github.com/motomo-org/docker/issues](https://github.com/matomo-org/docker/issues).
+We'd love to hear your feedback and suggestions in the issue tracker: [github.com/matomo-org/docker/issues](https://github.com/matomo-org/docker/issues).
```
tianon, line 69: Oops! :see_no_evil:

We have both `%%ISSUES%%` and `%%GITHUB-REPO%%` if you think one of those would be helpful/useful here (but not terribly important).

PR #2446, opened 2024-05-10
Add Docker Hub categories

yosifkit:
Docker Hub allows owners to set categories for their repos (https://docs.docker.com/docker-hub/repos/categories/), so let's allow our image maintainers in DOI to do it. Categories are not yet visible to users.

To our image maintainers: these initial categories are based on some rough google searching of the contained software and `content.md`, we apologize for any that are inaccurate. Docker Hub currently allows setting up to 3, so if you'd like to add more, just add them to the list in `[repo]/metadata.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ This file should contain a link to the maintainers of the Dockerfile.
 
 ## `metadata.json`
 
-This file contains data about the repo for Docker Hub. The minimum file is defined below. `./metadata.sh [repo-name]` must be used to correctly format it (use `-w` to apply its suggested format changes). Only three sorted unique Docker Hub categories are allowed. `metadata.json` in the root contains the list of categories to choose from.
+This file contains data about the repo for Docker Hub. The minimum file is defined below. `./metadata.sh [repo-name]` must be used to correctly format it (use `-w` to apply its suggested format changes). Only three sorted unique Docker Hub categories are allowed. `metadata.json` in the root contains the list of categories to choose from. See descriptions for the categories on the [Docker docs site](https://docs.docker.com/docker-hub/repos/categories/).
 
 ```json
 {
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ This file should contain a link to the maintainers of the Dockerfile.
 
 ## `metadata.json`
 
-This file contains data about the repo for Docker Hub. The minimum file is defined below. `./metadata.sh [repo-name]` must be used to correctly format it (use `-f` to apply its suggested changes). There is a limit to the number of Docker Hub categories allowed; run with `-c` to check the limit and categories. `metadata.json` in the root contains the list of categories to choose from.
+This file contains data about the repo for Docker Hub. The minimum file is defined below. `./metadata.sh [repo-name]` must be used to correctly format it (use `-w` to apply its suggested format changes). Only three sorted unique Docker Hub categories are allowed. `metadata.json` in the root contains the list of categories to choose from.
 
 ```json
 {
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ After opening your Pull Request the changes will be checked by an automated `mar
 -	Create a `license.md` (required)
 -	Create a `maintainer.md` (required)
 -	Create a `github-repo` (required)
+-	Create a `metadata.json` (required)
 -	Add a `logo.png` (recommended)
 
 Optionally:
@@ -127,6 +128,18 @@ The image is automatically scaled to a 120 pixel square for the top of the Docke
 ## `maintainer.md`
 
 This file should contain a link to the maintainers of the Dockerfile.
+
+## `metadata.json`
+
+This file contains data about the repo for Docker Hub. The minimum file is defined below. `./metadata.sh [repo-name]` must be used to correctly format it (use `-f` to apply its suggested changes). There is a limit to the number of Docker Hub categories allowed; run with `-c` to check the limit and categories. `metadata.json` in the root contains the list of categories to choose from.
+
+```json
+{
+    "hub": {
+         "categories": []
+    }
+}
+```
 
 ## `README-short.txt`
 
--- a/busybox/metadata.json
+++ b/busybox/metadata.json
@@ -1,0 +1,7 @@
+{
+  "hub": {
+    "categories": [
+      "operating-systems"
+    ]
+  }
+}
--- a/metadata.sh
+++ b/metadata.sh
@@ -14,38 +14,31 @@ self="$(basename "$0")"
 usage() {
 	cat <<EOUSAGE
 
-usage: $self [--categories] [--diff] [--write] repo [...]
+usage: $self [--write] repo [...]
    ie: $self debian
-       $self -d */
-       $self -dc python
+       $self -w python
 
-This script checks a repo's metadata.json and checks categories, diffs, or formats it.
+This script checks a givens repo's metadata.json. It checks formating (providing a diff), checks categories, and can write the formatting changes.
 
--c, --categories    Check that the categories are from the valid set of categories and that there are no more than $maxCategories. Exits non-zero if there are too many categories or they are unkown.
--d, --diff          Check formatting of the '[repo]/metadata.json' and print a diff. Default action if no flags are supplied. Exits non-zero if there is a difference.
 -h, --help          Print this help output and exit.
--w, --write         Apply the formatting that the '-d' flag would output.
+-w, --write         Apply json formatting (run without to see the diff that would be applied).
 
-Arguments are the list of repos with a 'metadata.json' in them. 'metadata.json' is expected in every repo
+Arguments are the list of repos with a 'metadata.json' in them. 'metadata.json' is expected in every repo.
 '.' can also be passed to check the format of the canonical './metadata.json' at
 the root of the repo, but the max categories of '-c' is skipped for it.
 EOUSAGE
 }
 
 # arg handling
-opts="$(getopt -o 'cdhw' --long 'categories,diff,help,write' -- "$@" || { usage >&2 && false; })"
+opts="$(getopt -o 'hw' --long 'help,write' -- "$@" || { usage >&2 && false; })"
 eval set -- "$opts"
 
-categories=
-diff=
 write=
 
 while true; do
 	flag="$1"
 	shift
 	case "$flag" in
-		--categories | -c) categories=1 ;;
-		--diff | -d) diff=1 ;;
 		--help | -h) usage && exit 0 ;;
 		--write | -w) write=1 ;;
 		--) break ;;
@@ -59,11 +52,6 @@ while true; do
 	esac
 done
 
-# default to print diff
-if [ -z "$diff" ] && [ -z "$categories" ] && [ -z "$write" ]; then
-	diff=1
-fi
-
 repos=( "$@" )
 if [ "${#repos[@]}" -eq 0 ]; then
 	repos=( */ )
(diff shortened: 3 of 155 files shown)


yosifkit:
For now, `get-categories.sh` will be manually run since the available categories shouldn't change that often.

busybox/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "hub": {
+    "categories": [
+      "operating-systems"
```
LaurentGoderre, line 4: Maybe add `"developer-tools"` as well here

mongo-express/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "hub": {
+    "categories": [
+      "databases-and-storage"
```
LaurentGoderre, line 4: Maybe `"developer-tools"` also applies here

php-zendserver/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "hub": {
+    "categories": [
+      "languages-and-frameworks"
```
LaurentGoderre, line 4: This should also have `"web-servers"`

phpmyadmin/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "hub": {
+    "categories": [
+      "databases-and-storage"
```
LaurentGoderre, line 4: Maybe "developer-tools" like for mongo express

tianon (commented):
I only reviewed the implementation for now (I have not re-reviewed the categories yet like I plan to), and have a few thoughts/suggestions.

I'm definitely really concerned about having canonical human-maintained JSON as an input (since JSON is not a terribly human or diff friendly format), but it's probably fine for now and I think there are future benefits to having this as JSON. :+1:

metadata.sh
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+workdir="$(readlink -f "$BASH_SOURCE")"
+workdir="$(dirname "$workdir")"
+#cd "$workdir"
```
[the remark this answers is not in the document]
yosifkit: That was me debating whether to just `cd` or use the value where necessary. I went for the latter but not opposed to the former. It's only directly used in two places.

metadata.sh
```diff
@@ -0,0 +1,140 @@
...
+			} >&2
+			exit 1
+			;;
+	esac
+done
+
+# default to print diff
+if [ -z "$diff" ] && [ -z "$categories" ] && [ -z "$fmt" ]; then
+	diff=1
+fi
```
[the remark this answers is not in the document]
yosifkit: One possibility is to drop them into an associative array (for easy lookup later) and check its length.

```sh
declare -A checks= ()
case ...
   checks[diff]=1
...
if [ "${#checks[@]}" -eq 0 ]; then
```

metadata.sh
```diff
@@ -0,0 +1,140 @@
...
+
+repos=( "$@" )
+if [ "${#repos[@]}" -eq 0 ]; then
+	repos=( */ )
+fi
+repos=( "${repos[@]%/}" )
+
+canonicalMetadataFileJson="$(cat "$canonicalMetadataFile")"
+export canonicalMetadataFileJson
```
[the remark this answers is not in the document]
yosifkit: I thought slurpfile would be slower, since we'd have to open/read the canonical file for every repo, but seems negligible locally, so I'm fine with it.

metadata.sh
```diff
@@ -0,0 +1,140 @@
...
+
+failure=
+for repo in "${repos[@]}"; do
+	repoFile="$workdir/$repo/$jsonFile"
+	if [ ! -f "$repoFile" ]; then
+		failure=1
+		echo "error file does not exist: $repo/$jsonFile"
+		continue
+	fi
+	repoFile="$(readlink -f "$repoFile")"
```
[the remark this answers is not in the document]
yosifkit: Moving to just drop the `readlink` here. With a `cd` at the start of the file, we can use `repoFile="$repo/$jsonFile"` and just let the first `jq` fail if it doesn't exist. I don't think we need to care if `$repo/$jsonFile` has a symlink.

Could alternatively keep the original with `-e` instead.

busybox/metadata.json
yosifkit: We should fix ones that are incorrect in this PR, but I think extra categories should be a separate PR to prevent this one from ballooning.

whalelines: And it doesn't hurt to give image maintainers an opportunity to provide input.

metadata.sh
```diff
@@ -0,0 +1,140 @@
...
+			fi
+			if [ -n "$fmt" ]; then
+				# TODO should fmt + categories remove invalid or categories over max?
+				echo "$repoFilejson" > "$repoFile"
+			fi
+		fi
+	fi
+
+	# TODO also check for required keys and/or types?
+	if [ -n "$categories" ]; then
```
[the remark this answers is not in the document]
yosifkit: 🤔 I was envisioning two use cases for the script: CI to verify that the json files are formatted correctly and not invalid (diff with categories check); and the contributor running it interactively to see a diff or write their file. I wasn't sure if the interactive user would want to see the verification check when they do `-w`, for example. I'm open to removing or renaming more of the flags.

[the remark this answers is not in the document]
tianon: Yeah, that makes sense -- I think what I'm struggling with is that I don't see a strong use case for wanting to format the file but leave invalid data in it, and if we change the script to always check, then the script itself gets simpler (less variance to code and maintain and think about) _and_ users are more likely to get earlier feedback that their file is invalid than if they wait all the way to CI.

We could make the validation happen _after_ the formatting/diffing such that users _can_ combine "write the formatted result" and keep their invalid data, but I don't feel strongly about that (whatever makes it easier to write this such that it's performant and straightforward to maintain :eyes:).

README.md
```diff
@@ -128,6 +129,18 @@ The image is automatically scaled to a 120 pixel square for the top of the Docke
 
 This file should contain a link to the maintainers of the Dockerfile.
 
+## `metadata.json`
+
+This file contains data about the repo for Docker Hub. The minimum file is defined below. `./metadata.sh [repo-name]` must be used to correctly format it (use `-f` to apply its suggested changes). There is a limit to the number of Docker Hub categories allowed; run with `-c` to check the limit and categories. `metadata.json` in the root contains the list of categories to choose from.
```
whalelines, line 134: Should we provide example usages?

```sh
$ ./metadata.sh -f -c REPO_NAME
```

push.pl
```diff
@@ -211,6 +212,35 @@ sub prompt_for_edit {
...
+				last;
+			}
+		}
+	}
+	if ($needCat) {
+		say 'updating ' . $repoName . ' categories';
+		my $catsPatch = $ua->patch($repoUrl . 'categories/' => { %$authorizationHeader, Accept => 'application/json' } => json => [
+				map { {
+					slug => $_,
+					name => 'All those moments will be lost in time, like tears in rain... Time to die.',
```
whalelines, line 239: A reminder that the `name` property is not used

metadata.sh
```diff
@@ -0,0 +1,140 @@
...
+   ie: $self debian
+       $self -d */
+       $self -dc python
+
+This script checks a repo's metadata.json and checks categories, diffs, or formats it.
+
+-c, --categories    Check that the categories are from the valid set of categories and that there are no more than $maxCategories. Exits non-zero if there are too many categories or they are unkown.
+-d, --diff          Check formatting of the '[repo]/metadata.json' and print a diff. Default action if no flags are supplied. Exits non-zero if there is a difference.
+-h, --help          Print this help output and exit.
+-w, --write         Apply the formatting that the '-d' flag would output.
```
whalelines, line 27: There is a `-f` option, right?

metadata.sh
```diff
@@ -0,0 +1,140 @@
...
+
+# arg handling
+opts="$(getopt -o 'cdhw' --long 'categories,diff,help,write' -- "$@" || { usage >&2 && false; })"
+eval set -- "$opts"
+
+categories=
+diff=
+write=
+
+while true; do
```
whalelines, line 43:
```suggestion
while :; do
```

metadata.sh
```diff
@@ -0,0 +1,140 @@
...
+-w, --write         Apply the formatting that the '-d' flag would output.
+
+Arguments are the list of repos with a 'metadata.json' in them. 'metadata.json' is expected in every repo
+'.' can also be passed to check the format of the canonical './metadata.json' at
+the root of the repo, but the max categories of '-c' is skipped for it.
+EOUSAGE
+}
+
+# arg handling
+opts="$(getopt -o 'cdhw' --long 'categories,diff,help,write' -- "$@" || { usage >&2 && false; })"
```
whalelines, line 36: I guess there is no `-f` option. Perhaps the README needs updated?

whalelines (commented):
Still working through metadata.sh, but submitting this so it does not get lost

README.md
yosifkit: `-f` was changed to `-w` and now I dropped the `-c` and `-d` flags (commit incoming with script and README adjustments).

push.pl
yosifkit: Yeah, it is required to be non-empty, but is unused as far as I can tell.

metadata.sh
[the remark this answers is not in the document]
yosifkit: Diff and categories flag have been removed; now just `-w` to write the jq formatting.

metadata.sh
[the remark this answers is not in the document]
yosifkit: Simplified in https://github.com/docker-library/docs/commit/af088807dd6d7f4d98aeaa7b365873bf73dc9f23; no`-d` or `-c` flags

metadata.sh
```diff
@@ -135,6 +104,20 @@ for repo in "${repos[@]}"; do
 			continue
 		fi
 	fi
+
+	# check for categories that aren't in the canonical set
```
tianon, line 108: This should be part of the `if` statement just above right? (so we're not checking `canonicalMetadataFile` against itself?)

Could probably then collapse the two `jq` invocations into one, but getting decent output out of that might be annoying?  Maybe have `jq` generate the error message string contents instead?

yosifkit:
Before and after most recent commit unifying the jq expression for categories (with some local broken files):
```console
$ time ./metadata.sh */ .
error: test/metadata.json does not exist or is empty
--- unit/metadata.json  2024-05-16 15:50:54.308055944 -0700
+++ /dev/fd/63  2024-05-16 15:50:57.018164177 -0700
@@ -1,9 +1,6 @@
 {
   "hub": {
     "categories": [
-      "web-servers",
-      "web-servers",
-      "web-servers",
       "servers",
       "web-servers"
     ]
error: unit/metadata.json: too many categories: 5 (max 3)
error: unit/metadata.json: unkown categories: ["servers"]

real    0m6.432s
user    0m6.110s
sys     0m0.258s
```
```console
$ time ./metadata.sh */ .
error: test/metadata.json does not exist or is empty
--- unit/metadata.json  2024-05-16 15:50:54.308055944 -0700
+++ /dev/fd/63  2024-05-16 15:51:15.227048108 -0700
@@ -1,9 +1,6 @@
 {
   "hub": {
     "categories": [
-      "web-servers",
-      "web-servers",
-      "web-servers",
       "servers",
       "web-servers"
     ]
error: unit/metadata.json: too many categories: 5 (max 3)
error: unit/metadata.json: unknown categories ["servers"]

real    0m4.327s
user    0m4.145s
sys     0m0.162s
```

zookeeper/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "hub": {
+    "categories": [
+      "message-queues"
```
tianon, line 4:
```suggestion
      "databases-and-storage"
```

(it takes a bit to unpack the tech lingo on many official zookeeper descriptions, but it's a distributed key-value store, _not_ a message queue :eyes:)

tianon (commented):
A few category adjustments (some wrong, some with a better match).

FWIW, I used `for r in */; do printf '%s\n%s\n%s\n\n' "${r%/}" "$(jq -r '.hub.categories | join(" ")' "$r/metadata.json")" "$(< "$r/README-short.txt")"; done` in a local checkout to review these with slightly more context. :smile:

metadata.sh
```diff
@@ -0,0 +1,118 @@
...
+canonicalMetadataFile="./$jsonFile"
+export maxCategories=3
+
+self="$(basename "$0")"
+
+usage() {
+	cat <<EOUSAGE
+
+usage: $self [--write] repo [...]
+   ie: $self debian
```
whalelines, line 18:
```suggestion
   eg: $self debian
```

metadata.sh
```diff
@@ -0,0 +1,118 @@
...
+jsonFile='metadata.json'
+canonicalMetadataFile="./$jsonFile"
+export maxCategories=3
+
+self="$(basename "$0")"
+
+usage() {
+	cat <<EOUSAGE
+
+usage: $self [--write] repo [...]
```
whalelines, line 17:
```suggestion
usage: $self [--write] REPO[...]
```

metadata.sh
```diff
@@ -0,0 +1,118 @@
...
+		else
+			(( failures++ )) || :
+		fi
+	fi
+
+	# TODO also check for required keys and/or types?
+	# the canonicalMetadataFile doesn't have too many categories since it is the source of categories
+	# all other metadata.json files must not be more than maxCategories or have categories that aren't in the canonical set
+	if [ "$repoFile" != "$canonicalMetadataFile" ]; then
+		export repoFile
```
whalelines, line 90: Is this necessary?

whalelines (commented):
Looks good. A couple minor things

metadata.sh
yosifkit: yes, the `jq` on the next lines accesses it via `env.repoFile` to generate the error strings which is slightly better than embedding it via shell by ending the single quote string, starting a double quote string and back.

whalelines (approved):
LGTM

tianon (approved):

PR #2449, opened 2024-05-17
ArangoDB: How to initialize the server language / license update

Simran-B:
Also add remark about --platform for ARM and update links


KVS85:
I approve these changes from [maintainers](https://github.com/docker-library/official-images/blob/master/library/arangodb#L1) prospective.

arangodb/content.md
```diff
@@ -31,32 +31,48 @@ In order to start an ArangoDB instance, run:
 unix> docker run -e ARANGO_RANDOM_ROOT_PASSWORD=1 -d --name arangodb-instance %%IMAGE%%
 ```
 
-Will create and launch the %%IMAGE%% docker instance as background process. The Identifier of the process is printed. By default ArangoDB listen on port 8529 for request and the image includes `EXPOSE 8529`. If you link an application container it is automatically available in the linked container. See the following examples.
+On macOS/ARM, the processor architecture chosen by default may be x86-64 (`linux/amd64`), which is not optimal. You can pass the `--platform` flag to the `docker run` command to specify the appropriate operating system and architecture for the container (`linux/arm64/v8`):
```
tianon, line 34: > On macOS/ARM, the processor architecture chosen by default may be x86-64 (`linux/amd64`), which is not optimal.

:grimacing: this is definitely a bug -- do you have more details for reproducing?

Simran-B: The co-worker who had trouble with her M1 device isn't with the company anymore. I would need to find someone else with Apple Silicon for testing to get more details. This may have been fixed since, either on our end or in Docker, I don't know. Choosing `linux/amd64` despite ARM hardware isn't wrong generally speaking because there is Rosetta, but using ARM natively should perform better and I believe you can also run into segfaults with ArangoDB in particular because the x86-64 builds use instructions that Rosetta can't emulate (AVX), leading to SIGILL.

tianon: What I mean is that on an M1 that's properly configured, it _will_ prefer `linux/arm64/v8`, not `amd64`, and if it isn't, that's a definite bug (in the Docker Engine, Docker Desktop, or the environment itself being configured incorrectly).

Simran-B: How can it be (mis-)configured to pick `amd64`, assuming it's not a bug? Are there multiple ways to affect the default platform choice?

tianon: Yes -- setting `DOCKER_DEFAULT_PLATFORM` in your local environment (or, for example, following a guide that suggests putting it into your `.bashrc`, which is a common problem I've seen) will override Docker's built-in automatic detection.

Simran-B: A coworker tried `echo $DOCKER_DEFAULT_PLATFORM; docker run --rm busybox:uclibc arch` on Apple silicon. The env var wasn't set and the selected architecture was `aarch64`. I rephrased the remark accordingly as the automatic selection seems to work as it should. I suppose the other coworker had an env var override, which is probable because of some tooling she used that originally supported x86 only.

arangodb/license.md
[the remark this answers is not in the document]
Simran-B: I'm not sure what you're asking for. The relevant links are in the updated text with more details about the license change.

LaurentGoderre (approved):

arangodb/license.md
```diff
@@ -1 +1,7 @@
-[ArangoDB itself is licensed under the Apache License](https://github.com/arangodb/arangodb/blob/devel/LICENSE), but it contains [software of third parties under their respective licenses](https://github.com/arangodb/arangodb/blob/devel/LICENSES-OTHER-COMPONENTS.md).
+The official Docker images of the ArangoDB Community Edition are governed by the [ArangoDB Community License](https://arangodb.com/2024/02/update-evolving-arangodbs-licensing-model-for-a-sustainable-future/). The use for commercial purposes is limited to a 100 GB on dataset size in production within a single cluster and a maximum of three clusters.
```
whalelines, line 1: Is the text for the ArangoDB Community License available anywhere? I could not find a link to it in the blog post.

arangodb/license.md
```diff
@@ -1 +1,7 @@
...
+
+The source code of the Community Edition is available under the [Business Source License 1.1 (BUSL-1.1)](https://github.com/arangodb/arangodb/blob/devel/LICENSE). Copying, modification, redistribution, non-commercial use, and commercial use in a non-production context are always allowed. Additionally, you can deploy BUSL-licensed ArangoDB source code for any purpose (including production) as long as you are not creating a commercial derivative work or offering, or are including it in a commercial product, application, or service. On the fourth anniversary of the first publicly available distribution of a specific version, the license changes to the permissive Apache 2.0 open-source license.
```
whalelines, line 3: Perhaps the DOI should be built from source so the BUSL-1.1 can be used rather than the ArangoDB Community License?

arangodb/license.md
Simran-B: No, the blog post is the license text at the moment. I asked this internally, too, but there is no other source it seems.

arangodb/license.md
Simran-B: The official images are specifically licensed under the ArangoDB community license. There will be no images or prepackaged binaries under a different license from ArangoDB.

Simran-B:
@tianon This can be merged now

arangodb/content.md
```diff
@@ -31,32 +31,48 @@ In order to start an ArangoDB instance, run:
 unix> docker run -e ARANGO_RANDOM_ROOT_PASSWORD=1 -d --name arangodb-instance %%IMAGE%%
 ```
 
-Will create and launch the %%IMAGE%% docker instance as background process. The Identifier of the process is printed. By default ArangoDB listen on port 8529 for request and the image includes `EXPOSE 8529`. If you link an application container it is automatically available in the linked container. See the following examples.
+Docker chooses the processor architecture for the image that matches your host CPU by default. If this is not the case, for example, because you have the `DOCKER_DEFAULT_PLATFORM` environment variable set to a different architecture, you can pass the `--platform` flag to the `docker run` command to specify the appropriate operating system and architecture for the container. For x86-64, use `linux/amd64`. On ARM, especially Apple silicon with no emulation for the required AVX instruction set extension, use `linux/arm64/v8`:
 
-In order to get the IP arango listens on, run:
+```console
+docker run --platform linux/arm64/v8 -p 8529:8529 -e ARANGO_RANDOM_ROOT_PASSWORD=1 arangodb-instance %%IMAGE%%
```
yosifkit, line 37:
```suggestion
docker run -d --platform linux/arm64/v8 -p 8529:8529 -e ARANGO_RANDOM_ROOT_PASSWORD=1 --name arangodb-instance %%IMAGE%%
```
Add `-d` for "Run container in background and print container ID" so that logs aren't printed to the terminal and add missing `--name` flag.

Simran-B: @yosifkit Well spotted, thanks! I pushed a fix.

yosifkit (approved):

PR #2450, opened 2024-06-05
Update chronograf and kapacitor: Specify how to run as root

powersj:
In a recent release we stopped running as the root user, but provided users a way to change this if they wanted to via an env variable. This adds this variable to the docs.


whalelines (approved):
LGTM

LaurentGoderre (approved):

PR #2452, opened 2024-06-11
update emqx content

id:


emqx/content.md
```diff
@@ -124,7 +115,7 @@ Let's create a static node list cluster from docker-compose.
 -	View cluster
 
 ```bash
-  $ docker exec -it my_emqx_emqx1_1 sh -c "emqx_ctl cluster status"
+  $ docker exec -it my_emqx_emqx1_1 sh -c "emqx ctl cluster status"
```
LaurentGoderre, line 118: For this one, maybe expand it to which version uses which command. i.e.

```bash
# Version x and below
$ docker exec -it my_emqx_emqx1_1 sh -c "emqx_ctl cluster status"

# Version and above
$ docker exec -it my_emqx_emqx1_1 sh -c "emqx ctl cluster status"

>  Cluster status: #{running_nodes => ['emqx@node1.emqx.io','emqx@node2.emqx.io'],
                    stopped_nodes => []}
```

emqx/content.md
```diff
@@ -32,51 +32,44 @@ The EMQX broker runs as Linux user `emqx` in the docker container.
...
-DEV_LISTENER__SSL__EXTERNAL__ACCEPTORS <--> listener.ssl.external.acceptors
-DEV_MQTT__MAX_PACKET_SIZE              <--> mqtt.max_packet_size
-DEV_LISTENERS__TCP__DEFAULT__BIND      <--> listeners.tcp.default.bind
-```
-
 For example, set MQTT TCP port to 1883
 
 ```console
-$ docker run -d --name emqx -e DEV_LISTENERS__TCP__DEFAULT__BIND=1883 -p 18083:18083 -p 1883:1883 %%IMAGE%%:latest
+$ docker run -d --name emqx -e EMQX_DASHBOARD__DEFAULT_PASSWORD=mysecret -p 18083:18083 -p 1883:1883 emqx/emqx:latest
```
tianon, line 52:
```suggestion
$ docker run -d --name emqx -e EMQX_DASHBOARD__DEFAULT_PASSWORD=mysecret -p 18083:18083 -p 1883:1883 %%IMAGE%%:latest
```

tianon (approved):
Generally looks OK to me -- just one reference to the `emqx/emqx` image which should instead use the template-provided `%%IMAGE%%` substitution so it refers to the appropriate image automatically.

emqx/content.md
id: `emqx_ctl` is an alias for `emqx ctl`.
The former is a legacy one, was used in v4, all v5 images support both.

emqx/content.md
```diff
@@ -32,51 +32,44 @@ The EMQX broker runs as Linux user `emqx` in the docker container.
 
 All EMQX Configuration in [`etc/emqx.conf`](https://github.com/emqx/emqx/blob/master/apps/emqx/etc/emqx.conf) can be configured via environment variables.
 
-By default, the environment variables with `EMQX_` prefix are mapped to key-value pairs in configuration files.
-
-You can change the prefix by overriding `HOCON_ENV_OVERRIDE_PREFIX`.
-
 Example:
 
 ```bash
```
whalelines, line 37: This code block should probably omit a syntax hint.

emqx/content.md
```diff
@@ -32,51 +32,44 @@ The EMQX broker runs as Linux user `emqx` in the docker container.
...
-EMQX_ZONES__DEFAULT__MQTT__MAX_PACKET_SIZE <--> zones.default.mqtt.max_packet_size
+EMQX_DASHBOARD__DEFAULT_PASSWORD       <--> dashboard.default_password
+EMQX_NODE__COOKIE                      <--> node.cookie
+EMQX_LISTENERS__SSL__default__ENABLE   <--> listeners.ssl.default.enable
 ```
 
+Note: The lowercase use of 'default' is not a typo. It is used to demonstrate that lowercase environment variables are equivalent.
+
 -	Prefix `EMQX_` is removed
 -	All upper case letters is replaced with lower case letters
```
whalelines, line 46:
```suggestion
-	All upper case letters are replaced with lower case letters
```

emqx/content.md
```diff
@@ -32,51 +32,44 @@ The EMQX broker runs as Linux user `emqx` in the docker container.
...
-Please read more about EMQX configuration in the [official documentation](https://www.emqx.io/docs/en/v5.0/admin/cfg.html).
+Please read more about EMQX configuration in the [official documentation](https://docs.emqx.com/en/emqx/latest/configuration/configuration.html)
 
 #### EMQX node name configuration
 
-| Options     | Default        | Mapped | Description                |
-|-------------|----------------|--------|----------------------------|
-| `EMQX_NAME` | container name | none   | EMQX node short name       |
-| `EMQX_HOST` | container IP   | none   | EMQX node host, IP or FQDN |
+A node name consists of two parts, `EMQX_NAME` part and `EMQX_HOST` part connected by a the symbol `@`. For example: `emqx@127.0.0.1`.
```
whalelines, line 59:
```suggestion
A node name consists of two parts, `EMQX_NAME` part and `EMQX_HOST` part connected by the symbol `@`. For example: `emqx@127.0.0.1`.
```

emqx/content.md
```diff
@@ -32,51 +32,44 @@ The EMQX broker runs as Linux user `emqx` in the docker container.
...
 
-These environment variables are used during container startup phase only in [docker-entrypoint.sh](./docker-entrypoint.sh).
+Environment variables `EMQX_NODE_NAME` or `EMQX_NODE__NAME` can be used to set a EMQX node name. If neither of them is set, EMQX will resolve its node name from the running environment or other environment varialbes used for node discovery.
```
whalelines, line 61: This confuses me a bit. The implication from the previous diff chunk is that one can set `EMQX_NAME` and `EMQX_HOST` to set the node name. Is that true? If it is, is one or the other approach preferred or take priority?

emqx/content.md
```diff
@@ -32,51 +32,44 @@ The EMQX broker runs as Linux user `emqx` in the docker container.
...
 
-If `EMQX_NAME` and `EMQX_HOST` are set, and `EMQX_NODE_NAME` is not set, `EMQX_NODE_NAME=$EMQX_NAME@$EMQX_HOST`. Otherwise `EMQX_NODE_NAME` is taken verbatim.
+When running in docker, by default, `EMQX_NAME` and `EMQX_HOST` are resolved as below:
```
whalelines, line 63: And this implies that `EMQX_NAME` and `EMQX_HOST` are derived from `EMQX_NODE_NAME`. Could the logic used be made more explicit here?

whalelines (commented):
Thanks for the update. I've requested a couple clarifications.

id:
@whalelines updated the text, please take a look.

whalelines (approved):
Turns out less is more, more clear that is. Thanks, LGTM

tianon (approved):

PR #2463, opened 2024-07-05
Update examples from "ruby:3.0" to "ruby:3.3"

Earlopain:
Previous: https://github.com/docker-library/docs/pull/2051

3.0 images are not build/updated anymore

-------

In a separate commit, I've dropped mentions of evn variables that were once set but aren't anymore. See https://github.com/docker-library/ruby/pull/209 and https://github.com/docker-library/ruby/pull/306

### DIFF
--- a/ruby/content.md
+++ b/ruby/content.md
@@ -11,7 +11,7 @@ Ruby is a dynamic, reflective, object-oriented, general-purpose, open-source pro
 ## Create a `Dockerfile` in your Ruby app project
 
 ```dockerfile
-FROM %%IMAGE%%:3.0
+FROM %%IMAGE%%:3.3
 
 # throw errors if Gemfile has been modified since Gemfile.lock
 RUN bundle config --global frozen 1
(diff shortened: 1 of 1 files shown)


LaurentGoderre (approved):

ruby/content.md
```diff
@@ -11,7 +11,7 @@ Ruby is a dynamic, reflective, object-oriented, general-purpose, open-source pro
 ## Create a `Dockerfile` in your Ruby app project
 
 ```dockerfile
-FROM %%IMAGE%%:3.0
+FROM %%IMAGE%%:3.3
```
whalelines, line 14: Perhaps it would be better to just use the `latest` tag in these examples?

ruby/content.md
```diff
@@ -40,15 +40,15 @@ $ docker run -it --name my-running-script my-ruby-app
 The above example `Dockerfile` expects a `Gemfile.lock` in your app directory. This `docker run` will help you generate one. Run it in the root of your app, next to the `Gemfile`:
 
 ```console
-$ docker run --rm -v "$PWD":/usr/src/app -w /usr/src/app %%IMAGE%%:3.0 bundle install
+$ docker run --rm -v "$PWD":/usr/src/app -w /usr/src/app %%IMAGE%%:3.3 bundle install
```
whalelines, line 43: It definitely seems safe to use the `latest` tag on the CLI.

whalelines (commented):
Thanks for the update! A couple thoughts

ruby/content.md
tianon: Using an explicit version was intentional so that we're modelling the appropriate behavior for end-users who _shouldn't_ be pinning to `:latest` here -- this is a persistent issue though that we've considered changing our template engine for so that we could have a per-image calculation of this "best example tag" value.  (Barring that, IMO this change is correct :heart:)

tianon (approved):
(approving, but not merging yet since I am open to discussion/convincing on the points where I've disagreed :heart:)

whalelines (approved):
LGTM

PR #2465, opened 2024-07-05
Trim descriptions based on bytes, not characters

tianon:
It turns out that the Hub description length limit is 25k *bytes*, not characters (contrary to https://github.com/docker/hub-feedback/issues/238 and https://github.com/docker/roadmap/issues/475).  We've gotten lucky that our few unicode glyphs have landed so far such that we've never run into a problem with this before now, but we're currently hitting this while trying to push the `amd64/nextcloud` description in that the trimmed version ends up being exactly 25,003 bytes thanks to the unicode glyphs throwing off our trimming calculations (and getting really [un]lucky with the exact positioning of the markdown such that our trimming doesn't walk back far enough).

This fixes that by calculating and trimming based on bytes, accounting for cases of unicode glyphs directly spanning the 25k byte boundary.

After applying this fix, the same breaking `amd64/nextcloud` trimmed description is 24,725 bytes instead.

(It's also admittedly possible that this is actually a regression in Hub and that the previous implementation's limit actually _was_ 25k characters, not bytes, but I haven't done that digging and have to live in today's world, so this change is still correct.)

### DIFF
--- a/push.pl
+++ b/push.pl
@@ -15,7 +15,9 @@ use Mojo::Util qw(decode encode trim url_escape);
 use Term::UI;
 use Term::ReadLine;
 
-my $hubLengthLimit = 25_000;
+require bytes; # this is not recommended, but we *only* use "bytes::length" from it to determine whether we need to do a more correct conversion to/from bytes for trimming (see $hubLengthLimit and usages)
+
+my $hubLengthLimit = 25_000; # NOTE: this is *bytes*, not characters 🙃
 my $githubBase = 'https://github.com/docker-library/docs/tree/master'; # TODO point this at the correct "dist-xxx" branch based on "namespace"
 
 my $username;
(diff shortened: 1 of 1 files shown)


tianon:
I think `bytes::substr` is screwing up the encoding of the variable (so this still isn't quite right yet).

tianon:
Ok, thanks to perlmonks, got a pretty clean fix that _also_ handles the "glyphs at the boundary" edge case :smile:

tianon:
To attempt to test the boundary cases, I created a new description via `yes '💩' | head -n12500` which is exactly 25k characters (12500x💩 + 12500x`\n`), and this new code safely trimmed it to 24,999 bytes where the old code happily tried to push it (and got `patch to https://hub.docker.com/v2/repositories/tianon/test/ failed: {"message":"httperror 400: request had 1 validation error","errinfo":{"fielderrors":{"full_description":"Exceeded max number of characters"}}}`).

whalelines (approved):
LGTM

yosifkit (approved):

PR #2466, opened 2024-07-05
Update `push.pl` runtime to Perl 5.39 and Debian Bookworm

tianon:
(verified successful/working while testing https://github.com/docker-library/docs/pull/2465)


whalelines (approved):
LGTM

yosifkit (approved):

PR #2468, opened 2024-07-16
Update README.md - prefer stable tag

TomorrowToday:
Many users of the nginx images may not be familiar with Nginx's odd-even/unstable-stable versioning that results in the default docker version often being unstable. This pull request updates documentation to use `nginx:stable` and explains why it's preferred over `nginx` / `nginx:latest`.

### DIFF
--- a/nginx/README.md
+++ b/nginx/README.md
@@ -84,13 +84,13 @@ Nginx (pronounced "engine-x") is an open source reverse proxy server for HTTP, H
 ## Hosting some simple static content
 
 ```console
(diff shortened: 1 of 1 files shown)


TomorrowToday:
closing, should have read the main README.md sorry for the noise

tianon:
In the future, please force push to your existing branch instead of creating a new one -- that will update the existing PR and decrease the amount of notifications for maintainers/watchers (especially those with threading enabled in their email clients).

tianon:
Oh, my apologies, I misinterpreted https://github.com/docker-library/docs/issues/2470 as a new PR :facepalm:

PR #2477, opened 2024-07-29
Officially deprecate ClefOS

tianon:
cc @nealef for a vibe and wording check :heart:

(I'm honestly kind of sad to see it go, because it was fun and even sometimes helpful for catching bugs/assumptions to have a DOI that _didn't_ support `amd64`, but alas - end of an era :smile:)

Refs https://github.com/docker-library/official-images/pull/7964#issuecomment-2253636315

### DIFF
--- a/clefos/README-short.txt
+++ b/clefos/README-short.txt
@@ -1,1 +1,1 @@
-The official build of ClefOS.
+DEPRECATED; The official build of ClefOS.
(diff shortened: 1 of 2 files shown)


yosifkit (approved):

whalelines (approved):
LGTM

PR #2478, opened 2024-07-29
Remove deprecated/removed repos

tianon:
Once their `library/` files are gone, we don't need to keep updating their descriptions.

(Depends on https://github.com/docker-library/official-images/pull/17276)


tianon:
This is ready for review now :eyes:

yosifkit (approved):

PR #2479, opened 2024-07-31
Plone: Add info about Plone 6 Docker images

avoinea:


tianon:
I have to admit I'm a little confused -- does this mean the https://hub.docker.com/_/plone image is effectively deprecated and won't be getting 6.x+? :sweat_smile:

(If so, we should make this change a little differently, especially with a `deprecated.md` so the note shows up at the top of the image description.)

plone/content.md
stevepiercy, line 1: Both instances of:

`https://docs.plone.org/manage/docker/docs/index.html`

need to be updated to:

`https://5.docs.plone.org/manage/docker/docs/index.html`

plone/content.md
```diff
@@ -6,7 +6,8 @@
 
 ## Features
 
--	Images for Plone 5.x and Plone 4.x
+-	Images for **Plone 5.x** and **Plone 4.x**
```
stevepiercy, line 9:
```suggestion
```

plone/content.md
```diff
@@ -6,7 +6,8 @@
...
+-	Images for **Plone 6.x** are available at [plone/plone-backend](https://hub.docker.com/r/plone/plone-backend) and [plone/plone-frontend](https://hub.docker.com/r/plone/plone-frontend)
```
stevepiercy, line 10:
```suggestion
```

stevepiercy (commented):
Bah, I cannot request changes, only make a comment.

Anyway, versions are not features, so I created a new section, which should handle the concerns expressed.

I also added suggestions to update links to the old documentation for previous versions.

stevepiercy:
Apparently my suggestion to add a section did not get saved in my review. I'll try again, now that I have permission.

plone/content.md
avoinea: Fixed.

avoinea:
> I have to admit I'm a little confused -- does this mean the https://hub.docker.com/_/plone image is effectively deprecated and won't be getting 6.x+? 😅

@tianon As Plone 6 is not a monolith app anymore and it comes with a decoupled React front-end we don't have a consensus in the Plone Community on what we should do with this image, yet :see_no_evil:

I can see there are other apps that tries to put everything in a all-in-one-docker-image like https://github.com/nextcloud/all-in-one?tab=readme-ov-file#nextcloud-all-in-one but I'm not sure what is the best way to handle this situation for the Official Docker Images. Maybe you can point out to some good examples / practices. 

> (If so, we should make this change a little differently, especially with a `deprecated.md` so the note shows up at the top of the image description.)

Meanwhile, until we have a decision, I added the `deprecated.md` file pointing to the officially supported Plone 6 images.

mauritsvanrees (commented):
As Plone Release Manager this change looks fine to me.

I don't have experience with how Docker Hub handles these files though, and what the recommendations are.

Quick background:

* For Plone 5 this image is probably fine, but Plone 5 only gets security support until October this year.
* For Plone 6 indeed `plone/plone-backend` and `plone/plone-frontend` should be used and have been available since at least 2022.
* I don't know if those two could somehow be promoted to official Docker images, or if it would be better to somehow combine them in one (which seems a bit tricky as two processes should run).

stevepiercy:
FYI, I'll be back from vacation tonight, and I'll look into how these docs are auto-generated from their counterparts in the Plone GitHub repos, and work on PRs there. From my quick glance of Docker's docs, it looks like the PRs will go under Plone first, then will get generated in Docker repos.

tianon:
Ah, interesting!  I would definitely _not_ recommend doing something like an all-in-one, because I have yet to see a truly reliable (and still minimal) "process supervisor" to help manage multiple processes in one container.  Two images is perfectly reasonable, especially if the documentation describes how to run them (something like `plone:6-frontend` + `plone:6-backend` is what I'd personally suggest, but `plone:frontend-6` and `plone:backend-6` would be pretty reasonable too, however the latter to me personally suggests that the versioning between frontend and backend might _not_ stay in sync over time, where the former makes it clear that each are individual components from the "6" release).

stevepiercy (commented):
LGTM. Please merge!

stevepiercy:
Closes https://github.com/plone/plone.docker/issues/179.

stevepiercy:
I don't have merge permission. Who does?

After this PR is merged, then I can start work on promoting plone-backend and plone-frontend as official Docker images. I'll merge the docs files from this PR with those from the other repos, then open a pull request for each image.

When Plone 5.x exits security support, then I think we should remove just `plone` from the official images. See https://hub.docker.com/search?q=plone&image_filter=official.

stevepiercy:
Ready for another run of workflows.

stevepiercy:
I missed a couple of details, and I think I finally got them all this time. Ready for one more run.

I didn't realize I could run the CI check locally with `./update.sh plone` followed by `./markdownfmt.sh -d plone`. Sorry for the noise.

whalelines:
Just to make sure we all have the same understanding, the intention is to deprecate the `plone` DOI entirely and only move forward with `plone/plone-backend` and `plone/plone-frontend`, i.e., to not provide two variants in the `plone` DOI, `plone:6-frontend` and `plone:6-backend`, as @tianon suggested?

stevepiercy:
@whalelines, I don't understand what you mean by two variants. @tianon said:

> Two images is perfectly reasonable

---

Are the docs in this PR technically approved? I've been waiting for that piece before I start. If approved, then I can start today on the frontend and backend images' docs in separate PRs. Please let me know. Thank you!

whalelines:
There are currently three _repositories_ serving Plone container images.

1.  `plone`: the Docker Official Image repository for Plone that has images for versions 4 and 5
2.  `plone/plone-backend`: the Docker-Sponsored Open Source repository for the backend image for Plone 6
3.  `plone/plone-frontend`: the Docker-Sponsored Open Source repository for the front image for Plone 6

Currently, this PR for updating the documentation for the `plone` DOI effectively deprecated the entire `plone` DOI _repository_, suggesting people get the latest Plone container images from the two Plone DSOS repositories.

What @tianon suggested was that the `plone` DOI repository can provide both the backend and frontend images for Plone 6. In contrast to how the Plone DSOS repositories are set up, the frontend and backend variants could both be available in the single `plone` DOI repository by using separate _tags_ for each variant, e.g., `plone:6-backend` and `plone:6-frontend`. Lots of DOI repositories provide multiple image variants, e.g., leveraging different base images, language runtimes, or use cases.

I just want to make sure your intention is to deprecate the `plone` DOI entirely as this PR suggests.

whalelines:
Currently the [plone bashbrew file](https://github.com/docker-library/official-images/blob/master/library/plone) has a single stanza. You would just add a distinct stanza for each variant, e.g., [aerospike](https://github.com/docker-library/official-images/blob/master/library/aerospike), [arangodb](https://github.com/docker-library/official-images/blob/master/library/arangodb), [backdrop](https://github.com/docker-library/official-images/blob/master/library/backdrop), etc.

tianon (commented):
This seems OK now, although perhaps you'd like to hold off on the deprecation notice now that you're planning to add `plone:6` images to DOI? (https://github.com/docker-library/official-images/pull/17281#issuecomment-2916575746)

plone/deprecated.md
```diff
@@ -0,0 +1,12 @@
+# ⚠️ DEPRECATION NOTICE ⚠️
+
+-	Plone 4.x is no longer supported.
+-	Plone 5.x is out of maintenance support, and will exit security support on 2024-10-31.
```
stevepiercy, line 4:
```suggestion
-	Plone 5.x is no longer supported.
```

stevepiercy (commented):
@avoinea please have a look.

plone/deprecated.md
```diff
@@ -0,0 +1,12 @@
+# ⚠️ DEPRECATION NOTICE ⚠️
+
+-	Plone 4.x is no longer supported.
+-	Plone 5.x is no longer supported.
+-	It is strongly recommended to migrate to the latest available Plone version. See https://plone.org/download/release-schedule for details.
+
+The official Docker images for Plone 5.x and Plone 4.x are available here, and are maintained by community volunteers.
```
tianon: Should this be present tense, or past tense?  (I don't think they're being actively maintained right now, are they?)

We should also add a `DEPRECATED; ` prefix to `README-short.txt` (as seen in https://github.com/docker-library/docs/pull/2583).

tianon: To add a little more detail for https://github.com/docker-library/docs/pull/2583, my goal there was to make it really clear (immediately) to users on the page that any images they find here are old, outdated, and unsupported, and they should look elsewhere ASAP - I think this description gets there, but it's not quite as immediately obvious.

I don't have a strong preference either way, as long as we make it clear that this repository is officially deprecated and unsupported (and I think the tense here is the only blocking issue I have with that).

stevepiercy: @tianon I lack permission to make a comment on your PR https://github.com/docker-library/docs/pull/2583/, so here it is:

`plone/deprecated.md` is not listed as one of those in `docker-library/docs`. Is it used or ignored? See https://github.com/docker-library/docs?tab=readme-ov-file#files-related-to-an-images-docs.

I'm revising this PR to incorporate your suggestions, once we verify that question.

stevepiercy: I made a few revisions. Hopefully this file gets included in the repo.

stevepiercy: I verified that this file gets prepended to the top of the automatically generated `README.md`. You can test by running `./update.sh plone` from the root of the repo.

@tianon @avoinea @fredvd would you please take another look? Thank you!

avoinea: LGTM.

plone/content.md
```diff
@@ -4,13 +4,19 @@
 
 %%LOGO%%
 
+## Supported versions
+
+-	Plone 5.x
+-	Plone 4.x
+
+See the [deprecation notice](./deprecated.md) for Plone 6 and later.
```
tianon: Once this description ends up on Docker Hub (which is ultimately what all these files are for), this cross-file link won't work.  I would suggest just wording this something like "See the deprecation notice above" or similar.

plone/content.md
```diff
@@ -4,13 +4,19 @@
 
 %%LOGO%%
 
+## Supported versions
```
tianon: This wording is really tripping me up -- as I understand it, neither of these versions are "supported" either upstream _or_ here, which is the whole point of adding the deprecation notice.  Also, why is this duplicating the content from the deprecation notice which will appear on Docker Hub just above it?

plone/content.md
```diff
@@ -43,99 +49,100 @@ $ docker run --link=zeo -e ZEO_ADDRESS=zeo:8080 -p 8082:8080 %%IMAGE%%
...
 
 ```console
  -e ADDONS="eea.facetednavigation collective.easyform" \
  -e VERSIONS="eea.facetednavigation=13.3 collective.easyform=2.1.0"
 ```
 
-RestAPI:
+To use Plone REST API, start the Plone Docker image, then issue a command to `curl`.
 
 ```console
```
tianon: If you're dropping the `$` prefix on the command lines, this is no longer `console` but rather `shell` or `bash`:

````suggestion
```bash
````

File: plone/deprecated.md [hunk 8]
```diff
@@ -0,0 +1,15 @@
+# ⚠️ DEPRECATION NOTICE ⚠️
+
+-	Plone 4.x is no longer supported.
+-	Plone 5.x is no longer supported.
+-	It is strongly recommended to migrate to the latest available Plone version. See https://plone.org/download/release-schedule for details.
+
+The official Docker images for Plone 5.x and Plone 4.x are available here.
+
+Beginning with Plone 6.x, the frontend and backend run in separate Docker images. There's also a a ZEO server Docker image. All Plone 6.0 and later images are supported by the Plone Foundation and are maintained by community volunteers.
```
tianon, line 7: I don't love that it's not completely clear that _this_ repository (https://hub.docker.com/_/plone) is completely deprecated, but I can live with this wording.

plone/content.md
stevepiercy: I removed this section in e4d1c6ec3aeb8947e354beb0b44b15c2da83bcd8 as redundant.

plone/content.md
stevepiercy: I removed this section in e4d1c6ec3aeb8947e354beb0b44b15c2da83bcd8 as redundant.

plone/content.md
stevepiercy: I went with the generic `shell` instead of a specific flavor. Fixed in 3abd58fd9bef968718f9bebbf5b264b04c107dad.

plone/deprecated.md
stevepiercy: You're right. Fixed in c1ccceb174f1f7eda050f0cda2a95f3ccf2ab5a4.

stevepiercy:
@tianon would you please make one more pass? Thank you for your feedback!

plone/deprecated.md
```diff
@@ -0,0 +1,15 @@
+# ⚠️ DEPRECATION NOTICE ⚠️
+
+-	Plone 4.x is no longer supported.
+-	Plone 5.x is no longer supported.
+-	It is strongly recommended to migrate to the latest available Plone version. See https://plone.org/download/release-schedule for details.
+
+The official Docker images for Plone 4.x and Plone 5.x are available at https://hub.docker.com/_/plone. These images are depecrated.
+
```
tianon: I'm sorry, but this wording is still really weird (and buries the lede) -- for context, this entire markdown file is going to be the very top of the image description _on_ https://hub.docker.com/_/plone so this whole sentence is going to read there as if it's written in the third-person.

I think the text above this already makes it clear that 4.x and 5.x are not supported, and we're already very obviously labelled a deprecation notice (and the following text is the "call to action" for where to find 6.x+), so I honestly think we should just remove this whole line:

```suggestion

```

stevepiercy: Thanks for the context. I agree. Committing.

tianon (approved):

stevepiercy:
Thank you @tianon. I'm grateful for your help to push this one across the finish line.

Now we Plonistas can turn to our remaining Docker image tasks.

- https://github.com/plone/dockerhub-docs/issues/1
- https://github.com/plone/plone.docker/issues/179

@fredvd @avoinea @mauritsvanrees @ericof @pbauer would you please take a look at the checklists in those issues and update them accordingly? Thank you!

PR #2483, opened 2024-08-24
Update docs for Caddy image with a fix to its reload feature

BjoernLudwigPTB:
Last August unexpected behaviour of the graceful reload feature was discussed in this [upstream issue](https://github.com/caddyserver/caddy/issues/5735) and a fix proposed alongside the suggestion to add that fix to the docs. This has not yet happened but should be solved by this PR.

This is my first contribution here, so if there is anything I should address to finalize this PR, I ask you kindly to let me know.

This one resolves #1858 as well.

### DIFF
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -123,7 +123,7 @@ See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for more details.
 
 ### Docker Compose example
 
-If you prefer to use `docker compose` to run your stack, here's a sample service definition which goes in a file named `compose.yaml`. 
+If you prefer to use `docker compose` to run your stack, here's a sample service definition which goes in a file named `compose.yaml`.
 
 ```yaml
 services:
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -123,7 +123,7 @@ See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for more details.
 
 ### Docker Compose example
 
-If you prefer to use `docker-compose` to run your stack, here's a sample service definition which goes into your `compose.yaml`. 
+If you prefer to use `docker compose` to run your stack, here's a sample service definition which goes in a file named `compose.yaml`. 
 
 ```yaml
 services:
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -46,7 +46,7 @@ $ docker run -d -p 80:80 \
 
 #### ⚠️ Do not mount the Caddyfile directly at `/etc/caddy/Caddyfile`
 
-This effectively disables Caddy's graceful reload feature in many cases depending on the way you apply changes to the file as discussed in [this issue](https://github.com/caddyserver/caddy/issues/5735#issuecomment-1675896585).
+If vim or another editor is used that changes the inode of the edited file, the changes will only be applied within the container when the container is recreated , which is explained in detail in this [Medium article](https://medium.com/@jonsbun/why-need-to-be-careful-when-mounting-single-files-into-a-docker-container-4f929340834). When using such an editor, Caddy's graceful reload functionality might not work as expected, as described in [this issue](https://github.com/caddyserver/caddy/issues/5735#issuecomment-1675896585).
 
 ### Automatic TLS with the Caddy image
 
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -35,7 +35,7 @@ $ curl http://localhost/
 hello world
 ```
 
-To override the default [`Caddyfile`](https://github.com/caddyserver/dist/blob/master/config/Caddyfile), you can create one in the subfolder `caddyfile` at `$PWD/caddyfile/Caddyfile` and mount this folder at `/etc/caddy`:
+To override the default [`Caddyfile`](https://github.com/caddyserver/dist/blob/master/config/Caddyfile), you can create one in the subfolder `conf` at `$PWD/conf/Caddyfile` and mount this folder at `/etc/caddy`:
 
 ```console
 $ docker run -d -p 80:80 \
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -150,4 +150,4 @@ volumes:
 
 Defining the data volume as [`external`](https://docs.docker.com/compose/compose-file/compose-file-v3/#external) makes sure `docker-compose down` does not delete the volume. You may need to create it manually using `docker volume create [project-name]_caddy_data`.
 
-Graceful reloads can then be conducted via `docker compose exec --workdir /etc/caddy caddy caddy reload`.
+Graceful reloads can then be conducted via `docker compose exec -w /etc/caddy caddy caddy reload`.
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -144,10 +144,7 @@ services:
 
 volumes:
   caddy_data:
-    external: true
   caddy_config:
 ```
-
-Defining the data volume as [`external`](https://docs.docker.com/compose/compose-file/compose-file-v3/#external) makes sure `docker-compose down` does not delete the volume. You may need to create it manually using `docker volume create [project-name]_caddy_data`.
 
 Graceful reloads can then be conducted via `docker compose exec -w /etc/caddy caddy caddy reload`.
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -46,7 +46,7 @@ $ docker run -d -p 80:80 \
 
 #### ⚠️ Do not mount the Caddyfile directly at `/etc/caddy/Caddyfile`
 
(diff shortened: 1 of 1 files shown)


whalelines (approved):
Looks reasonable to me, as long as there remains only a single file under the image's `/etc/caddy` directory.

tianon (approved):
Agreed, seems OK to me -- cc @hairyhenderson @francislavoie (`caddy` image maintainers)

tianon:
(Looks like there's also a bit of trailing whitespace that our `markdownfmt` is balking at)

caddy/content.md
```diff
@@ -35,15 +35,19 @@ $ curl http://localhost/
 hello world
 ```
 
-To override the default [`Caddyfile`](https://github.com/caddyserver/dist/blob/master/config/Caddyfile), you can mount a new one at `/etc/caddy/Caddyfile`:
+To override the default [`Caddyfile`](https://github.com/caddyserver/dist/blob/master/config/Caddyfile), you can create one in the subfolder `caddyfile` at `$PWD/caddyfile/Caddyfile` and mount this folder at `/etc/caddy`:
```
francislavoie, line 38: Use a different name for the dir, because it can be confusing the difference between the dir named `caddyfile` and the config file named `Caddyfile`. Using `conf/` is less confusing, I think.

```suggestion
To override the default [`Caddyfile`](https://github.com/caddyserver/dist/blob/master/config/Caddyfile), you can create one in the subfolder `conf` at `$PWD/conf/Caddyfile` and mount this folder at `/etc/caddy`:
```

caddy/content.md
```diff
@@ -35,15 +35,19 @@ $ curl http://localhost/
...
 
 ```console
 $ docker run -d -p 80:80 \
-    -v $PWD/Caddyfile:/etc/caddy/Caddyfile \
+    -v $PWD/caddyfile:/etc/caddy \
```
francislavoie, line 42:
```suggestion
    -v $PWD/conf:/etc/caddy \
```

caddy/content.md
```diff
@@ -35,15 +35,19 @@ $ curl http://localhost/
...
     -v caddy_data:/data \
     %%IMAGE%%
 ```
 
+#### ⚠️ Do not mount the Caddyfile directly at `/etc/caddy/Caddyfile`
+
+This effectively disables Caddy's graceful reload feature in many cases depending on the way you apply changes to the file as discussed in [this issue](https://github.com/caddyserver/caddy/issues/5735#issuecomment-1675896585).
```
francislavoie, line 49: I don't agree with this wording/framing. This isn't quite true, it only breaks if you happen to be using `vim`/`vi` in which case the way it saves files isn't tracked properly by Docker. But if you use a different editor like `nano` then the issue doesn't happen.

So I would prefer if this warned about making note of which editor you use to edit the file, rather than saying that it "breaks reloading" in general.

caddy/content.md
```diff
@@ -119,11 +123,9 @@ See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for more details.
 
 ### Docker Compose example
 
-If you prefer to use `docker-compose` to run your stack, here's a sample service definition.
+If you prefer to use `docker-compose` to run your stack, here's a sample service definition which goes into your `compose.yaml`.
```
francislavoie, line 126:
```suggestion
If you prefer to use `docker compose` to run your stack, here's a sample service definition which goes in a file named `compose.yaml`. 
```

caddy/content.md
```diff
@@ -147,3 +149,5 @@ volumes:
 ```
 
 Defining the data volume as [`external`](https://docs.docker.com/compose/compose-file/compose-file-v3/#external) makes sure `docker-compose down` does not delete the volume. You may need to create it manually using `docker volume create [project-name]_caddy_data`.
```
francislavoie, line 151:
```suggestion
Defining the data volume as [`external`](https://docs.docker.com/compose/compose-file/compose-file-v3/#external) makes sure `docker compose down` does not delete the volume. You may need to create it manually using `docker volume create [project-name]_caddy_data`.
```

caddy/content.md
```diff
@@ -147,3 +149,5 @@ volumes:
...
+
+Graceful reloads can then be conducted via `docker compose exec --workdir /etc/caddy caddy caddy reload`.
```
francislavoie, line 153: Can be slightly shortened. We use `-d` and `-p` earlier so I think we should stay consistent and use the short form:

```suggestion
Graceful reloads can then be conducted via `docker compose exec -w /etc/caddy caddy caddy reload`.
```

caddy/content.md
BjoernLudwigPTB: I am absolutely happy with this suggestion.

caddy/content.md
BjoernLudwigPTB: Agree, I will come up with a better wording reflecting the actual scope of the circumstances in which there is something to warn about. I submitted this proposal from the perspective of someone having followed the documentation exactly and due to having used vim to edit the file, I ran into the described situation.

caddy/content.md
BjoernLudwigPTB: In fact, now that we are talking about this section as well: May I suggest to delete this paragraph and the parameter `external = true` in the suggested _compose.yaml_  as a `docker compose down` wouldn't actually delete the volume. One would need to provide an addional `-v` or `--volumes` in addition to loose its content. In many cases providing this flag additionally would cause damage to applications, which usually do not mention this in their sample configurations or examples as Docker users usually seem to be aware of the consequences. If you agree @francislavoie  I would simply delete these two parts in another commit.

caddy/content.md
```diff
@@ -119,11 +123,9 @@ See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for more details.
 
 ### Docker Compose example
 
-If you prefer to use `docker-compose` to run your stack, here's a sample service definition.
+If you prefer to use `docker compose` to run your stack, here's a sample service definition which goes in a file named `compose.yaml`.
```
BjoernLudwigPTB, line 126:
```suggestion
If you prefer to use `docker compose` to run your stack, here's a sample service definition which goes in a file named `compose.yaml`.
```

BjoernLudwigPTB: This solves the trailing whitespace issue.

caddy/content.md
BjoernLudwigPTB: @francislavoie I have now suggested a wording that better reflects the situation that is causing the actual problem. This makes the paragraph a little lengthier, but also more precise. What do you think? Would you agree to these changes?

caddy/content.md
BjoernLudwigPTB: @francislavoie And please review my comment above and give your opinion on that.
Suggestion 1: remove this paragraph and the `external = true` parameter for the volume in the compose.yaml.
Option 2: Apply your suggested changes and leave everything else as is.

BjoernLudwigPTB: @francislavoie Please get back to this soon, so we can resolve the issues and merge this PR. :pray:

francislavoie: Hmm yeah, sure. We don't specify `external: true` in https://caddyserver.com/docs/running#setup anymore, so I figure we can strip it out here too.

caddy/content.md
```diff
@@ -35,15 +35,19 @@ $ curl http://localhost/
...
 $ docker run -d -p 80:80 \
-    -v $PWD/Caddyfile:/etc/caddy/Caddyfile \
+    -v $PWD/conf:/etc/caddy \
     -v caddy_data:/data \
     %%IMAGE%%
 ```
 
+#### ⚠️ Do not mount the Caddyfile directly at `/etc/caddy/Caddyfile`
+
+If vim or another editor is used that changes the inode of the edited file, the changes will only be applied within the container when the container is recreated , which is explained in detail in this [Medium article](https://medium.com/@jonsbun/why-need-to-be-careful-when-mounting-single-files-into-a-docker-container-4f929340834). When using such an editor, Caddy's graceful reload functionality might not work as expected, as described in [this issue](https://github.com/caddyserver/caddy/issues/5735#issuecomment-1675896585).
```
francislavoie, line 49:
```suggestion
If vim or another editor is used that changes the inode of the edited file, the changes will only be applied within the container when the container is recreated, which is explained in detail in this [Medium article](https://medium.com/@jonsbun/why-need-to-be-careful-when-mounting-single-files-into-a-docker-container-4f929340834). When using such an editor, Caddy's graceful reload functionality might not work as expected, as described in [this issue](https://github.com/caddyserver/caddy/issues/5735#issuecomment-1675896585).
```

caddy/content.md
francislavoie: Looks good. I don't love linking to a medium article (awkward website, lots of annoying banners, content may not be there forever) although the content is good. But not a big deal.

caddy/content.md
BjoernLudwigPTB: Alright, thanks for your look into it.

BjoernLudwigPTB:
@francislavoie Thanks for your recent review. I applied the agreed upon changes. Ready for approval.

BjoernLudwigPTB:
@whalelines @tianon All conversations and issues are resolved. Is a formal approval by Francis needed or can we merge this PR as is?

francislavoie:
LGTM :+1:

tianon:
Looks good, thanks to you both :+1:

PR #2488, opened 2024-09-12
Update Redis repository URL to the Correct URL

adamiBs:
This PR updates the `git-repo` URL to the updated location of the Redis URL.

### DIFF
--- a/redis/github-repo
+++ b/redis/github-repo
@@ -1,1 +1,1 @@
-https://github.com/docker-library/redis
+https://github.com/redis/docker-library-redis


whalelines (approved):
LGTM

tianon:
We should probably also de-symlink `maintainer.md` and update it with more accurate text (perhaps other changes too?)

tianon:
Doing a quick glance, another good example is the "`32bit` variant" section of `content.md` that no longer applies (as that variant no longer exists).

adamiBs:
I'll create a PR for that (I'll consult with peers on updated docs in general, since we could update other sections too).

PR #2494, opened 2024-10-10
Make memcached configuration section more generic

LaurentGoderre:


memcached/content.md
```diff
@@ -12,12 +12,31 @@ Memcached's APIs provide a very large hash table distributed across multiple mac
...
+
+```console
+$ docker run --rm %%IMAGE%% -h
+```
+
+Using `docker run`:
 
 ```console
-$ docker run --name my-memcache -d %%IMAGE%% memcached -m 64
+docker run --name my-memcache -d %%IMAGE%% memcached:alpine --memory-limit=64
```
tianon, line 26: This instance of `memcached` was the command to run, not the image name :sweat_smile:

I also don't see a good reason to prefer `:alpine`, especially as we have a section below that explains why Alpine is perhaps not ideal (I don't think they've explicitly stated it, but I'd be very surprised if upstream considers Alpine / musl as an "officially supported" target :see_no_evil:)

```suggestion
$ docker run --name my-memcache -d %%IMAGE%% --memory-limit=64
```

memcached/content.md
```diff
@@ -12,12 +12,31 @@ Memcached's APIs provide a very large hash table distributed across multiple mac
...
+```yaml
+services:
+  memcached:
+    image: %%IMAGE%%:alpine
+    command: [
+      "--conn-limit=1024",
+      "--memory-limit=64",
+      --threads=4"
+    ]
+```
```
tianon, line 40: I think the trailing comma situation makes the `[]` syntax here a little harder to maintain correctly (and we're missing a `"` :sweat_smile:), so I think we should swap this to the more standard "list" syntax, even though `- --xxx` looks kind of funny:

````suggestion
or using Docker Compose:

```yaml
services:
  memcached:
    image: %%IMAGE%%
    command:
      - --conn-limit=1024
      - --memory-limit=64
      - --threads=4
```
````

Alternatively, we could use the single-line format:

````suggestion
or using Docker Compose:

```yaml
services:
  memcached:
    image: %%IMAGE%%
    command: --conn-limit=1024 --memory-limit=64 --threads=4
```
````

LaurentGoderre: IMO, the list is better if you use more options

tianon (approved):

jaydorsey:
Nice! Thank you for doing this. I didn't even realize the docs were part of the repo. This is looks good, it's on the live site already: https://hub.docker.com/_/memcached

I appreciate that you put the command in there to get the configuration options, in addition to the docker compose info.

PR #2497, opened 2024-10-17
Update README.md Tishanth

TishanthP:
Comments can help readers better understand the meaning you are trying to convey through this code, and this feels more necessary in the table of contents, so this would be my suggestion.


tianon:
I'm sorry, but this PR as-is just looks like spam. :confused:

If you want to know more about an entry in the table of contents, you should click on it and read the appropriate section of the document that it's a table of contents for.

PR #2501, opened 2024-11-11
docs: Update MySQL/Percona random root password generator

mbrodala:
See: https://github.com/docker-library/mysql/commit/24e6931ec8187930dfad5ef2d205a9874f86ca4c

### DIFF
--- a/mysql/content.md
+++ b/mysql/content.md
@@ -106,7 +106,7 @@ This is an optional variable. Set to a non-empty value, like `yes`, to allow the
 
 ### `MYSQL_RANDOM_ROOT_PASSWORD`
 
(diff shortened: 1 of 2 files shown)


tianon:
Percona actually uses `pwmake` :sweat_smile:

https://github.com/percona/percona-docker/blob/3670d1afb11ec6f32a5adba432a002ab4de7fc48/percona-server-8.0/ps-entry.sh#L149

mbrodala:
Nice hint, I don't use it personally and didn't check. I'd question that the same should be used everywhere but for now I'll update the PR at least.

yosifkit (approved):
Rebased to drop out `README.md` changes since those are automatically committed by a bot.

LGTM

yosifkit:
Thanks!

PR #2504, opened 2024-11-22
Run update.sh

thomasboussekey:


thomasboussekey:
Hi,

I've run this command

```console
sudo ./update.sh postgres
```

... thinking that it will refresh Postgres image with the latest ones published yesterday

https://www.postgresql.org/about/news/postgresql-172-166-1510-1415-1318-and-1222-released-2965/

But the diff doesn't contain pointers to the new versions

Can someone help me?

I've lost myself in `./update.sh` with namespace concept

tianon:
I'm actually confused what problem you're trying to solve :sweat_smile:   can you elaborate?
(this is something that normally happens automatically via @docker-library-bot)

tianon:
See https://github.com/docker-library/docs/commit/b7155fe1613928edfd1ac4032f2d986ececa4eff#diff-bd9a13633673056216b5a31e1c0bf22a1f2a98a7da54e5b0eeef54d1ea90e8c2 for a concrete example of that, which maybe is the one you're looking for?

(It needed https://github.com/docker-library/official-images/pull/17951 to merge first, which is probably why it didn't update anything for you.)

thomasboussekey:
Thanks @tianon for the explanation!

I understand how it works, so my manual PR is USELESS, I close it

PR #2505, opened 2024-11-24
Improve PGDATA warning about anonymous volumes

mikelococo:
The people who most need to see this warning will struggle to understand the meaning of the pre-existing language.

This updated warning:

- Refers to anonymous volumes by their correct name.
- Describes where the mystery mount is coming from.
- States the required behavior quite plainly, you must mount at the correct location.

For additional context see:
https://stackoverflow.com/a/76217607

### DIFF
--- a/postgres/content.md
+++ b/postgres/content.md
@@ -96,7 +96,7 @@ See the PostgreSQL documentation on [`pg_hba.conf`](https://www.postgresql.org/d
 
 ### `PGDATA`
 
-> **Important Note:** when mounting a volume to `/var/lib/postgresql`, the `/var/lib/postgresql/data` path is a local volume from the container runtime, thus data is not persisted on the mounted volume.
+> **Important Note:** Mount the data volume at `/var/lib/postgresql/data` and not at `/var/lib/postgresql`. The Dockerfile that builds the image declares a volume at `/var/lib/postgresql/data` and if no data volume is mounted at that path then the container runtime will automatically create an [anonymous volume](https://docs.docker.com/engine/storage/#volumes) that is not reused across container re-creations. Data will be written to the anonymous volume rather than your intended data volume and WILL NOT PERSIST when the container is deleted and re-created.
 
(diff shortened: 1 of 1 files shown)


LaurentGoderre:
I would put the more important part first. The important info is about persisting the data so that should be among the first thing in the sentence.

postgres/content.md
```diff
@@ -96,7 +96,7 @@ See the PostgreSQL documentation on [`pg_hba.conf`](https://www.postgresql.org/d
 
 ### `PGDATA`
 
-> **Important Note:** when mounting a volume to `/var/lib/postgresql`, the `/var/lib/postgresql/data` path is a local volume from the container runtime, thus data is not persisted on the mounted volume.
+> **Important Note:** Mount the data volume at `/var/lib/postgresql/data` and not at `/var/lib/postgresql`. The Dockerfile that builds the image declares a volume at `/var/lib/postgresql/data` and if no data volume is mounted at that path then the container runtime will automatically create an [anonymous volume](https://docs.docker.com/engine/storage/#volumes) that is not reused across container re-creations. Data will be written to the anonymous volume rather than your intended data volume and WILL NOT PERSIST when the container is deleted and re-created.
```
mikelococo, line 99:
```suggestion
> **Important Note:** Mount the data volume at `/var/lib/postgresql/data` and not at `/var/lib/postgresql`, because mounts at the latter path WILL NOT PERSIST database data when the container is re-created. The Dockerfile that builds the image declares a volume at `/var/lib/postgresql/data` and if no data volume is mounted at that path then the container runtime will automatically create an [anonymous volume](https://docs.docker.com/engine/storage/#volumes) that is not reused across container re-creations. Data will be written to the anonymous volume rather than your intended data volume and won't persist when the container is deleted and re-created.
```

mikelococo: @LaurentGoderre How do you feel about this rephrasing?

LaurentGoderre: I think that's good!

mikelococo:
Committed, I think we're ready for a merge.

LaurentGoderre (approved):

tianon (approved):

PR #2509, opened 2024-11-28
Add dark version of the wordpress logo

LaurentGoderre:
Not this won't affect the display on Hub until multiple logos are supported in the API and the push script is updated


tianon:
Hmm, don't SVGs have ways to accomplish this without a separate file? (ie one SVG that changes based on the viewing mode directly)

We currently have a number of repos that have SVG logos, but we convert them before uploading them to Docker Hub -- do you think that team might be interested in supporting SVGs directly, especially as an alternative to supporting _two_ logo uploads per repository?

LaurentGoderre:
In theory it is very easy to do this with SVG butbthe complexity is that not all of the logo are in that format and we would need different way to add the logos for the SVG.

tianon:
I imagine it's probably easier on the Hub side to update the existing endpoints to accept SVG upload in addition to / in place of PNG and serve them appropriately throughout the platform than it is to implement a _second_ upload flow for a "dark mode" version of the repository logo _and_ add some kind of conditional logic everywhere to use it appropriately (which might be necessary even with the SVG solution, but at least then it's still "just" a single image file/upload, and the conditional logic is more "if SVG" and can be server-side instead of "if dark mode" and have to be dependent on the client-side).

Then the only change here would be that if a given repository has `logo.svg`, we upload that directly instead of `logo[-120].png` (with an appropriate `content-type`)

PR #2511, opened 2024-12-02
SonarQube rebrand

davividal:
Since we are rebranding SonarQube with a new logo, naming, and offerings, we need to update the Docker documentation to properly reflect that.

Can these commits be squashed or must I do that on my side before?


yosifkit:
> Can these commits be squashed or must I do that on my side before?

We don't mind doing a "squash and merge", but if you really want to squash and force push, that is fine too.

There are a couple failing tests that need addressed: 
- `README-short.txt` is a little bit long (Docker Hub only allows 100 characters)
- `sonarqube/README.md` should not be in the commits as it will be re-generated automatically after merge.

davividal:
Thanks for the answer, @yosifkit.

I addressed your comments and squashed the commits.

yosifkit (approved):

PR #2512, opened 2024-12-03
Correct IBM Semeru Tag info in Example 

jayasg12:
The example in https://hub.docker.com/_/ibm-semeru-runtimes has incorrect tag, which is corrected as part of this update.


jayasg12:
@tianon There is a small correction to the content.md file , the semeru tag  was not correct in example in the description which is now corrected. Request to review the change. Thanks !!

yosifkit (approved):

PR #2514, opened 2024-12-06
docs: update deprecated/outdated compose syntax and filenames

dvdksn:
Not an exhaustive update (didn't want to touch the `stack.yml` stuff yet, could do in a follow-up) but updated references to old `docker-compose` binary and changed the filename from `docker-compose.yml` to `compose.yaml`, where it made sense to do so.

PR #2516, opened 2024-12-14
fix: remove obsolete docker compose file version

ghorbani-ali:
Based on [this reference](https://docs.docker.com/reference/compose-file/version-and-name/), there is no need to compose file version.

### DIFF
--- a/wordpress/stack.yml
+++ b/wordpress/stack.yml
@@ -1,5 +1,3 @@
-version: '3.1'
-
 services:
 
   wordpress:


ghorbani-ali:
Hi @LaurentGoderre , Thanks for the tip. I updated my PR.

tianon:
This was still here because this is actually a Docker Swarm "Stack" file, from back when it and Compose were being co-designed intentionally. :see_no_evil:

I think this really ought to have been part of a larger change (like what's discussed/hinted at in https://github.com/docker-library/docs/pull/2514), but that's more complicated to orchestrate/coordinate. :disappointed: :heart:

(IMO, dropping `version:` in Compose was a mistake, because now compatibility is harder to break over time, but :shrug:)

PR #2517, opened 2024-12-16
fix(README): Fix indent problem for environments keyword

blankoworld:
As I use docker-compose.yml example, I saw an indentation problem.

Is that OK?


blankoworld:
> Sorry for the delay! Thanks

Thank you for the merge 😎
It was not critical as changes 😉.
It’s ok for me 👌

PR #2521, opened 2025-01-02
Add PHP PIE for installing extensions to docs

jnoordsij:
Based on the discussion in https://github.com/docker-library/php/issues/1554, PHP PIE is not likely to be included as part of the official PHP image. However, as it aims to replace PECL as the official way for installing PHP extensions, consumers of the PHP image should probably be informed on using this tool.

## To Do/Discuss
- [x] Decide on whether or not to suggest an explicit way of installing PIE in the example, or just leave the reference to the docs as I did now
  - In particular the core extension `zip` is required, so installation is not entirely trivial
- [ ] Check whether adding the deprecation notice for PECL is not a bit too eager
- [x] Pending what happens with https://github.com/php/pie/issues/157, adding something like `rm -rf ~/.composer ~/.pie` to the suggested installation steps might be relevant to reduce the size of the built image
- [ ] Wait for further upstream to become more stable
  - See also https://github.com/php/pie/blob/main/docs/supported-extensions.md for a list of supported extensions

### DIFF
--- a/php/content.md
+++ b/php/content.md
@@ -78,8 +78,14 @@ The latest recommended way of installing PHP extensions is through [PIE](https:/
 ```dockerfile
 FROM %%IMAGE%%:8.2-cli
 
-# Install PIE here (see https://github.com/php/pie/blob/main/docs/usage.md)
+# Install PIE here; e.g. through using Docker
+RUN export DEBIAN_FRONTEND="noninteractive"; \
+    set -eux; \
+    apt-get update; apt-get install -y --no-install-recommends unzip; \
+    rm -rf /var/lib/apt/lists/*
+COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
 
+# Use PIE to install extensions
 RUN pie install phpredis/phpredis:^6.1 \
 	&& pie install xdebug/xdebug:^3.4
 ```


jnoordsij:
@asgrim (mentioning you as you've been helping out with https://github.com/php/pie/issues/151): I've created this as a way to (hopefully) notify (new) Docker image users of PIE. However this is just a first stab from my side and would very much like some upstream feedback if you can provide any (or know of someone who can).

In particular regarding the stability/status I'm wondering if putting this in right now or at some later moment is more suitable, when it is more stabilized.

php/content.md
```diff
@@ -71,8 +71,23 @@ See also ["Dockerizing Compiled Software"](https://tianon.xyz/post/2017/12/26/do
...
 Some extensions are compiled by default. This depends on the PHP version you are using. Run `php -m` in the container to get a list for your specific version.
 
+### PIE extensions
+
+The latest recommended way of installing PHP extensions is through [PIE](https://github.com/php/pie). To install a PIE-compatible extension, use `pie install` to download, compile and enable it.
+
+```dockerfile
+FROM %%IMAGE%%:8.2-cli
+
+# Install PIE here (see https://github.com/php/pie/blob/main/docs/usage.md)
```
asgrim, line 81: This could probably do:

```suggestion
# Install PIE here (see https://github.com/php/pie/blob/main/docs/usage.md)
COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
```

perhaps? to demo a real working example...

asgrim: This is a working example:

```
FROM php:8.4-cli                                                                                                                            
                                                                                                                                            
RUN apt-get update \
    && apt-get install -y libzip-dev zip \
    && docker-php-ext-install zip

COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie

RUN pie install xdebug/xdebug
```

Ironically, you can't use PIE to install the `zip` extension as it depends on it (see php/pie#168)

jnoordsij: Thanks a lot for looking into this so quickly! I too do prefer the `COPY --from=ghcr.io/php/pie:bin` call for installing this. Given that [multi-stage builds](https://github.com/docker-library/faq?tab=readme-ov-file#multi-stage-builds) are officially supported nowadays this probably should be fine. Given that is an example and official upstream source referencing a non-official image is probably also fine, but I'll wait for maintainer feedback on that.

I also just figured the `zip` extension requirement makes it somewhat cumbersome to use when paired with Docker compared to PECL working out of the box. But to solve that it seems that https://github.com/php/pie/issues/133 probably would also need to be addressed.

tianon:
Thanks for getting this started!  I'm not outright _opposed_ to including it, but I do have some reservations.

My primary reservation comes from having lived through the saga of https://github.com/golang/dep, which was a pretty "official" project, but still a WIP, and ended up completely abandoned, so I hesitate to give an explicit recommendation like this to users.  Combine that with the notice at the top of https://github.com/php/pie ("PIE is still in early development. We welcome you to try this out, but please exercise caution when using PIE, as things may go wrong!") and I think we're probably still a little early here.  I'm happy to be convinced otherwise, but I want to make sure I'm being clear in where (and why) I'm feeling reserved. :smile:

php/content.md
asgrim: I've just released [PIE 0.5.0](https://github.com/php/pie/releases/tag/0.5.0) which removes the `ext-zip` dependency. However, just like using Composer, you'll need some kind of way to extract zip; and installing `unzip` is probably the best way. Here's [a working example](https://github.com/php/pie/blob/main/docs/usage.md#example-of-pie-working-in-a-dockerfile) with the latest version; just tested and verified it builds:

```Dockerfile
FROM php:8.4-cli

# Add the `unzip` package which PIE uses to extract .zip files
RUN export DEBIAN_FRONTEND="noninteractive"; \
    set -eux; \
    apt-get update; apt-get install -y --no-install-recommends unzip; \
    rm -rf /var/lib/apt/lists/*

# Copy the pie.phar from the latest `:bin` release
COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie

# Use PIE to install an extension...
RUN pie install asgrim/example-pie-extension
```

asgrim:
We did release PIE 1.0.0 now btw :)

php/content.md
```diff
@@ -71,8 +71,29 @@ See also ["Dockerizing Compiled Software"](https://tianon.xyz/post/2017/12/26/do
...
+
+```dockerfile
+FROM %%IMAGE%%:8.2-cli
+
+# Install PIE here; e.g. through using Docker
+RUN export DEBIAN_FRONTEND="noninteractive"; \
+    set -eux; \
+    apt-get update; apt-get install -y --no-install-recommends unzip; \
+    rm -rf /var/lib/apt/lists/*
+COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
```
TimWolla, line 86: Prefer the `RUN --mount` from https://github.com/docker-library/php/issues/1554#issuecomment-3722916430 to not unnecessarily include PIE in the final image.

php/content.md
```diff
@@ -71,8 +71,29 @@ See also ["Dockerizing Compiled Software"](https://tianon.xyz/post/2017/12/26/do
...
+
+# Use PIE to install extensions
+RUN pie install phpredis/phpredis:^6.1 \
+	&& pie install xdebug/xdebug:^3.4
+```
+
 ### PECL extensions
 
+⚠️ Note: Installation of PEAR and thus also PECL with PHP is deprecated and thus subject to removal in a future PHP version. See also https://github.com/php/php-src/commit/e93d6d97aab7a5de1f7b8dc750ca9d08214de8c4.
+
```
TimWolla, line 96: I think it would be okay to drop the PECL section entirely. See also: docker-library/php#1654.

php/content.md
TimWolla: Perhaps the correct solution is just linking towards: https://github.com/php/pie/blob/1.4.x/docs/usage.md#example-of-pie-working-in-a-dockerfile

PR #2524, opened 2025-01-10
Update odoo docs to fix printing/pdf

adnoh:
add information about a required system parameter (report.url ) so that pdf generation inside the container works like expected


tianon:
I'll admit I'm not familiar with Odoo enough to know whether this is appropriate/necessary, but let me cc the image maintainer @d-fence :bow:

adnoh:
@d-fence Any opinion on that? It did cost me a long time as I followed the docs and it was not mentioned there anywhere.

d-fence:
I don't see the issue you are trying to fix.
The pdf generation works out of the box with the actual image.

PR #2534, opened 2025-01-29
Update README.md

andanders:
Correct filename


LaurentGoderre:
This file is autogenerated, could you make the change on this file instead? https://github.com/docker-library/docs/blob/master/ros/content.md

PR #2535, opened 2025-01-30
Update content.md

andanders:
correct filename

### DIFF
--- a/ros/content.md
+++ b/ros/content.md
@@ -22,7 +22,7 @@ RUN apt-get update && apt-get install -y \
     rm -rf /var/lib/apt/lists/*
 
 # launch ros package
-CMD ["ros2", "launch", "demo_nodes_cpp", "talker_listener_launch.py"]
+CMD ["ros2", "launch", "demo_nodes_cpp", "talker_listener.launch.py"]
 ```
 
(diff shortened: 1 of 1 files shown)


tianon:
In the future, please update your existing pull request instead of creating a new one (https://github.com/docker-library/docs/pull/2534; https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request#making-changes-to-files-in-your-pull-request).

cc `ros` image maintainer @tfoote -- this looks correct to me, but I'd love if you could confirm :bow:

tfoote (commented):
This is rolling back to the old syntax. We switched to the `_launch` format 2 years ago: https://github.com/ros2/demos/pull/588 This should be the correct syntax for all versions after Humble. 

We could add a side comment for Humble and older to use the alternative syntax, and support that until Humble goes EOL. But I would be inclined to keep this as targeted for the default version. 

What's your usage @andanders ?

andanders:
My use-case: I am coming to the README as a beginner in using docker and ros1_bridge, and the entire readme focuses on using foxy. So the examples fail, which gives a poor experience for newcomers. If the syntax is correct from Humble and onwards, then should the README be changed to use docker images for humble instead of foxy?

Line 16, 47 and 240 in the readme.

tfoote:
Yeah, I think it would be best for us to move the docs forward to the latest LTS release so replace foxy with jazzy

PR #2536, opened 2025-02-04
fix: typo on version name

zisuu:
The examples are showing JDK version 21 but the description text above states JDK version 11

### DIFF
--- a/eclipse-temurin/content.md
+++ b/eclipse-temurin/content.md
@@ -54,7 +54,7 @@ ENV PATH="${JAVA_HOME}/bin:${PATH}"
 
 ### Creating a JRE using jlink
 
-On OpenJDK 11+, a JRE can be generated using `jlink`, see the following Dockerfile:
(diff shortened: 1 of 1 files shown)


tianon:
(per the failing CI, I've dropped the commit touching `README.md` as it's auto-generated)

zisuu:
> (per the failing CI, I've dropped the commit touching `README.md` as it's auto-generated)

thanks, sorry was not aware of that

tianon (approved):
Thanks!

(FYI @gdams @sxa - I'm merging because it looks correct to me :+1:)

gdams:
I’m actually not sure about this change. Admittedly the wording isn’t great but what we’re trying to say is that jlink is supported on OpenJDK 11 and newer.

tianon:
Oh, good point, sorry!  We can revert if you want, or maybe you have some new wording you'd like to add?

PR #2540, opened 2025-02-07
Remove unnecessary "go mod verify"

### LINKED ISSUE #2538: golang: why "go mod verify"?
In #2110 the example steps for golang was updated to include among other things this step:
```
RUN go mod download && go mod verify
```
My question is why the `go mod verify` step is present. Reading https://go.dev/ref/mod#go-mod-verify it leads with:
> go mod verify checks that dependencies of the [main module](https://go.dev/ref/mod#glos-main-module) stored in the [module cache](https://go.dev/ref/mod#glos-module-cache) have not been modified since they were downloaded.

Given that we perform `go mod download` just before, why would anything have been modified directly after the download has finished?

eest:
Fixes #2538

### DIFF
--- a/golang/content.md
+++ b/golang/content.md
@@ -21,7 +21,7 @@ WORKDIR /usr/src/app
 
 # pre-copy/cache go.mod for pre-downloading dependencies and only redownloading them in subsequent builds if they change
 COPY go.mod go.sum ./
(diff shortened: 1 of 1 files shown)


yosifkit (approved):
Thanks!

PR #2476, opened 2024-07-26
Add `adminer` deprecation notice

tianon:
See https://www.youtube.com/watch?v=OrOtiu_nfHE&lc=Ugy8pAL8wgAL3_iKkzZ4AaABAg.9pj_kQ2rkuw9pp813OyHha (youtube comment), https://github.com/TimWolla/docker-adminer/issues/147, and https://github.com/vrana/adminer -- upstream is not actively maintained (and has not been for years now).  We should reflect/document that status accurately on Docker Hub.

PR #2544, opened 2025-02-20
Remove deprecation notice

vrana:
The project is now actively maintained.

### DIFF
--- a/adminer/README-short.txt
+++ b/adminer/README-short.txt
@@ -1,1 +1,1 @@
-DEPRECATED; Database management in a single PHP file.
+Database management in a single PHP file.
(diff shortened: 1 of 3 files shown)


tianon (approved):
Nice, thank you!  Welcome back. :smile:

I've force pushed to remove the touching of `README.md` as that file is generated (and will be updated by @docker-library-bot post-merge).

Normally I'd rather see the image updated before we merge this, but since we emptied out the `library/adminer` file over in https://github.com/docker-library/official-images, it has no active tags anyhow, so this is harmless to merge before that happens. :+1:

tianon:
(just to link the threads, this is essentially a revert of https://github.com/docker-library/docs/pull/2476 :+1:)

vrana:
Thanks.

TimWolla:
> Normally I'd rather see the image updated before we merge this,

Now in docker-library/official-images#18524.

PR #2548, opened 2025-03-01
Add a blurb about Debian being explicitly and intentionally minimal

tianon:
Also, add more notes and a few links to help explain what "minimal" means in this context (and who decides it).

Closes https://github.com/debuerreotype/docker-debian-artifacts/issues/15

@paultag if/when you have a moment (it's not terribly long), would you be willing to review this too? :innocent: :heart: :bow:

### DIFF
--- a/debian/content.md
+++ b/debian/content.md
@@ -8,6 +8,8 @@ Debian is an operating system which is composed primarily of free and open-sourc
 
 # About this image
 
+The images in this repository are intended to be as minimal as possible (because of the immutable/layered nature of containers, it's much easier to add than it is to remove). More specifically, they're built from [the "minbase" variant](https://manpages.debian.org/stable/debootstrap/debootstrap.8.en.html#variant=minbase_buildd_fakechroot), which only installs "required" packages, and thus creates the smallest possible footprint that is still "Debian" (as defined/managed by [the Release and FTP teams](https://www.debian.org/intro/organization#distribution) within the project).
+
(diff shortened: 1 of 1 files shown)


tianon:
I contemplated adding an explicit example of `buildpack-deps:bookworm-curl` for folks who want `ca-certificates` and/or `wget`/`curl`, but I'm torn on whether that's worth including here.

debian/content.md
```diff
@@ -8,6 +8,8 @@ Debian is an operating system which is composed primarily of free and open-sourc
 
 # About this image
 
+The images in this repository are intended to be as minimal as possible (because of the immutable/layered nature of containers, it's much easier to add than it is to remove). More specifically, they're built from [the "minbase" variant](https://manpages.debian.org/stable/debootstrap/debootstrap.8.en.html#variant=minbase_buildd_fakechroot), which only installs "required" packages, and thus creates the smallest possible footprint that is still "Debian" (as defined/managed by [the Release and FTP teams](https://www.debian.org/intro/organization#distribution) within the project).
```
justinmchase, line 11: This may add a nice highlight effect to really call it out.

```suggestion
> [!NOTE]
> The images in this repository are intended to be as minimal as possible (because of the
immutable/layered nature of containers, it's much easier to add than it is to remove). More specifically, they're built from [the "minbase" variant](https://manpages.debian.org/stable/debootstrap/debootstrap.8.en.html#variant=minbase_buildd_fakechroot), which only installs "required" packages, and thus creates the smallest possible footprint that is still "Debian" (as defined/managed by [the Release and FTP teams](https://www.debian.org/intro/organization#distribution) within the project).
```

paultag (commented):
Looks great to me!

debian/content.md
tianon: Unfortunately the target of this is Docker Hub, not GitHub, and that's a GitHub-specific markup.

PR #2486, opened 2024-08-29
docs: update readme to indicate the port binding

felipementel:
In order to make it easier for developers who want to run SonarQube in a container, I think it is worth indicating the port binding when running the docker run command.

Our exemple are using the tag 10.6.0-community

PR #2552, opened 2025-03-06
docs: update readme from mariadb with port binding, tags, remove double spaces and informations

felipementel:
This pull request includes updates to the `mariadb` documentation and configuration files to improve clarity and compatibility, as well as to specify versions for certain images.

Documentation improvements:

* [`mariadb/content.md`](diffhunk://#diff-e7ab7ccc549c315a1512ce8cfc184a2be06cf7011b83d7d9a8bc87f2b5b28702L11-R12): Added an important note about the `%%IMAGE%%` tags to highlight the `latest` and `lts` versions.
* [`mariadb/content.md`](diffhunk://#diff-e7ab7ccc549c315a1512ce8cfc184a2be06cf7011b83d7d9a8bc87f2b5b28702L39-R40): Updated the Docker compose command to use `docker compose` instead of `docker-compose` for consistency with the latest Docker CLI.
* [`mariadb/content.md`](diffhunk://#diff-e7ab7ccc549c315a1512ce8cfc184a2be06cf7011b83d7d9a8bc87f2b5b28702R49-R57): Added a section on port binding to explain how to expose the container port to the host port using the `-p` argument in the `docker run` command.

Configuration updates:

* [`mariadb/stack.yml`](diffhunk://#diff-86280722e23a1fe81d693e9f423077a4fb84ff951eea7663544c75f54960ba1fL2-R14): Specified the image versions for `mariadb` and `adminer` to ensure consistency and compatibility. The `mariadb` image is updated to `11.7-ubi` and `adminer` to `4.17.1`.
* Remove docker compose version on file header - the docker compose version is ***obsolete***

### DIFF
--- a/mariadb/content.md
+++ b/mariadb/content.md
@@ -8,7 +8,8 @@ The intent is also to maintain high compatibility with MySQL, ensuring a library
 
 # How to use this image
 
-The %%IMAGE%% has a number of tags, and of note is `latest`, as the latest stable version, and `lts`, as the last long term support release.
+> [!IMPORTANT]
+> The %%IMAGE%% has a number of tags, and of note is `latest`, as the latest stable version, and `lts`, as the last long term support release.
 
 ## Running the container
 
@@ -17,26 +18,26 @@ The %%IMAGE%% has a number of tags, and of note is `latest`, as the latest stabl
 The environment variables required to use this image involves the setting of the root user password:
 
 ```console
-$ docker run --detach --name some-%%REPO%% --env MARIADB_ROOT_PASSWORD=my-secret-pw  %%IMAGE%%:latest
+$ docker run --detach --name some-%%REPO%% --env MARIADB_ROOT_PASSWORD=my-secret-pw %%IMAGE%%:latest
 ```
 
 or:
 
 ```console
-$ docker run --detach --name some-%%REPO%% --env MARIADB_ALLOW_EMPTY_ROOT_PASSWORD=1  %%IMAGE%%:latest
+$ docker run --detach --name some-%%REPO%% --env MARIADB_ALLOW_EMPTY_ROOT_PASSWORD=1 %%IMAGE%%:latest
 ```
 
 or:
 
 ```console
-$ docker run --detach --name some-%%REPO%% --env MARIADB_RANDOM_ROOT_PASSWORD=1  %%IMAGE%%:latest
+$ docker run --detach --name some-%%REPO%% --env MARIADB_RANDOM_ROOT_PASSWORD=1 %%IMAGE%%:latest
 ```
 
 ... where the container logs will contain the generated root password.
 
 ## %%STACK%%
 
-Run `docker stack deploy -c stack.yml %%REPO%%` (or `docker-compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
+Run `docker stack deploy -c stack.yml %%REPO%%` (or `docker compose -f stack.yml up`), wait for it to initialize completely, and visit `http://swarm-ip:8080`, `http://localhost:8080`, or `http://host-ip:8080` (as appropriate).
 
 ### Start a `%%IMAGE%%` server instance with user, password and database
 
@@ -45,7 +46,15 @@ Starting a MariaDB instance with a user, password, and a database:
 ```console
 $ docker run --detach --name some-%%REPO%% --env MARIADB_USER=example-user --env MARIADB_PASSWORD=my_cool_secret --env MARIADB_DATABASE=exmple-database --env MARIADB_ROOT_PASSWORD=my-secret-pw  %%IMAGE%%:latest
 ```
+#### Configuration
 
+##### Port binding
+
+By default, the database running within the container will listen on port 3306. You can expose the container port 3306 to the host port 3306 with the `-p 3306:3306` argument to `docker run`, like the command below:
+
+```console
+$ docker run --name some-mariadb -p 3306:3306 mariadb:latest
+```
 ### Start a `%%IMAGE%%` server instance in a network
 
 As applications talk to MariaDB, MariaDB needs to start in the same network as the application:
--- a/mariadb/stack.yml
+++ b/mariadb/stack.yml
@@ -3,7 +3,7 @@
 services:
 
   db:
-    image: mariadb:latest
+    image: mariadb:11.7-ubi
     restart: always
     environment:
       MARIADB_ROOT_PASSWORD: example
--- a/mariadb/stack.yml
+++ b/mariadb/stack.yml
@@ -1,16 +1,17 @@
 # Use root/example as user/password credentials
-version: '3.1'
 
 services:
 
   db:
-    image: mariadb
+    image: mariadb:latest
     restart: always
     environment:
       MARIADB_ROOT_PASSWORD: example
+    ports:
+      - 3306:3306
 
   adminer:
-    image: adminer
+    image: adminer:4.17.1
     restart: always
     ports:
       - 8080:8080


mariadb/content.md
```diff
@@ -8,42 +8,53 @@ The intent is also to maintain high compatibility with MySQL, ensuring a library
 
 # How to use this image
 
-The %%IMAGE%% has a number of tags, and of note is `latest`, as the latest stable version, and `lts`, as the last long term support release.
+> [!IMPORTANT]
+> The %%IMAGE%% has a number of tags, and of note is `latest`, as the latest stable version, and `lts`, as the last long term support release.
```
tianon, line 12: The target for this description is Docker Hub, not GitHub, so this GitHub-specific syntax probably won't land well.

mariadb/content.md
```diff
@@ -8,42 +8,53 @@ The intent is also to maintain high compatibility with MySQL, ensuring a library
...
 
 ## Running the container
 
+### Configuration
+
+#### Port binding
+
+By default, the database running within the container will listen on port 3306. You can expose the container port 3306 to the host port 3306 with the `-p 3306:3306` argument to `docker run`, like the command below:
```
tianon, line 20: More than simply exposing to the host, this exposes to the entire host network, which in many users' cases is (intentionally or unintentionally) the entire internet because they're often on a VPC with an internet-facing IP, so this ends up putting their database directly on the public internet (and even with authentication, that's not a super great idea, so we try to avoid recommending it).

mariadb/stack.yml
```diff
@@ -1,16 +1,17 @@
 # Use root/example as user/password credentials
-version: '3.1'
```
tianon, line 2: If we drop this, we also need to drop the references to `docker stack deploy`; see also https://github.com/docker-library/docs/pull/2516#issuecomment-2574035913

mariadb/stack.yml
```diff
@@ -1,16 +1,17 @@
...
+    image: mariadb:11.7-ubi
     restart: always
     environment:
       MARIADB_ROOT_PASSWORD: example
+    ports:
+      - 3306:3306
 
   adminer:
-    image: adminer
+    image: adminer:4.17.1
```
tianon, line 14: These really specific version numbers are likely to bitrot quickly, so we don't typically suggest being so particular unless it's actually necessary to be.

(Also, see the prior note about exposed database ports.)

tianon (commented):
I've left a few notes :+1:

I'll also explicitly CC the `mariadb` image maintainers: @grooverdan @dbart @fauust

mariadb/content.md
grooverdan: Does 'last' -> 'latest' make more sense? I'm hoping there are going to be more LTS releases :-)

mariadb/content.md
grooverdan: should numbers be different on the entire host network side to highlight the differences between the two numbers?

As @tianon said, highlighting the risks with this block would be good.

mariadb/stack.yml
grooverdan: especially with 11.7 as a short term support release. Perhaps use a `lts` /`lts-ubi` tag and I guess something for adminer. The purpose of compose files was self contained, so especially with something link adminer, exposing ports seems unnecessary.

On env variables including something like `MARIADB_AUTO_UPGRADE=1` would smooth the upgrade paths.

I wouldn't mind seeing the healthcheck and service dependency like https://mariadb.com/kb/en/using-healthcheck-sh/#compose-file-example here as a practically of showing of getting the database up before the services that connect to it (and might fail if they can't connect).

mariadb/content.md
```diff
@@ -63,7 +74,7 @@ $ docker run --detach --network some-network --name some-application --env APP_D
 The following command starts another `%%IMAGE%%` container instance and runs the `mariadb` command line client against your original `%%IMAGE%%` container, allowing you to execute SQL statements against your database instance:
 
 ```console
-$ docker run -it --network some-network --rm %%IMAGE%% mariadb -hsome-%%REPO%% -uexample-user -p
+$ docker run -it --network some-network --rm %%IMAGE%% mariadb -h some-%%REPO%% -u example-user -p example
```
grooverdan, line 77: for weird historical reasons (I choose not to look at) there isn't a space between `-p` and the password "example". As you worked out, without a password it just prompts for the password. Isn't it safer to not have a password on a command? If its scripting chances are you don't need the `-it` (though would need for `mariadb` a `-e` or something to execute.

So some additional clarity why this change perhaps with a note on constraints comments.

grooverdan (commented):
thanks for the cleanup.

mariadb/content.md
felipementel: @tianon Ok, I'll remove it

@grooverdan it was written last, I just changed the formatting to put IMPORTANT. If you prefer, I can change like you wrote.

grooverdan: I'd appreciate it if you did. Thanks, it was probably my wording mistake.

mariadb/content.md
felipementel: > More than simply exposing to the host, this exposes to the entire host network, which in many users' cases is (intentionally or unintentionally) the entire internet because they're often on a VPC with an internet-facing IP, so this ends up putting their database directly on the public internet (and even with authentication, that's not a super great idea, so we try to avoid recommending it).

I understand your point and we can put this warning so that users are careful. On the other hand, there are people who want to run locally and connect to some graphical interface, as an administrator. Since some people don't know the default port, it might be interesting to share the default port.
See what you think is best for the community, and I'll remove the part, adjust the text to put a warning about exporting the database, or remove the whole part.

> should numbers be different on the entire host network side to highlight the differences between the two numbers?
> 
> As @tianon said, highlighting the risks with this block would be good.

Whenever I run local containers, I always set different local ports, precisely so that I can make my configuration explicit. Let's discuss in this topic what would be the best way to put this in the documentation.

mariadb/stack.yml
felipementel: > If we drop this, we also need to drop the references to `docker stack deploy`; see also [#2516 (comment)](https://github.com/docker-library/docs/pull/2516#issuecomment-2574035913)

https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete

mariadb/stack.yml
felipementel: > These really specific version numbers are likely to bitrot quickly, so we don't typically suggest being so particular unless it's actually necessary to be.
> 
> (Also, see the prior note about exposed database ports.)

updated.



> especially with 11.7 as a short term support release. Perhaps use a `lts` /`lts-ubi` tag and I guess something for adminer. The purpose of compose files was self contained, so especially with something link adminer, exposing ports seems unnecessary.
> 
> On env variables including something like `MARIADB_AUTO_UPGRADE=1` would smooth the upgrade paths.
> 
> I wouldn't mind seeing the healthcheck and service dependency like https://mariadb.com/kb/en/using-healthcheck-sh/#compose-file-example here as a practically of showing of getting the database up before the services that connect to it (and might fail if they can't connect). (edit: assuming `docker stack` compatibility).

I like the idea.

I think another section, with this more complete example, would be better. Then we can have a more basic section and a more advanced one.

mariadb/content.md
felipementel: @grooverdan I removed the -p and left the user with just the name example-user.

felipementel:
@tianon, @grooverdan 
There are other points to yours check?

felipementel (commented):
@tianon @grooverdan 

I did the adjusts. Can you verify it?

felipementel (commented):
Do I need to make any further adjustments for analysis and approval?

mariadb/stack.yml
```diff
@@ -1,16 +1,15 @@
...
   db:
-    image: mariadb
+    image: mariadb:11.7
     restart: always
     environment:
       MARIADB_ROOT_PASSWORD: example
 
   adminer:
-    image: adminer
+    image: adminer:4.17.1
```
yosifkit, line 12: Adminer upstream has been [very active recently](https://github.com/vrana/adminer/releases/) with many new version releases added to [Docker Official Images](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fadminer+is%3Aclosed). I don't know how long the very specific `4.17.1` image will continue to be "maintained". While we will be very unlikely to delete the tag from Docker Hub, this version will bit rot fairly quickly. So, I suggest keeping this as `latest`.

grooverdan:
sorry for delay. I'll get to it next week.

felipementel:
@grooverdan 
What can I do to merge this pull request?

mariadb/stack.yml
tianon: This thread isn't resolved (references in `stack.yml` are still too-specific).

mariadb/content.md
grooverdan: A more verbose `-p` -> `--publish` and a non-0.0.0.0 host  to the example and/or a warning about potential internet expose is required.

mariadb/stack.yml
```diff
@@ -1,16 +1,15 @@
 # Use root/example as user/password credentials
-version: '3.1'
 
 services:
 
   db:
-    image: mariadb
+    image: mariadb:11.7
```
grooverdan, line 6: there's still specific versions here.

grooverdan: `MARIADB_AUTO_UPGRADE=1` isn't there in env

mariadb/stack.yml
```diff
@@ -1,16 +1,15 @@
...
     restart: always
     environment:
       MARIADB_ROOT_PASSWORD: example
 
   adminer:
-    image: adminer
+    image: adminer:5.0.1
```
grooverdan, line 12: while I do generally like the idea of pinning versions, on a UI is it important? With this versioned pinned the updates to this are going to be quite frequent - current adminer versions: https://github.com/docker-library/official-images/blob/master/library/adminer

mariadb/stack.yml
grooverdan: more explicitly:

https://docs.docker.com/engine/swarm/stack-deploy "The docker stack deploy command uses the legacy [Compose file version 3](https://docs.docker.com/reference/compose-file/legacy-versions/) format, used by Compose V1"

[https://docs.docker.com/compose/intro/history/ "](https://docs.docker.com/compose/intro/history/#introduction) (from image - Compose V1 / Compose file version  - "Used version top level element". So if you remove version, remove the `docker stack deploy` like @tianon  said.

grooverdan (commented):
previous comments remain unaddressed.

felipementel:
> previous comments remain unaddressed.

I update the both tags to ***latest***

felipementel:
> previous comments remain unaddressed.

Hi @grooverdan,
Can you verify it now?

mariadb/content.md
```diff
@@ -12,38 +12,48 @@ The %%IMAGE%% has a number of tags, and of note is `latest`, as the latest stabl
...
 
 ```console
-$ docker run --detach --name some-%%REPO%% --env MARIADB_RANDOM_ROOT_PASSWORD=1  %%IMAGE%%:latest
+$ docker run --detach --name some-%%REPO%% --env MARIADB_RANDOM_ROOT_PASSWORD=1 %%IMAGE%%:latest
 ```
 
 ... where the container logs will contain the generated root password.
 
-## %%STACK%%
+## %%Docker Compose%%
```
yosifkit, line 47: This has to be `%%STACK%%` for the templating to replace it with the content of [`.template-helpers/stack.md`](https://github.com/docker-library/docs/blob/c8fefb5ed65cec4609a4cc64b1154ffeb240d2fa/.template-helpers/stack.md) (inserting in `stack.yml` as well). If the content of `stack.md` is not correct, then the file would have to be copied into the `mariadb/` folder and modified.

Though looking at `update.sh`, we can just rename `mariadb/stack.yml` to `mariadb/docker-compose.yml` and change this to `%%COMPOSE%%` to get the older [compose template](https://github.com/docker-library/docs/blob/c8fefb5ed65cec4609a4cc64b1154ffeb240d2fa/.template-helpers/compose.md).

tianon:
(force pushed to squash all the commits into a single commit and rebase on top of https://github.com/docker-library/docs/pull/2566)

yosifkit (approved):

felipementel:
@yosifkit, @grooverdan, @tianon 

Thank you so much!

felipementel:
@tianon,

Can you help me earn a Pair Extraordinaire badge, like you did in PR https://github.com/docker-library/docs/pull/2486?

you can see the result here: https://github.com/felipementel?achievement=pair-extraordinaire&tab=achievements

PR #2554, opened 2025-03-17
Add a `golang:tip` variant section

tianon:
This describes [the "tip" variant](https://github.com/docker-library/golang/pull/531), but most notably provides a rough estimate for how often it's updated.

(https://github.com/docker-library/golang/issues/464#issuecomment-2730113332)


tianon:
I was hoping to find a good upstream "if you hear 'Go tip', here's what it means" kind of page/blurb/one-liner, but https://github.com/golang/build/blob/6383021611af0e07cbf0a60222e066662557c796/cmd/coordinator/internal/legacydash/build.go#L313-L314 is the closest I could find to such a thing. :sweat_smile:

yosifkit (approved):

Issue #2558, opened 2025-04-01
`php/variant-fpm.md` refers to non-official resources


TimWolla:
via php/php-src#18217 and https://www.reddit.com/r/PHP/comments/1joe1fq/phpfpm_with_latest_news_from_2011/.

A link to https://www.php.net/fpm would certainly be helpful.

PR #2559, opened 2025-04-01
Update php/variant-fpm.md

TimWolla:
Some opinionated suggestions.

Fixes docker-library/docs#2558

### DIFF
--- a/php/variant-fpm.md
+++ b/php/variant-fpm.md
@@ -1,12 +1,12 @@
 ## `%%IMAGE%%:<version>-fpm`
 
-This variant contains [PHP’s FastCGI Process Manager (FPM)](https://www.php.net/fpm), which is the recommended FastCGI implementation for PHP.
+This variant contains [PHP's FastCGI Process Manager (FPM)](https://www.php.net/fpm), which is the recommended FastCGI implementation for PHP.
 
 In order to use this image variant, some kind of reverse proxy (such as NGINX, Apache, or other tool which speaks the FastCGI protocol) will be required.
 
 Some potentially helpful resources:
 
--	[FPM’s Official Configuration Reference](https://www.php.net/manual/en/install.fpm.configuration.php)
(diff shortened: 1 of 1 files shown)


php/variant-fpm.md
```diff
@@ -1,14 +1,14 @@
 ## `%%IMAGE%%:<version>-fpm`
 
-This variant contains PHP-FPM, which is a FastCGI implementation for PHP. See [the PHP-FPM website](https://php-fpm.org/) for more information about PHP-FPM.
+This variant contains [PHP’s FastCGI Process Manager (FPM)](https://www.php.net/fpm), which is the recommended FastCGI implementation for PHP.
```
tianon, line 3:
```suggestion
This variant contains [PHP's FastCGI Process Manager (FPM)](https://www.php.net/fpm), which is the recommended FastCGI implementation for PHP.
```

tianon (approved):
Seems fine, although I'd rather use plain apostrophes than fancy ones for these docs :+1:

tianon (approved):
Nice, thank you!

TimWolla:
> although I'd rather use plain apostrophes than fancy ones for these docs 👍

Sure. Just a habit, because they are just too conveniently available on my Linux keyboard layer 3 😁

TimWolla:
> (I do wish more of the content from https://php-fpm.org/ moved to the PHP docs since it's interesting/useful background, but it's totally fair to make users find it instead of linking to it directly)

I just looked through that and didn't see anything particularly interesting. Anything you are specifically interested in? Given there is php/php-src#18217, perhaps add a short comment of what you would like to see moved to the official docs.

tianon:
I meant more of the superfluous overview stuff, like that bulleted list right on the homepage.  It's all very arguably "useless" information now that FPM is merged upstream (because it's information that probably helped convince folks it _should_ go upstream, as opposed to meaningful information for people choosing/running it now).

PR #2564, opened 2025-05-03
MediaWiki example `docker-compose.yml`: remove obsolete "version" attribute

DougInAMug:
This PR follows from discussion on https://phabricator.wikimedia.org/T393171 and makes slight formatting improvement following https://phabricator.wikimedia.org/T254405

PR #2566, opened 2025-05-05
Convert all `stack.yml` to `compose.yaml`

tianon:
Docker Swarm Mode has been in severely declining usage and maintenance for years now.  Meanwhile, Docker Compose has had quite a lot of changes (notably deprecating the `version:` field required by Swarm / forward compatibility).  As a result, it no longer makes a ton of sense for us to prefer/suggest Swarm usages.

This changes all references to be `compose.yaml` (the upstream-preferred canonical filename: https://github.com/compose-spec/compose-go/blob/9a9cc5d9c3453326bb8d506259bcfccd68af110b/cli/options.go#L384-L385) directly, leaving "swarm compatibility" as an exercise for the minority of readers who might still need/want it.

Unfortunately, this also means we need to remove all the play-with-docker links (since it only supports stack deploy via URL arguments, not compose), but my experience with that service is that it hasn't been terribly performant for a while, so I don't think this is a huge loss.

(There were also a few trailing references to the long-since-defunct Docker Machine project which have also been cleaned up.)

See also:

- https://github.com/docker-library/docs/pull/2514
- https://github.com/docker-library/docs/pull/2516#issuecomment-2574035913
- https://github.com/docker-library/docs/pull/2552#discussion_r1983899522
- https://github.com/docker-library/docs/pull/2564


tianon:
See also https://github.com/docker-library/docs/pull/913, which this is essentially the opposite of.

PR #2568, opened 2025-05-08
postgres: Fix incorrect shebang in example initialization script

reidb1:
The `/bin/bash` shebang points does not point to the correct path for bash in the container. Because of this, the initialization scripts will not run. Changing to `/usr/bin/env bash` fixes this problem.

Alternatively `/usr/bin/bash` would also work.


reidb1:
Interesting... I ran init scripts with `#!/bin/bash` and they did not run. _ONLY_ change I made was `#!/usr/bin/env bash` and they ran... 

Maybe there was something else that changed, but I was trying to be careful to make single changes at a time.

PR #2571, opened 2025-05-14
emqx: add deprecation notice

id:


emqx/deprecated.md
```diff
@@ -0,0 +1,5 @@
+Starting from v5.9.0, EMQX has unified all features from the previous Open Source and Enterprise editions into a single, powerful offering with the Business Source License (BSL) 1.1.
+
+If you want to understand why we made the change, please read this [blog post](https://www.emqx.com/en/news/emqx-adopts-business-source-license), and if you want to know more about the new license, please read the [EMQX Licensing FAQ](https://www.emqx.com/en/content/license-faq).
+
+Consequently, we stopped publishing the official Dockehub `emqx` image. EMQX v5.9.0+ will only be available in the `emqx/emqx` and `emqx/emqx-enterprise` Docker repositories.
```
yosifkit, line 5:
```suggestion
Consequently, we stopped publishing the `emqx` Docker Official Image. EMQX v5.9.0+ will only be available in the [`emqx/emqx`](https://hub.docker.com/r/emqx/emqx) and [`emqx/emqx-enterprise` ](https://hub.docker.com/r/emqx/emqx-enterprise) Docker Hub repositories.
```

yosifkit:
Will the `5.7` or `5.8` release series have any more updates, like from `5.7.2` to `5.7.3` or `5.8.6` to `5.8.7`? Or are they actually end of life and will receive no more updates/patches now that `5.9.0` is released?

id:
It's possible. 5.7.0 EOL is on November 27, 2025. We may release a patch if someone reports a critical vulnerability.

yosifkit (approved):

yosifkit:
We are sad to see `emqx` go but understand. Thanks for all your hard work. Until their respective EOL, there may still be fixes for 5.7 and 5.8 that come to Docker Official Images, is that correct?

From our side, we are happy to continue building them until their EOL date (which would be rebuilds when the base image is updated and for any 5.7.x or 5.8.x patch version bumps that are received). We can also stop the rebuilds as soon as you want us to (by removing them from the `library/emqx` file in https://github.com/docker-library/official-images).

id:
Correct!

Thank you for the support, and all the work that you do for DockerHub @yosifkit!

PR #2578, opened 2025-05-29
[ROS] Update Official Docker image docs

ruffsl:
Refreshes the ROS documentation using updated links, cleans out old content, and improves the demo examples.

Related:
- https://github.com/docker-library/docs/pull/1381

### DIFF
--- a/ros/content.md
+++ b/ros/content.md
@@ -62,12 +62,11 @@ repositories:
 EOF
 
 # update package cache
-RUN rm /etc/apt/apt.conf.d/docker-clean && apt-get update && \
-    rosdep update --rosdistro $ROS_DISTRO && \
-    cat <<EOF > /etc/apt/apt.conf.d/docker-no-recommends
+RUN cat <<EOF > /etc/apt/apt.conf.d/docker-clean
 APT::Install-Recommends "false";
 APT::Install-Suggests "false";
-EOF
+EOF && apt-get update && \
+    rosdep update --rosdistro $ROS_DISTRO
 
 # derive build/exec dependencies
 RUN bash -e <<'EOF'
--- a/ros/content.md
+++ b/ros/content.md
@@ -61,8 +61,10 @@ repositories:
     version: ${ROS_DISTRO}
 EOF
 
-# update apt cache
-RUN rm /etc/apt/apt.conf.d/docker-clean && apt-get update
+# update package cache
+RUN --mount=type=cache,sharing=locked,target=/var/cache/apt \
+    --mount=type=cache,sharing=locked,target=/var/lib/apt/lists \
+    apt-get update && rosdep update --rosdistro $ROS_DISTRO
 
 # derive build/exec dependencies
 RUN bash -e <<'EOF'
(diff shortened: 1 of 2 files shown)


tfoote (commented):
It generally looks like a good improvement. The naming of installer versus runner seem a little bit confusing to me. I feel like installer might be clearer with a name more like "upstream" or "prebuilt" maybe.

mikaelarguedas:
looks great overall !

agreed with Tully's comment

maybe using `latest` instead of `rolling` for the base images would be as future-proof and give example of using stable distros

ros/content.md
```diff
@@ -44,59 +44,85 @@ $ docker run -it --rm my/ros:app
...
 # multi-stage for caching
 FROM $FROM_IMAGE AS cacher
+ARG OVERLAY_WS
+
+# update package cache
+RUN rosdep update --rosdistro $ROS_DISTRO && \
+    cat <<EOF > /etc/apt/apt.conf.d/docker-clean && apt-get update
+APT::Install-Recommends "false";
+APT::Install-Suggests "false";
+EOF
```
tianon, line 59: This file (`/etc/apt/apt.conf.d/docker-clean`) already exists in the image -- did you overwrite that content on purpose, or did you intend for this to go to a different filename?

ros/content.md
```diff
@@ -44,59 +44,85 @@ $ docker run -it --rm my/ros:app
...
-      --mixin $OVERLAY_MIXINS
+      --mixin release
+
+# multi-stage for running
+FROM $FROM_IMAGE-ros-core AS runner
+ARG OVERLAY_WS
+
+# install exec dependencies
+COPY --from=cacher /tmp/exec_debs.txt /tmp/exec_debs.txt
+RUN --mount=type=cache,target=/etc/apt/apt.conf.d,from=cacher,source=/etc/apt/apt.conf.d \
```
tianon, line 118: You could do this in the `builder` stage too, but since this is the final stage it might be worth adding `exec_debs.txt` to a `--mount` instead so the file doesn't end up in the final image.

```suggestion
RUN --mount=type=bind,from=cacher,source=/tmp/exec_debs.txt,target=/tmp/exec_debs.txt \
    --mount=type=cache,target=/etc/apt/apt.conf.d,from=cacher,source=/etc/apt/apt.conf.d \
```

ros/content.md
ruffsl: Yeah, the overwrite was intentional to remove the prior rule that cleans the deb cache after dpkg is invoked.

It might be a too cute, but wanted to keep the example succinct , avoiding unnecessary rm or mv commands.

ros/content.md
ruffsl: I thought about doing that, but figured including the derived list of runtime dependencies in the output image could be more helpful for traceability and transparent debugging of the same layerer downstream. Users could of course `docker build --target=cacher ...` and inspect that intermediate image, but wouldn't propagate receipts.

ros/content.md
tianon: It wasn't immediately clear to me _why_ you'd want to remove that configuration -- maybe it warrants a comment that explains why / that this is intentional?

ruffsl: 7ea2ed688d114ab14fc7aa5cc7d9e6c468c0df4b ?

PR #2581, opened 2025-06-04
Officially deprecate SATOSA for inactivity

tianon:
See:

- https://github.com/docker-library/official-images/pull/15964#issuecomment-2741871173
- https://github.com/docker-library/official-images/pull/16594
- https://github.com/docker-library/official-images/pull/17307
- https://github.com/docker-library/official-images/pull/17382
- https://github.com/IdentityPython/satosa-docker/issues/11

To be explicitly clear, we'd be happy to bring this back / "cancel" this deprecation, but it will require re-commitment and updates to the image.

### DIFF
--- a/satosa/README-short.txt
+++ b/satosa/README-short.txt
@@ -1,1 +1,1 @@
-SATOSA translates between authentication protocols such as SAML2, OpenID Connect, and OAuth2.
+DEPRECATED; SATOSA translates between auth protocols such as SAML2, OpenID Connect, and OAuth2.
--- a/satosa/deprecated.md
+++ b/satosa/deprecated.md
@@ -1,0 +1,1 @@
+This image is deprecated due to maintainer inactivity (last updated Dec 2023; [docker-library/official-images#15964](https://github.com/docker-library/official-images/pull/15964)).


satosa/README-short.txt
```diff
@@ -1 +1 @@
-SATOSA translates between authentication protocols such as SAML2, OpenID Connect, and OAuth2.
+DEPRECATED; SATOSA translates between auth protocols such as SAML2, OpenID Connect, and OAuth2.
```
tianon, line 1: Unfortunately this was too long (over 100 chars, the Docker Hub length limit) once I added our standard prefix, so I made the best swap I could come up with to not lose much fidelity otherwise.

yosifkit:
Any recent pushes to Docker Hub would've just been from our automatic rebuilds caused by a base image update of any of the few tags left that still build. There hasn't been any [PRs to update the `satosa` image](https://github.com/docker-library/official-images/pulls?q=is%3Apr+label%3Alibrary%2Fsatosa) that weren't us disabling consistently failing builds since https://github.com/docker-library/official-images/pull/15964.

PR #2584, opened 2025-06-10
Rocky Linux: Adding note on where to find the latest versions

chris-short:
Rocky Linux users are finding the diff between Docker official images and Rocky Linux images confusing. I spoke with @tianon yesterday about this problem and it seems like oci-import will ease the burden for creating official images.

It's going to take us some time to re-factor, though. Adding a note to help folks find the latest images on the rockylinux/rockylinux page will help reduce confusion between the two. I'm happy to tune the language here, if needed.

### DIFF
--- a/rockylinux/deprecated.md
+++ b/rockylinux/deprecated.md
@@ -1,0 +1,3 @@
+# IMPORTANT NOTE
+
+The Docker team maintains official images. For the most up-to-date container images, please refer to the [Rocky Linux Docker Hub repository](https://hub.docker.com/r/rockylinux/rockylinux).


rockylinux/deprecated.md
```diff
@@ -0,0 +1,3 @@
+# IMPORTANT NOTE
+
+The Docker team maintains official images. For the most up-to-date container images, please refer to the [Rocky Linux Docker Hub repository](https://hub.docker.com/r/rockylinux/rockylinux).
```
tianon: We maintain the program, but y'all maintain the images within it, so this doesn't feel quite accurate :sweat_smile:

How about something like this?

```suggestion
The Docker team curates the Official Images program, and there are currently some technical constraints preventing Rocky Linux from publishing updates here. For the most up-to-date container images, please refer to the [Rocky Linux Docker Hub repository](https://hub.docker.com/r/rockylinux/rockylinux) for now.
```

chris-short:
sgtm, thank you @tianon

yosifkit (approved):

PR #2588, opened 2025-06-26
Add .gitattributes: Ensure Unix line endings for text files

Simran-B:
On Windows, Git may be configured to checkout LF line endings as CRLF, but this breaks shell scripts.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


Simran-B:
I meanwhile realized that you can't actually run markdownfmt.sh in e.g. Git Bash because something with the paths doesn't work with Windows as the host system...

PR #2589, opened 2025-06-27
Update Outdated Path in MySQL Configuration Docs

Disservin:
Resolves https://github.com/docker-library/mysql/issues/1083


tianon (approved):
Nice, thanks!

PR #2602, opened 2025-07-19
Add Clear Linux OS deprecation notice

tianon:
This text is copied from the official announcement at https://community.clearlinux.org/t/all-good-things-come-to-an-end-shutting-down-clear-linux-os/10716 (I only added the date of the announcement to clarify when "immediately" means since the image description is a more static context than the forum post).


tianon:
Oop, yep, good catch thanks

yosifkit (approved):

PR #2609, opened 2025-08-15
fix: point Ubuntu link to Ubuntu website

cjdcordeiro:


yosifkit (approved):
This was originally because the description above was taken from Wikipedia, but I think that hasn't been true for a while. 😅

LGTM

PR #2616, opened 2025-08-28
Update to golang 1.25

jbleduigou:
## What?

## Why?

Go version 1.25 was released on 12 of August.  
See https://go.dev/doc/devel/release#go1.25.0

## How?
Updated `content.md` to replace `1.24` by `1.25`.


tianon (approved):
Nice, thank you!

PR #2617, opened 2025-08-28
Update Matomo logo

J0WI:
https://github.com/matomo-org/docker/issues/379


tianon:
Perhaps this should swap to a `logo.svg` with https://commons.wikimedia.org/wiki/File:Matomo_Logo.svg ? :eyes:

PR #2620, opened 2025-09-02
Caddy: Added NET_BIND_SERVICE note

Minding000:
## Issue

The Caddy image needs the `NET_BIND_SERVICE` capability. There is no explanation on why and how to avoid it.
Especially in Kubernetes it is common to drop all capabilities and to use non privileged ports. Here the default behavior is confusing.

## Changes

Added explanation on why the capability is required by default and how to change it.
Included the error message to make it searchable.

### DIFF
--- a/caddy/content.md
+++ b/caddy/content.md
@@ -121,6 +121,8 @@ Giving the container this capability is optional and has potential, though unlik
 
 See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for more details.
 
+This image adds the `NET_BIND_SERVICE` capability to the Caddy executable, to allow a non-root user to bind to privileged ports (<1024) such as 80 and 443. If you're not using privileged ports and would like to drop the capability again, you can do so using `RUN setcap cap_net_bind_service=-ep /usr/bin/caddy`. This avoids the error `exec /usr/bin/caddy: operation not permitted` when the container process does not have the capability.
+
 ### Docker Compose example
 
 If you prefer to use `docker compose` to run your stack, here's a sample service definition which goes in a file named `compose.yaml`. The configuration assumes you put a custom Caddyfile into `$PWD/conf` as described [above](#basic-usage).


francislavoie:
We added that in 2023 @tianon, it was necessary for certain users who wanted to run the container as non-root, I think.

Added in https://github.com/caddyserver/caddy-docker/commit/b650917f3278ff4486f7e9e7b3cd63d6bad4ffc9, issue for context: https://github.com/caddyserver/caddy-docker/issues/104#issuecomment-1385288805

I don't really understand all the nuances here, maybe you could clarify with that context in mind :thinking:

tianon:
To illustrate further, with just Docker:

```console
$ docker run -it --rm --user 1000:1000 --cap-drop NET_BIND_SERVICE --name test "$(docker build -q - <<<$'FROM caddy\nRUN setcap cap_net_bind_service=-ep /usr/bin/caddy')"
2025/09/02 21:26:58.562	INFO	maxprocs: Leaving GOMAXPROCS=16: CPU quota undefined
2025/09/02 21:26:58.562	INFO	GOMEMLIMIT is updated	{"package": "github.com/KimMachineGun/automemlimit/memlimit", "GOMEMLIMIT": 60335323545, "previous": 9223372036854775807}
2025/09/02 21:26:58.562	INFO	using config from file	{"file": "/etc/caddy/Caddyfile"}
2025/09/02 21:26:58.563	INFO	adapted config to JSON	{"adapter": "caddyfile"}
2025/09/02 21:26:58.563	INFO	admin	admin endpoint started	{"address": "localhost:2019", "enforce_origin": false, "origins": ["//localhost:2019", "//[::1]:2019", "//127.0.0.1:2019"]}
2025/09/02 21:26:58.564	WARN	http.auto_https	server is listening only on the HTTP port, so no automatic HTTPS will be applied to this server	{"server_name": "srv0", "http_port": 80}
2025/09/02 21:26:58.564	INFO	tls.cache.maintenance	started background certificate maintenance	{"cache": "0xc0006d6400"}
2025/09/02 21:26:58.564	WARN	http	HTTP/2 skipped because it requires TLS	{"network": "tcp", "addr": ":80"}
2025/09/02 21:26:58.564	WARN	http	HTTP/3 skipped because it requires TLS	{"network": "tcp", "addr": ":80"}
2025/09/02 21:26:58.564	INFO	http.log	server running	{"name": "srv0", "protocols": ["h1", "h2", "h3"]}
2025/09/02 21:26:58.564	ERROR	unable to autosave config	{"file": "/config/caddy/autosave.json", "error": "open /config/caddy/autosave.json: permission denied"}
2025/09/02 21:26:58.564	INFO	serving initial configuration
2025/09/02 21:26:58.564	WARN	tls	unable to get instance ID; storage clean stamps will be incomplete	{"error": "open /data/caddy/instance.uuid: permission denied"}
2025/09/02 21:26:58.564	ERROR	tls	could not clean default/global storage	{"error": "unable to acquire storage_clean lock: creating lock file: open /data/caddy/locks/storage_clean.lock: no such file or directory"}
2025/09/02 21:26:58.564	INFO	tls	finished cleaning storage units
```

(listening just fine on port 80, even when started as user 1000:1000, and the capability in question explicitly dropped)

Minding000:
I think even `containerd` might allow privileged ports by default starting at version 2: https://github.com/containerd/containerd/issues/6924
I am running v1.7.27 (even there it is configurable) - the lastest available in the Docker APT repository.

I agree with @tianon that it makes sense to reconsider including `setcap` in the image. Especially with `docker` and `containerd` allowing unprivileged ports going forwards. Many projects in Kubernetes use unprivileged ports for HTTP and rely on the Ingress to route it to 443. Other projects can include `setcap` in their own image derivations.

But with many people still running older runtimes that would be a breaking change, so I don't expect this to happen (quickly).
Until then I think the documentation note is useful for anyone that is confused by the behavior.

Sidenote: `nginx` has a separate unprivileged image that exposes a different port by default avoiding the `setcap` issue: https://hub.docker.com/r/nginxinc/nginx-unprivileged

abjugard:
> FYI @abjugard, you had done the patch for setcap, do you have any objection to us reverting that, now that k8s should have caught up?

If the ports exposed is changed from 80/443 to higher numbers I don't see a reason for calling `setcap`, but as @Minding000 mentions many Linux distributions retain the configuration of ports below 1024 being privileged and requiring special capabilities be set on the binary in order to run as a non-root user.

JohanHellgrenTriona:
Just stumbled upon this issue running kubernetes + podman. Customer environment requires dropping all capabilities, so a note in the docs would have been handy.

PR #2627, opened 2025-10-07
MySQL: update readme; complete initializing a fresh instance paragraph

gouglhupf:
Completes and clarifies the “Initializing a fresh instance” section in mysql/content.md to accurately document how initialization files are discovered and executed by the image.

### What changed
Expanded the paragraph to explicitly document:

- Supported file extensions in "/docker-entrypoint-initdb.d": .sh, .sql, .sql.gz, .sql.bz2, .sql.xz, .sql.zst.
- Shell script handling: .sh without execute permission are sourced; executable .sh are run.

### Rationale
The previous text was incomplete/ambiguous about which files run and sourcing vs executing .sh files.
This aligns the docs with the actual behavior in the [entrypoint](https://github.com/docker-library/mysql/blob/3eee88bcaf4066ecef9edce0879053ce0dc3374b/8.4/docker-entrypoint.sh#L56) and aligns with the wording used in the `mariadb` image.

### Impact
Docs-only change

### DIFF
--- a/mysql/content.md
+++ b/mysql/content.md
@@ -134,7 +134,7 @@ Currently, this is only supported for `MYSQL_ROOT_PASSWORD`, `MYSQL_ROOT_HOST`, 
 
 # Initializing a fresh instance
 
-When a container is started for the first time, a new database with the specified name will be created and initialized with the provided configuration variables. Furthermore, it will execute files with extensions `.sh`, `.sql` and `.sql.gz` that are found in `/docker-entrypoint-initdb.d`. Files will be executed in alphabetical order. You can easily populate your `%%IMAGE%%` services by [mounting a SQL dump into that directory](https://docs.docker.com/storage/bind-mounts/) and provide [custom images](https://docs.docker.com/reference/dockerfile/) with contributed data. SQL files will be imported by default to the database specified by the `MYSQL_DATABASE` variable.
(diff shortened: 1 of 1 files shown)


mysql/content.md
```diff
@@ -134,7 +134,7 @@ Currently, this is only supported for `MYSQL_ROOT_PASSWORD`, `MYSQL_ROOT_HOST`,
...
+When a container is started for the first time, a new database with the specified name will be created and initialized with the provided configuration variables. Furthermore, it will execute files with extensions `.sh`, `.sql`, `.sql.gz`, `.sql.bz2`, `.sql.xz` and `.sql.zst` that are found in `/docker-entrypoint-initdb.d`. Files will be executed in alphabetical order. `.sh` files without file execute permission are sourced rather than executed. You can easily populate your `%%IMAGE%%` services by [mounting a SQL dump into that directory](https://docs.docker.com/storage/bind-mounts/) and provide [custom images](https://docs.docker.com/reference/dockerfile/) with contributed data. SQL files will be imported by default to the database specified by the `MYSQL_DATABASE` variable.
```
tianon:
```suggestion
When a container is started for the first time, a new database with the specified name will be created and initialized with the provided configuration variables. Furthermore, it will execute files with extensions `.sh`, `.sql`, `.sql.gz`, `.sql.bz2`, `.sql.xz`, and `.sql.zst` that are found in `/docker-entrypoint-initdb.d`. Files will be executed in alphabetical order. When parsing `.sh` files without the execute bit set, they are `source`d rather than executed.

You can easily populate your `%%IMAGE%%` services by [mounting a SQL dump into that directory](https://docs.docker.com/storage/bind-mounts/) and provide [custom images](https://docs.docker.com/reference/dockerfile/) with contributed data. SQL files will be imported by default to the database specified by the `MYSQL_DATABASE` variable.
```

- oxford comma (I was going to ignore this one because the prior text _also_ didn't use it :facepalm:  but then I had another suggestion so I figured it's worth noting while nothing this other suggestion:)
- slight re-word to avoid `.sh` at the start of a sentence (which creates an awkward flow)
- making `source` in `source`d a code block (so it's more clear this is the shell's concept of `source`, specifically)
- separating what's now a very long single block into two blocks

PR #2632, opened 2025-10-22
Update content.md to reflect Liquibase 5.0 changes

jandroav:
This pull request updates the `liquibase/content.md` documentation to highlight major changes introduced in Liquibase 5.0. The most important updates clarify the separation between the open source and commercial editions, and document breaking changes regarding database drivers and extensions.

Liquibase 5.0 product separation and usage:

* Added a section describing the distinction between the OSS edition (`liquibase/liquibase`) under the Functional Source License (FSL) and the commercial Secure edition (`liquibase/liquibase-secure`), including guidance for users with a license key.

Breaking changes to driver and extension management:

* Documented that database drivers and extensions are no longer included by default in the OSS edition and official Docker image, requiring explicit installation via the Liquibase Package Manager (LPM) or manual mounting.
* Provided instructions and Dockerfile examples for adding database drivers using LPM, and clarified continued support for MySQL driver installation via the `INSTALL_MYSQL=true` environment variable.

Reference to release announcement:

* Included a link to the official Liquibase 5.0 release announcement for further details.

### DIFF
--- a/liquibase/deprecated.md
+++ b/liquibase/deprecated.md
@@ -2,7 +2,7 @@ With the release of Liquibase 5.0, this repository is officially deprecated and 
 
 Users are encouraged to switch to one of the following alternatives, depending on your needs:
 
-- `liquibase/liquibase` -- Community Edition under Functional Source License (if you have a valid Liquibase License Key, you should use `liquibase/liquibase-secure` instead)
-- `liquibase/liquibase-secure` -- Commercial version with enterprise features
+-	`liquibase/liquibase` -- Community Edition under Functional Source License (if you have a valid Liquibase License Key, you should use `liquibase/liquibase-secure` instead)
+-	`liquibase/liquibase-secure` -- Commercial version with enterprise features
 
 As of Liquibase 5.0, the Community edition and the official Docker Community liquibase image no longer include database drivers or extensions by default. You must now explicitly add database drivers using the Liquibase Package Manager (LPM), or manually install/mount extensions into the container. See [the Liquibase 5.0 Release Announcement](https://www.liquibase.com/blog/liquibase-5-0-release) for more details.


jandroav:
I fixed the markdown lint issue, but amending the commit to a single commit doesn’t re-trigger the CI job.

tianon:
Thanks for getting this started!

I think this needs to be more explicit about the fact that 5.0 will be the last version supported here and that users should very actively transition over to `liquibase/liquibase{,-secure}` images instead.  We support a `deprecated.md` file that's intended for this exact purpose that I think we need to make explicit use of here, especially as it will render above everything else.

Here's a proposed start to something like that (consistent with how we've done this in the past): https://github.com/docker-library/docs/commit/72010b9a022f51750963ec2c48e02495a875e09a

That necessarily implies some changes to the contents of _this_ PR too :see_no_evil:
(if only we'd caught this sooner and stopped 5.x from entering this program entirely so the breaking changes don't need to be documented here too :sweat_smile: :disappointed:)

One thing we could consider is moving `latest` back to 4.x here (or even dropping the `latest` tag entirely).

Then we could also add something like "for details of breaking changes in 5.x, see X" if you've got some upstream documentation somewhere that explains all this nicely?  (that would shrink the needed documentation _here_ for 5.0 that then needs to be maintained over time)

liquibase/deprecated.md
```diff
@@ -0,0 +1,8 @@
+With the release of Liquibase 5.0, this repository is officially deprecated and will not receive any further releases beyond patch updates.
+
+Users are encouraged to switch to one of the following alternatives, depending on your needs:
+
+- `liquibase/liquibase` -- Community Edition under Functional Source License (if you have a valid Liquibase License Key, you should use `liquibase/liquibase-secure` instead)
+- `liquibase/liquibase-secure` -- Commercial version with enterprise features
```
tianon:
```suggestion
-	`liquibase/liquibase` -- Community Edition under Functional Source License (if you have a valid Liquibase License Key, you should use `liquibase/liquibase-secure` instead)
-	`liquibase/liquibase-secure` -- Commercial version with enterprise features
```

PR #2637, opened 2025-11-12
Correct examples in 'Changing DocumentRoot'

No5251:
Replace single quotes with double quotes so that the variable is replaced


yosifkit:
I believe that the single quotes are on purpose since Apache httpd will evaluate them in the config at runtime. This allows extra ability to change the root at runtime as well setting a default during a custom build.

PR #2639, opened 2025-11-21
Update `push.pl` runtime to Perl 5.40 and Debian Trixie

yosifkit:
Mostly trying to update to Trixie and that required to move to a supported version of Perl like 5.40 which is probably good since 5.39 is unmaintained/unsupported. Unfortunately, the Dockerfile fails to build when switched to Perl 5.42 (failing tests in EV).

I haven't yet tested this like https://github.com/docker-library/docs/pull/2466.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM perl:5.39-bookworm
+FROM perl:5.40-trixie
 
 RUN set -eux; \
 	apt-get update; \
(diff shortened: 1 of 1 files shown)


ljhlh49-dot:
> > I haven't yet tested this like https://github.com/docker-library/docs/pull/2466.
> 
> 
> 
> I have (now) :+1: :heart:

PR #2666, opened 2026-04-22
update to golang 1.26

jbleduigou:
## What?

## Why?

Go version 1.25 was released in February.
See https://go.dev/doc/devel/release#go1.26.0

## How?
Updated `content.md` to replace `1.25` by `1.26`.


yosifkit (approved):
Thank you!

PR #2672, opened 2026-05-14
Nim: Add docs.

moigagoo:


nim/logo.png
tianon, line 1: FYI, this could be `logo.svg` if you'd rather -- our tooling will convert/resize it as necessary for Docker Hub and then we'd not need to bug you if Docker Hub ever needs something bigger :smile:

(it looks like you probably have an SVG version upstream in https://github.com/nim-lang/assets? :eyes:)

tianon (approved):
Nice, looks good to me, just one minor non-blocking note about the logo file 😄

nim/logo.png
moigagoo: Thanks for the tip! Replaced png with svg.

yosifkit (approved):

PR #2674, opened 2026-05-21
geonetwork: fix Elasticsearch version for 4.4 images

juanluisrp:
## What

Updates the GeoNetwork Docker Hub documentation to fix an incompatibility between the compose example and the current `geonetwork:4.4` image.

## Why

The `compose.yaml` and quick-start examples specified `elasticsearch:7.17.15`, but `geonetwork:4.4` now resolves to 4.4.9+ which requires Elasticsearch 8.x. Starting from GeoNetwork 4.4.3, the required ES version moved to 8.x. ES 8 changed how text-field aggregations work, causing 400 errors on the Map and Search pages.

Reported in: https://github.com/geonetwork/docker-geonetwork/issues/144

## Changes

**`compose.yaml`**
- Upgrade `elasticsearch` and `kibana` from `7.17.15` to `8.14.3`
- Add `xpack.security.enabled: "false"` and `xpack.security.enrollment.enabled: "false"` (ES 8 enables auth by default; required for a local demo setup)
- Remove three Jetty-specific JVM flags that have no effect on the current Tomcat-based images
- Change `GEONETWORK_DB_TYPE` from `postgres-postgis` to `postgres`
- Replace `postgis/postgis:17-3.5` with `postgres:17` (multi-arch; PostGIS not required with `GEONETWORK_DB_TYPE=postgres`)
- Improve database healthcheck: use `psql` to verify the geonetwork database and user exist before starting GeoNetwork, instead of `pg_isready -U postgres` which passes too early during the PostgreSQL init cycle
- Improve GeoNetwork healthcheck: verify the actual app endpoint rather than the bare root URL, add `start_period: 60s`
- Remove `pglog` volume (was only used by the now-removed verbose postgres logging command)

**`content.md`**
- Update quick-start `docker run` example for GN 4.4 to use `elasticsearch:8.14.3` with `xpack.security.enabled=false`
- Add Elasticsearch/GeoNetwork version compatibility table
- Fix "Tomcat (GN 3) or Jetty (GN 4)" description: Jetty 9 was used in 4.0.0-4.2.14 and 4.4.0-4.4.9; Tomcat is used for GN 3, 4.2.15+, and 4.4.10+
- Correct H2 database path description for Tomcat vs Jetty images

## Testing

Brought up the updated stack locally, loaded templates and samples, and confirmed no "Query returned an error" errors on the Map and Search pages. All `_search` requests return 200 with correct aggregation results.

### DIFF
--- a/geonetwork/content.md
+++ b/geonetwork/content.md
@@ -19,12 +19,12 @@ GeoNetwork 4 uses an Elasticsearch server to store the index of the documents it
 This is a quick example of how to get GeoNetwork 4.4 Latest up and running for demo purposes. This configuration doesn't keep the data if containers are removed.
 
 ```console
-docker pull elasticsearch:7.17.15
+docker pull elasticsearch:8.14.3
 docker pull %%IMAGE%%:4
 
 docker network create gn-network
 
-docker run -d --name my-es-host --network gn-network -e "discovery.type=single-node" elasticsearch:7.17.15
+docker run -d --name my-es-host --network gn-network -e "discovery.type=single-node" -e "xpack.security.enabled=false" elasticsearch:8.14.3
 docker run --name %%REPO%%-host --network gn-network -e GN_CONFIG_PROPERTIES="-Des.host=my-es-host -Des.protocol=http -Des.port=9200 -Des.url=http://my-es-host:9200" -p 8080:8080 %%IMAGE%%:4
 ```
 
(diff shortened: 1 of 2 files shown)


tianon:
(when this is ready for review, be sure to ping us explicitly as GitHub will not notify us for PR status changes like "Draft" -> "Ready for review" :+1:)

juanluisrp:
@tianon ready for review.

geonetwork/content.md
```diff
@@ -120,11 +137,27 @@ docker run --name some-%%REPO%% -d -p 8080:8080 %%IMAGE%%
 
 Then, if you are running docker on Linux, you may access geonetwork at http://localhost:8080/geonetwork.
 
+### File permissions
+
+The Tomcat-based images (GN 3, 4.2.15+, and 4.4.10+) run as `root`, whereas the previous Jetty-based images (GN 4.0.0-4.2.14 and 4.4.0-4.4.9) ran as the `jetty` user. If you are **upgrading from a Jetty-based image** and mounting a host directory or named volume for the data directory, you may need to update the ownership of existing data to avoid permission errors:
+
+```console
+docker run --rm -v /host/%%REPO%%-docker:/catalogue-data busybox chown -R root:root /catalogue-data
```
tianon: This should either have the `console` sentinel or swap to `bash`:


````suggestion
```console
$ docker run --rm -v /host/%%REPO%%-docker:/catalogue-data busybox chown -R root:root /catalogue-data
````


````suggestion
```bash
docker run --rm -v /host/%%REPO%%-docker:/catalogue-data busybox chown -R root:root /catalogue-data
````

tianon (approved):
LGTM, just one super minor nit 👍

PR #2677, opened 2026-06-23
Update docs for ghost

acburdine:
Ran `./update.sh ghost` locally and for me it would remove the section of README.md on image variants, not sure if that's a quirk of my local setup or something I forgot to update, flagging it here in case that's a known issue

### DIFF
--- a/ghost/compose.yaml
+++ b/ghost/compose.yaml
@@ -1,7 +1,7 @@
 services:
 
   ghost:
-    image: ghost:5-alpine
+    image: ghost:6-alpine
     restart: always
     ports:
       - 8080:2368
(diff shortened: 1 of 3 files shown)


yosifkit (approved):

yosifkit:
I'm not sure what local difference would cause it to remove the variant section 🤔. Perhaps a bug in [`variant.sh`](https://github.com/docker-library/docs/blob/baabee2bcfb186dbc8f13f8336361d8c0e08ffdf/.template-helpers/variant.sh)? 🤷. It canonically runs on the DOI Jenkins Linux workers, so it could be a host config or OS assumption.

acburdine:
> It canonically runs on the DOI Jenkins Linux workers

Ah you know that's probably it - macOS BSD `sed` strikes again 😭

tianon:
One trick I've used successfully is a bind-mount into a Linux container like `buildpack-deps:curl` or `tianon/network-toolbox` (depending on what you need available) and running the script from there. :joy:

PR #2682, opened 2026-07-24
Nim: Add info about compiling to JS and running.

moigagoo:


nim/content.md
```diff
@@ -18,10 +18,22 @@ $ docker run --rm -v "$PWD":/usr/src/app -w /usr/src/app %%IMAGE%% nim c -r main
...
+
+To compile and run, use a multi-stage Dockerfile with Node.js:
+
+```dockerfile
+FROM %%IMAGE%% AS builder
+COPY . .
+RUN nim js -o:app.js src/app.nim
+
+FROM node:latest
+COPY --from=builder /usr/src/app/app.js .
```
tianon, line 35: FYI, this is encoding the implicit assumption of `WORKDIR /usr/src/app` in the `nim` image (which is fine, but something you should be aware of / be doing intentionally if you're going to do it, since it makes that value slightly "stickier" to document it this way)

tianon (approved):
Looks fine to me -- do you want to wait for your other PR to merge this, or merge this as-is since it's correct even without image changes?

(also one non-blocking note/question below)

nim/content.md
moigagoo: True, and I was somewhat conflicted with this line... Initially, there was `WORKDIR /app` in both the builder and the runner images, which does make things more explicit but also noisier.

I think that as an example that is intended to be modified for your needs (I mean, you probably don't have an app called `app` anyway, duh) this is fine.

moigagoo: > Looks fine to me -- do you want to wait for your other PR to merge this, or merge this as-is since it's correct even without image changes?

I'd rather have that merged now as is.

PR #2693, opened 2026-08-27
Update Go image version from 1.26 to 1.27

jbleduigou:
## What?

## Why?

Go version 1.27 was released in August.
See https://go.dev/doc/devel/release#go1.27.0

## How?
Updated `content.md` to replace `1.26` by `1.27`.


tianon (approved):

tianon:
Thanks!

PR #2675, opened 2026-05-29
Updated documentation for Convertigo 8.4.3

nicolas-albert:
Thx!

PR #2701, opened 2026-09-15
Updated documentation for Convertigo 8.4.4

nicolas-albert:
Sync the Convertigo documentation with the 8.4.4 release.


yosifkit:
Previous comments on https://github.com/docker-library/docs/pull/2675 apply here as well:

https://github.com/docker-library/docs/pull/2675#pullrequestreview-4431238627:
> I'd recommend not removing the `$` at the beginning of command lines in a console code block, since that lets the syntax highlighter know that the line should be highlighted like a command and not the output of a command (with the line-end backslashes letting it know to continue the command syntax).

https://github.com/docker-library/docs/pull/2675#pullrequestreview-4431239682
> There are a few instances like this one where an en dash is sneaking in where it does not make sense, like a command line (and one where it already has on L284/R381). Plus, this is incorrectly replacing a `--name` with `-name` even if the dash was the correct dash (`-`) and not en dash (`–`).

convertigo/content.md
```diff
@@ -26,22 +26,62 @@ You can access the Server admin console on `http://[dockerhost]:28080/convertigo
...
+### CouchDB running in another container
+
+Create a user-defined Docker network and run both containers on it:
+
+```console
+$ docker network create c8o-net
+```
+
+```console
+$ docker run -d --name fullsync --network c8o-net couchdb:3.2.2
```
tianon, line 72: Not a blocker, but as a note, this is a pretty old version of CouchDB now (EOL 2 years ago, first released almost 5 years ago), so you might want to update to something more modern. 😅

convertigo/content.md
```diff
@@ -79,13 +124,91 @@ You can share the same workspace by all Convertigo containers. In this case, whe
...
+
+The image is based on the Eclipse Temurin JDK image, which ships an entrypoint able to add certificate authorities to the JVM truststore. It is opt-in: set the `USE_SYSTEM_CA_CERTS` environment variable and mount the certificates, in PEM format with a `.crt` extension, in the `/certificates` directory. A file may contain several certificates.
+
+```console
+$ mkdir -p custom-ca
+$ cp company-root-ca.crt custom-ca/
+$ cp partner-intermediate-ca.crt custom-ca/
+$ docker run --name C8O \
+    -e USE_SYSTEM_CA_CERTS=1 \
+    -v "$(pwd)/custom-ca:/certificates:ro" \
```
tianon, line 204: Just an FYI, you can avoid the subshell here with the shell-provided `$PWD` variable (and elsewhere):

```suggestion
    -v "$PWD/custom-ca:/certificates:ro" \
```

tianon (approved):

